PR #90, opened 2020-09-28
Frontend integration, node dependency removal, major endpoints overhaul

Artur64:
This PR is fairly big, excluding various minor stuff/adjustments, it mainly contains:
1. Integration of the Nuxtjs written frontend into this project (**NOTE**: Routes,endpoints and overall the frontend is **NOT** adjusted to work directly with current phoenix backend, rather than old rust node, this will be a scope of another issue). Changes made to the frontend itself are not really big(which definitely will require some revision and adjustments in future), mainly configuration files and swagger endpoint link are updated to match with our own hosted swagger page.
2. Phoenix endpoint configuration changes and adjustment(Serving of newly added and generated frontend files)
3. Router changes - After the integration, it was needed to separate static files endpoint - from mdw api endpoint, so it was decided to move all current mdw implemented endpoints to serve a different scope - `/mdw/*some_path` instead of just `/`, as `/` is now being redirected to render the frontend static files, all routes which are valid/invalid(in a scope of `/`), by default are matched with frontend's static file entrypoint `priv/static/frontend/index.html`. Ex. frontend can now be accessed by visiting `localhost:4000/` and the phoenix backend api `localhost:4000/mdw/old_phx_routes`
4. Documentation, swagger and tests adjustments(in regards of changes, described above)
Please, don't hesitate to test it and ask any question/give any suggestions you may have!


skkw (commented):
Looking forward to have this merged!

But, would it be possible to keep endpoints without `mdw/` prefix? 

Ideally a new change shouldn't require users of the MDW (Superhero team, ae analytics) to change their call sites, also for users of MDW who don't require frontend the `mdw` prefix is superfluous.

skkw (approved):

PR #210, opened 2021-08-17
fix: don't read from cache the last 6 blocks

### LINKED ISSUE #194: different synced mdw return different results, e.g. micro block count
https://mainnet.aeternity.art/mdw/blocks/444733-444733
https://mainnet.aeternity.io/mdw/blocks/444733-444733

count of microblocks is different

jyeshe:
## What

Change controller of blocks endpoint to not read from cache the last 6 blocks.

The expiration of inserted blocks in the cache remains the same, based on `generations_cache_expiration_minutes` config. 

## Why

Issue [#194](https://github.com/aeternity/ae_mdw/issues/194)

https://mainnet.aeternity.art/mdw/blocks/444733-444733
https://mainnet.aeternity.io/mdw/blocks/444733-444733

count of microblocks is different

## Cause

Once inserted in the cache, a block is only updated after the expiration.

## Additional Notes

The list comprehension with `reduce` was replaced with `into` as it's a little faster and saves memory (around 50% according benchee).

fixes #194

### DIFF
--- a/lib/ae_mdw_web/controllers/block_controller.ex
+++ b/lib/ae_mdw_web/controllers/block_controller.ex
@@ -15,10 +15,15 @@ defmodule AeMdwWeb.BlockController do
   import AeMdw.{Util, Db.Util}
 
   @tab __MODULE__
-  ##########
+
+  # number of blocks behind latest that are cached
+  @blocks_cache_threshold 6
 
   def table(), do: @tab
 
+  @doc """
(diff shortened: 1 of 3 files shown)


wj2016 (approved):
Looks good

sborrazas (approved):
At some point we should probably move this out of the controller and into AeMdw, but this shouldn't be part of this PR. Approved

PR #216, opened 2021-08-19
Do not return duplicate transactions [#162]

### LINKED ISSUE #162: Sending coins to itself makes appear transaction in the list of transactions two times
For example, here: https://mainnet.aeternity.io/mdw/txs/gen/421792-0?account=ak_gvxNbZf5CuxYVfcUFoKAP4geZatWaC2Yy4jpx5vZoCKank4Gc&limit=10&page=1

Transaction th_3rw5nk53rEQzZr8xrQ6sqt2Y9Cv4U8piqiHK1KVkXCaVFMTCq is listed two times.

Obviously, this list shouldn't return duplicate transactions, because the endpoint described as returning transactions related to the account, not incoming/outcoming entries.

wj2016:
Resolves #162


sborrazas (approved):

lib/ae_mdw_web/cont.ex
```diff
@@ -67,6 +67,17 @@ defmodule AeMdwWeb.Continuation do
     end
   end
 
+  def unique_take_and_drop(stream, limit, acc \\ []) do
```
jyeshe, line 70: nitpicking `defp`

lib/ae_mdw_web/cont.ex
```diff
@@ -67,6 +67,17 @@ defmodule AeMdwWeb.Continuation do
...
+    {data, rem_stream} = StreamSplit.take_and_drop(stream, limit)
+    acc = Enum.concat(acc, data) |> Enum.uniq()
```
jyeshe, line 72: For comprehending, as the `acc` argument is not passed, what case in the future do you think it would be needed? Would you mind adding it later if we are not sure it gonna help soon?

jyeshe (commented):
One minor and a question.

lib/ae_mdw_web/cont.ex
wj2016: it's used in line 75, right?

jyeshe: Yes. I mean the argument is not passed to `unique_take_and_drop` so for a first fix maybe we could have the function without the acc parameter if we won't use it soon.

thepiwo (changes requested):
So this is just fixed when fetching the txs, but they are still stored in duplicate. I assume the Enum.uniq() solution is not so performant when reading, as if we only had to do it once, when writing?

lib/ae_mdw_web/cont.ex
jyeshe: For the `Enum.uniq`, on @thepiwo consideration, `Enum.uniq_by` would be faster if we wanna solve the problem firstly in the read.

lib/ae_mdw/db/sync/transaction.ex
```diff
@@ -128,14 +128,40 @@ defmodule AeMdw.Db.Sync.Transaction do
...
+      resolve_pubkey(elem(tx, ids.recipient_id), :spend_tx, :recipient_id, block_index)
+
+    case sender_pk == recipient_pk do
+      true ->
+        write_field(:spend_tx, ids.sender_id, sender_pk, txi)
+
+      false ->
+        write_field(:spend_tx, ids.sender_id, sender_pk, txi)
+        write_field(:spend_tx, ids.recipient_id, recipient_pk, txi)
+    end
```
jyeshe, line 162: sync_spend_tx could return :ok to be explicit we don't need the result

jyeshe (commented):
suggestions on code style

lib/ae_mdw/db/sync/transaction.ex
jyeshe: it would help with the dialyzer complain

jyeshe: actually did not because Function sync_spend_tx/3 has no local return. is due to tx_ids as you already noticed.

jyeshe: but the :ok return would be welcome (:

thepiwo:
@wj2016 whats the state of this?

wj2016:
fixed stupid dialyzer check

thepiwo (commented):
looks good to me now, lets wait for the others to review

jyeshe (approved):

PR #217, opened 2021-08-19
ci: add dialyzer to project

### LINKED ISSUE #213: adjust code to comply to dyalizer

sborrazas:
closes #213


lib/ae_mdw/db/format.ex
```diff
@@ -234,7 +234,7 @@ defmodule AeMdw.Db.Format do
       log: logs.(:aect_call.log(call_rec))
     }
 
-    m = (is_map(fun_arg_res) && Map.merge(call_info, fun_arg_res)) || call_info
+    m = Map.merge(call_info, fun_arg_res)
```
jyeshe, line 237: nice catch

lib/ae_mdw/db/stream/query/parser.ex
```diff
@@ -45,6 +45,10 @@ defmodule AeMdw.Db.Stream.Query.Parser do
     {ids, types}
   end
 
+  def parse(_query_groups) do
+    {[], []}
```
jyeshe, line 49: `{Map.new(), MapSet.new()}` to match the types.

lib/ae_mdw/node.ex
```diff
@@ -0,0 +1,147 @@
+defmodule AeMdw.Node do
```
jyeshe, line 1: Wonderful, this will clean some warnings too. 
Wouldn't the calls need a dispatch to the Smart instance?

lib/ae_mdw_web/benchmark/bench.ex
```diff
@@ -92,8 +92,6 @@ defmodule Mix.Tasks.Bench do
 
   defp percentile([], _), do: nil
   defp percentile([x], _), do: x
-  defp percentile(list, 0), do: min(list)
-  defp percentile(list, 100), do: max(list)
```
jyeshe, line 96: maybe leave this optimization to avoid sorting all.

lib/ae_mdw_web/controllers/aex9_controller.ex
```diff
@@ -204,7 +204,7 @@ defmodule AeMdwWeb.Aex9Controller do
   end
 
   def balance_for_hash_reply(conn, contract_pk, account_pk, {type, block_hash, height}) do
-    {amount, _} = DBN.aex9_balance(contract_pk, account_pk, {height, block_hash})
+    {amount, _} = DBN.aex9_balance(contract_pk, account_pk, {type, height, block_hash})
```
jyeshe, line 207: I would rather avoid operational changes in this PR. If it's a bug, better report it to keep the tracking also.
`_type` would suffice for now.

lib/ae_mdw_web/controllers/aex9_controller.ex
```diff
@@ -250,7 +250,7 @@ defmodule AeMdwWeb.Aex9Controller do
   end
 
   def balances_for_hash_reply(conn, contract_pk, {block_type, block_hash, height}) do
-    {amounts, _} = DBN.aex9_balances(contract_pk, {height, block_hash})
+    {amounts, _} = DBN.aex9_balances(contract_pk, {block_type, height, block_hash})
```
jyeshe, line 253: same

mix.exs
```diff
@@ -62,7 +63,16 @@ defmodule AeMdw.MixProject do
       {:websockex, "~> 0.4.2"},
       {:phoenix_swagger, "~> 0.8"},
       {:temp, "~> 0.4"},
-      {:tesla, "~> 1.3.0"}
+      {:tesla, "~> 1.3.0"},
+      {:dialyxir, "~> 1.0", only: [:dev], runtime: false}
+    ]
+  end
+
+  defp dialyzer do
```
jyeshe, line 71: neat!

jyeshe (changes requested):

lib/ae_mdw/db/stream/query/parser.ex
sborrazas: Nice catch, done

lib/ae_mdw/node.ex
sborrazas: This module is compiled and then loaded before the app starts, then the app overrides everything in it once it starts. This can be [seen here](https://github.com/ks/smart_global/blob/master/lib/smart_global.ex#L68)

lib/ae_mdw_web/benchmark/bench.ex
sborrazas: Dialyzer was complaining about these two clauses because they weren't being used

lib/ae_mdw_web/controllers/aex9_controller.ex
sborrazas: Removed these changes from the PR

lib/ae_mdw_web/benchmark/bench.ex
jyeshe: Indeed

jyeshe (commented):
Got this with `mix dialyzer` (commit 76956efe):
```lib/ae_mdw_web/controllers/aex9_controller.ex:62:no_return
The created anonymous function has no local return.
________________________________________________________________________________
lib/ae_mdw_web/controllers/aex9_controller.ex:132:no_return
The created anonymous function has no local return.
________________________________________________________________________________
lib/ae_mdw_web/controllers/aex9_controller.ex:206:no_return
Function balance_for_hash_reply/4 has no local return.
________________________________________________________________________________
lib/ae_mdw_web/controllers/aex9_controller.ex:207:call
The function call will not succeed.

AeMdw.Node.Db.aex9_balance(_contract_pk :: any(), _account_pk :: any(), {_, _})

will never return since the 3rd arguments differ
from the success typing arguments:

(any(), any(), boolean() | {_, _, _})

________________________________________________________________________________
lib/ae_mdw_web/controllers/aex9_controller.ex:252:no_return
Function balances_for_hash_reply/3 has no local return.
________________________________________________________________________________
lib/ae_mdw_web/controllers/aex9_controller.ex:253:call
The function call will not succeed.

AeMdw.Node.Db.aex9_balances(_contract_pk :: any(), {_, _})

will never return since the 2nd arguments differ
from the success typing arguments:

(any(), boolean() | {_, _, _})

________________________________________________________________________________
done (warnings were emitted)
Halting VM with exit status 2
```

jyeshe (approved):
Great PR. 
The only pending `dialyzer` finding reported here is the expected one regarding the tuple passed to `AeMdw.Node.Db.aex9_balances(_contract_pk :: any(), {_, _})`

lib/ae_mdw/node.ex
```diff
@@ -0,0 +1,147 @@
...
+  @moduledoc """
+  Sample module to understand all of the functions the AwMdw.Node module
+  provides. Including it's specs as well.
+
+  Right now this module and its functions are defined using the
+  SmartGlobal library at runtime. The purpose of the module is to make
```
thepiwo, line 7: Why do we need this?

thepiwo (approved):

lib/ae_mdw/node.ex
sborrazas: For two reasons:

* It's very hard to know what this module functions are, because they are built during runtime
* Dialyzer complains about function calls made to this module (because they are not available at compile-time)

.dialyzer_ignore.exs
```diff
@@ -0,0 +1,9 @@
+[
+  # The "does not exist" warning happens because of all of the functions
+  # created during runtime that belong to the AE core code.
+  ~r/does not exist/,
+  # Following warnings are due to bad implementatino of riverside.
```
Artur64, line 5:
```suggestion
  # Following warnings are due to bad implementation of riverside.
```

PR #218, opened 2021-08-20
fix: adjust tuple structure sent on AEX9 balances endpoints

sborrazas:
This was found using dialyzer, and was causing the AEX9 balance endpoints to return a 500 error.


sborrazas:
@thepiwo should I create a GH issue for this?

thepiwo (commented):
looks good to me, will leave for others to review as well

thepiwo:
@sborrazas we don't need an issue if we already have a fix

jyeshe (approved):
:+1:

Artur64 (approved):
Looks good to me, I will also add some tests to aex9 controller

PR #221, opened 2021-08-24
test: separate unit/integration tests and add to ci

sborrazas:
This initial approach only moved two of the integration tests to be unit tests instead (by using Mock). All other tests should be moved in this same manner.


wj2016 (approved):
Really cool!

Artur64 (approved):

PR #222, opened 2021-08-25
fix: remote call log also indexed by called contract

jyeshe:
## What

1. Inserts the event log indexing by the remote called contract.  
2. Returns the caller/parent contract id on contract/logs endpoints looking up by the transaction call index (call_idx).

## Why

Currently the events are logged only for the caller contract on a remote call. 
It's expected to able to search for all events emitted by a contract, including those happening due to a remote call (by another contract).

## Additional Notes

The contract id is identified in the log by the `create_txi` (transaction id of the creation of the contract).
The `ext_caller_contract_id` field in the json response means the final caller (contract A when contract B calls contract A which emits the event).


lib/ae_mdw/db/sync/transaction.ex
```diff
@@ -192,6 +192,7 @@ defmodule AeMdw.Db.Sync.Transaction do
   def write_origin(tx_type, pubkey, txi, tx_hash) do
     m_origin = Model.origin(index: {tx_type, pubkey, txi}, tx_id: tx_hash)
     m_rev_origin = Model.rev_origin(index: {txi, tx_type, pubkey})
+    IO.inspect m_origin
```
sborrazas, line 195: I'm assuming this was for debugging purposes and no longer needed?

mix.exs
```diff
@@ -64,6 +64,7 @@ defmodule AeMdw.MixProject do
       {:phoenix_swagger, "~> 0.8"},
       {:temp, "~> 0.4"},
       {:tesla, "~> 1.3.0"},
+      {:assertions, "~> 0.18.1"},
```
sborrazas, line 67: I think we need to add `only: [:test]` here.

lib/ae_mdw_web/controllers/aex9_controller.ex
```diff
@@ -388,13 +382,13 @@ defmodule AeMdwWeb.Aex9Controller do
...
+  defp enc_block(:key, hash), do: :aeser_api_encoder.encode(:key_block_hash, hash)
+  defp enc_block(:micro, hash), do: :aeser_api_encoder.encode(:micro_block_hash, hash)
 
-  def enc_ct(pk), do: :aeser_api_encoder.encode(:contract_pubkey, pk)
-  def enc_id(pk), do: :aeser_api_encoder.encode(:account_pubkey, pk)
+  defp enc_ct(pk), do: :aeser_api_encoder.encode(:contract_pubkey, pk)
+  defp enc_id(pk), do: :aeser_api_encoder.encode(:account_pubkey, pk)
 
-  def enc(type, pk), do: :aeser_api_encoder.encode(type, pk)
+  defp enc(type, pk), do: :aeser_api_encoder.encode(type, pk)
```
sborrazas, line 391: Thank you for making all of these private, it makes the module API a lot clearer! :pray:

sborrazas (changes requested):
Really nice and clean! Just a few comments

lib/ae_mdw_web/controllers/aex9_controller.ex
jyeshe: It was only to validate one point. Gonna remove the aex9 branch files.

sborrazas (approved):

Issue #193, opened 2021-06-17
provide way to automate database migrations


thepiwo:

jyeshe:
Proposal draft: https://github.com/aeternity/ae_mdw/pull/232
Working with first commit: 29f254d
Last 2c96f71 still to be fixed with RocksDB.

PR #232, opened 2021-08-31
database: automate migration

jyeshe:
## What

Ecto-like database migration automation.

## Why

Issue #193
Avoids running migrations twice and saves the state to follow the changes in db structure and data.

## Validation steps

1. `make shell`
2. `mix migrate_db`
This migration is expected to be applied:
```
10:57:54.933 [info]  applying version 20210826171900 with Elixir.AeMdw.Migrations.ReindexRemoteLogs...
table size: 177072
Indexed 32142 records in 62s

10:58:56.870 [info]  applied version 20210826171900
```
3. `mix migrate_db`
Migration is already is last version:
```
11:00:05.639 [info]  current migration version: 20210826171900

11:00:05.656 [info]  migrations are up to date
```
4. Create a new migration at priv/migrations/20210826171902_new .ex with:
```
defmodule AeMdw.Migrations.New do
  def run(), do: {:ok, nil}
end
```
5. `mix migrate_db`
It applies the new migration:
```
11:10:43.194 [info]  applying version 20210826171902 with Elixir.AeMdw.Migrations.New...

11:10:43.207 [info]  applied version 20210826171902
```
6. `mix migrate_db`
Migration is already is last version:
```
11:12:50.312 [info]  current migration version: 20210826171902

11:12:50.322 [info]  migrations are up to date
```

## Additional Notes

- For safety/consistency it is still important that each migration is idempotent (in case it aborts half way).
- Rollbacks are not available.

### DIFF
--- a/lib/mix/tasks/migrate_db.ex
+++ b/lib/mix/tasks/migrate_db.ex
@@ -2,22 +2,16 @@ defmodule Mix.Tasks.MigrateDb do
   use Mix.Task
 
   require Ex2ms
-  require Record
 
   alias AeMdw.Db.Model
   alias AeMdw.Db.Util
   alias AeMdw.Log
 
-  import Record, only: [defrecord: 2]
-
-  # version is like 20210826171900 in 20210826171900_reindex_remote_logs.ex
-  @record_name :migrations
-  @defaults [version: -1, inserted_at: nil]
-
-  defrecord @record_name, @defaults
-
-  @fields for {k,_v} <- @defaults, do: k
+  @table Model.Migrations
+  @record_name Model.record(@table)
   @version_len 14
+
+  @migrations_code_path "priv/migrations/*.ex"
 
   def run(_) do
     :net_kernel.start([:aeternity@localhost, :shortnames])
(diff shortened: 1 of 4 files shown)


jyeshe:
Working with first commit: 29f254d
Last 2c96f71 still to be fixed with RocksDB.

lib/mix/tasks/migrate_db.ex
```diff
@@ -0,0 +1,80 @@
...
+    |> Path.wildcard()
+    |> Enum.map(fn path ->
+      version =
+        path
+        |> Path.basename()
+        |> String.slice(0..(@version_len - 1))
+
+      {String.to_integer(version), path}
+    end)
+    |> Enum.sort_by(fn {version, _path} -> version end)
```
sborrazas, line 60: Perhaps we can do 
```elixir
|> Enum.drop_while(fn {version, _path} -> version < current_version end)
```
 instead of doing `if version > current_version do` below?

lib/mix/tasks/migrate_db.ex
```diff
@@ -0,0 +1,80 @@
...
+    applied_count =
+      @migrations_code_path
+      |> list_migrations_modules()
+      |> Enum.map(&maybe_apply_migration(&1, current_version))
+      |> Enum.count(&(&1))
+
+    # assure filesystem sync
+    if applied_count > 0, do: :mnesia.dump_log()
+
+    System.stop(0)
```
sborrazas, line 35: This isn't really needed, it's the default

lib/mix/tasks/migrate_db.ex
```diff
@@ -0,0 +1,80 @@
...
+  @migrations_code_path "priv/migrations/*.ex"
+
+  @impl Mix.Task
+  def run(_) do
+    :net_kernel.start([:aeternity@localhost, :shortnames])
+    :ae_plugin_utils.start_aecore()
+    :lager.set_loglevel(:epoch_sync_lager_event, :lager_console_backend, :undefined, :error)
+    :lager.set_loglevel(:lager_console_backend, :error)
+    Log.info("================================================================================")
+    current_version = read_migration_version(@table)
```
sborrazas, line 23: `@table` is accessible everywhere, why pass it around?

sborrazas (changes requested):

lib/mix/tasks/migrate_db.ex
jyeshe: Yes, let's filter. Ecto only shows that is up to date.

lib/mix/tasks/migrate_db.ex
jyeshe: The reset_db had it. Supposedly would help stopping the plugin but indeed without it console comes fast too.

lib/mix/tasks/migrate_db.ex
jyeshe: I usually prefer making it explicit what the functions is depending on but good `read_migration_version` was enough.

thepiwo (commented):
Looks good to me by just having a quick look. Is there some caveats to consider when we might have heavy migration that take minutes or hours to complete?

sborrazas (approved):

jyeshe:
> Looks good to me by just having a quick look. Is there some caveats to consider when we might have heavy migration that take minutes or hours to complete?

There are two concerns. One we could avoid here, not creating a transaction that would bloat the write log. The other is needed to be handled by each migration like AeMdw.Migrations.ReindexRemoteLogs that reads and writes in chunks.

thepiwo:
good for me, @wj2016 what do you think?

wj2016 (approved):

Issue #220, opened 2021-08-23
make unit/integration tests independent from synced chain


thepiwo:

sborrazas:
So after some digging into the different transactions we have, I thought it would be useful to have some declarative way of specifying a sample blockchain. This would not only make tests independent from the synced chain, but it will also make them clearer.

There's already been some talking about something similar for a thing called "chain simulator" on [this GH issue](https://github.com/aeternity/aeternity/issues/3693). In the future, it can not only be an Elixir-only thing, but also turned into some DSL that the chain simulator interpreter can run.

For writing Elixir tests, using exclusively Elixir data structres, I thought of having something like this:

```elixir
initial_balances = %{alice: 20_000, charlie: 30_000}

with_blockchain initial_balances, [ # with_blockchain would be a macro
  {:b1, {:kb, :alice}},
  {:b2, {:mb, :b2, [
      {:t1, {:spend_tx, :alice, :bob, 10_000}},
      {:t2, {:oracle_register, :bob, <<>>, <<>>}}, # Query format, response format
    ]}},
  {:b3, {:mb, :b3, [
      {:t3, {:contract_register, :charlie, "./contract-source.aes"}}
    ]}}
  {:b4, {:kb, :charlie}}
] do
  # 3 locally accessible variables
  #  - blocks       => map with info about blocks previously declared
  #  - accounts     => map with info about accounts previously declared
  #  - transactions => map with info about transactions previously declared

  # SAMPLE TEST 1
  block3 = blocks[:b3]
  block3_hash = block3[:hash]
  block3_miner = block3[:miner]

  assert %{"hash" => ^block3_hash, "miner" => ^block3_miner} = get(conn, "/block/#{block3_hash}")
    |> json_response(200)

  # SAMPLE TEST 2
  oracle_hash = transactions[:t2][:hash]

  assert %{"data" => [%{"hash" => ^oracle_hash}]} = get(conn, "/oracles?direction=backwards&limit=1")
    |> json_response(200)

  # SAMPLE TEST 3
  bob = accounts[:bob]
  bob_pubkey = bob[:pkey]

  assert %{"data" => [%{"amount" => 10_000}]} = get(conn, "/accounts/#{bob_pubkey}")
    |> json_response(200)
end
```

Since there's no way right now to create new blocks without mining and consensus, I thought this initial implementation would simply add the blocks to mnesia directly, with random hashes that don't have any validation.

[...]

PR #235, opened 2021-09-02
test: update NameController tests to be unit tests

sborrazas:
Didn't need to use `with_blockchain/2` because the NameController exclusively depends on the mdw database.

Refs #220

### DIFF
--- a/test/ae_mdw_web/controllers/name_controller_test.exs
+++ b/test/ae_mdw_web/controllers/name_controller_test.exs
@@ -1,0 +1,575 @@
+defmodule AeMdwWeb.NameControllerTest do
+  use AeMdwWeb.ConnCase
+
+  alias AeMdw.EtsCache
+  alias AeMdw.Validate
+  alias AeMdw.Db.{Model, Name, Util}
+  alias AeMdw.Db.Stream.Name, as: StreamName
+
+  import Mock
+
+  require AeMdw.Db.Model
+
+  @moduletag :integration
+
+  @default_limit 10
+
+  describe "active_names" do
+    test "get active names with default limit", %{conn: conn} do
+      sample_data = %{"foo" => "bar"}
+      response_data = for _i <- 1..@default_limit, do: sample_data
+
+      with_mocks [
+        {Util, [:passthrough], last_gen: fn -> 20 end},
+        {EtsCache, [],
+         [
+           get: fn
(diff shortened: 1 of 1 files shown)


jyeshe (commented):
Could we keep the not mocked ones or add integration tests for the protocol please? For those not passing that are hard to fix we could `@skip` them. 
IMHO the controller tests could be mocked, to focus testing on input and output (parameter validation and formating response) but before that it would be good to have protocol unit tests as integration tests. In short, protocol real tests, controller mocked ones.

sborrazas:
@jyeshe I'm assuming by protocol you mean `AeMdw.*` modules, right? If that's the case then I agree, we should have tests for those. These should actually hit mnesia records that are inserted before the test runs, so they still shouldn't depend on a synced blockchain.

I think a lot of simplification can be done while creating unit tests under the middleware library (`AeMdw.*`). Meanwhile, we should probably keep mocking tests under the web modules (`AeMdwWeb.*`) like this PR does. Creating tests under `AeMdw.*` will probably take a couple of months.

jyeshe:
> @jyeshe I'm assuming by protocol you mean `AeMdw.*` modules, right? If that's the case then I agree, we should have tests for those. These should actually hit mnesia records that are inserted before the test runs, so they still shouldn't depend on a synced blockchain.
> 
> I think a lot of simplification can be done while creating unit tests under the middleware library (`AeMdw.*`). Meanwhile, we should probably keep mocking tests under the web modules (`AeMdwWeb.*`) like this PR does. Creating tests under `AeMdw.*` will probably take a couple of months.

Sorry, by protocol I mean chain, info that comes and goes from/to the chain nodes. 
I agree about the unit tests so we could leave the integration tests, as they are validating the interaction with the chain, and add new mocked tests to increase the coverage of unit tests.

thepiwo:
@Artur64 can you review this?

jyeshe (approved):

jyeshe (changes requested):
Could we keep the integration as they are and add unit tests?

sborrazas:
@jyeshe for some reason I changed the integration test instead of adding it to the unit `name_controller_test`. It's fixed now!

jyeshe (approved):
Tests passed

thepiwo (approved):

PR #237, opened 2021-09-03
feat: adds account and recipient name to spendtx

### LINKED ISSUE #226: more SpendTx details on mdw/txs/backward?account=...
When calling this endpoint the transactions of an account are returned.

This is a feature request to return more details for tx-type `SpendTx` if the recipient is a name id (`nm_...`). I would like to see following additional data returned in this case:
- name (`xyz.name`)
- account (`ak_...`)

jyeshe:
## What

Adds some details to /txs/forward and /txs/backward endpoints when using account as querystring param.
The details are account and recipient chain name.

Note: As the names have expiration and might be tranfered, the recipient account is the owner of the name at the time of the transaction.

## Why

Extra info/fixes #226 

## Validation steps

```
elixir --sname aeternity@localhost -S mix test.integration test/integration/ae_mdw_web/controllers/tx_controller_test.exs:583
elixir --sname aeternity@localhost -S mix test.integration test/integration/ae_mdw_web/controllers/tx_controller_test.exs:661
elixir --sname aeternity@localhost -S mix test.integration test/integration/ae_mdw_web/controllers/tx_controller_test.exs 
```

### DIFF
--- a/lib/ae_mdw_web/controllers/tx_controller.ex
+++ b/lib/ae_mdw_web/controllers/tx_controller.ex
@@ -15,6 +15,8 @@ defmodule AeMdwWeb.TxController do
   import AeMdwWeb.Util
   import AeMdw.Db.Util
 
+  @type_spend_tx "SpendTx"
+
   ##########
 
   def tx(conn, %{"hash" => enc_tx_hash}),
@@ -23,8 +25,8 @@ defmodule AeMdwWeb.TxController do
   def txi(conn, %{"index" => index}),
     do: handle_tx_reply(conn, fn -> read_tx(Validate.nonneg_int!(index)) end)
 
-  def txs(conn, _req),
-    do: Cont.response(conn, &json/2)
+  def txs(conn, params),
+    do: Cont.response(conn, &json/2, &add_spendtx_details(&1, params))
 
   def count(conn, _req),
     do: conn |> json(last_txi())
@@ -58,7 +60,28 @@ defmodule AeMdwWeb.TxController do
     end
   end
 
-  def read_tx_hash(tx_hash) do
+  #
+  # Private functions
+  #
+  defp add_spendtx_details(data_txs, %{"account" => account}) do
+    Enum.map(data_txs, fn block ->
+      spend_tx_recipient =
+        if block["tx"]["type"] == @type_spend_tx, do: block["tx"]["recipient_id"]
+
+      if nil != spend_tx_recipient and String.slice(spend_tx_recipient, 0..2) == "nm_" do
+        case Validate.plain_name(spend_tx_recipient) do
+          {:ok, name} ->
+            Map.merge(block, %{"name" => name, "account" => account})
+        end
+      else
+        block
+      end
+    end)
+  end
+
+  defp add_spendtx_details(data_txs, _), do: data_txs
+
+  defp read_tx_hash(tx_hash) do
     with <<_::256>> = mb_hash <- :aec_db.find_tx_location(tx_hash),
          {:ok, mb_header} <- :aec_chain.get_header(mb_hash),
          height <- :aec_headers.height(mb_header) do
--- a/lib/ae_mdw_web/controllers/tx_controller.ex
+++ b/lib/ae_mdw_web/controllers/tx_controller.ex
@@ -69,41 +69,46 @@ defmodule AeMdwWeb.TxController do
   # Private functions
   #
   defp add_spendtx_details(response_data, %{"account" => _account}) do
-    Enum.map(response_data, fn block ->
-      spend_tx_recipient =
-        if block["tx"]["type"] == @type_spend_tx, do: block["tx"]["recipient_id"]
+    Enum.map(response_data, fn %{"block_height" => block_height, "tx" => block_tx} = block ->
+      spend_tx_recipient = (block_tx["type"] == @type_spend_tx && block_tx["recipient_id"]) || nil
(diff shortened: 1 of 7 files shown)


lib/ae_mdw_web/controllers/tx_controller.ex
```diff
@@ -58,7 +61,32 @@ defmodule AeMdwWeb.TxController do
     end
   end
 
-  def read_tx_hash(tx_hash) do
+  #
+  # Private functions
+  #
+  defp add_spendtx_details(data_txs, %{"account" => account}) do
+    Enum.map(data_txs, fn block ->
```
sborrazas, line 68: We're iterating transactions (`data_txs`) but we're naming each of this transactions `block`. So, are they transactions or blocks?

lib/ae_mdw_web/controllers/tx_controller.ex
```diff
@@ -58,7 +61,32 @@ defmodule AeMdwWeb.TxController do
...
+      spend_tx_recipient =
+        if block["tx"]["type"] == @type_spend_tx, do: block["tx"]["recipient_id"]
```
sborrazas, line 70: Sometimes we're using the `(a && b) || c` notation (on the previous file) and some others we're using `if a, do: b, else: c`, we should probably stick to one

lib/ae_mdw_web/controllers/tx_controller.ex
```diff
@@ -58,7 +61,32 @@ defmodule AeMdwWeb.TxController do
...
+
+      if nil != spend_tx_recipient and String.slice(spend_tx_recipient, 0..2) == "nm_" do
+        case Validate.plain_name(spend_tx_recipient) do
+          {:ok, name} ->
+            Map.merge(block, %{"name" => name, "account" => account})
+
+          _ ->
```
sborrazas, line 77: I think unused variables are more useful to the reader if we explicitly name them what they are supposed to be (e.g. `{:error, _reason}` here and `add_spendtx_details(data_txs, _params)` down below.

lib/ae_mdw_web/controllers/tx_controller.ex
```diff
@@ -58,7 +61,32 @@ defmodule AeMdwWeb.TxController do
...
+        end
+      else
+        block
+      end
+    end)
+  end
+
+  defp add_spendtx_details(data_txs, _), do: data_txs
+
+  defp read_tx_hash(tx_hash) do
```
sborrazas, line 89: Thanks for making all of these private, we should keep at it! :pray:

sborrazas (changes requested):

lib/ae_mdw_web/controllers/tx_controller.ex
jyeshe: they are blocks with transactions with. I will call bock_txs.

lib/ae_mdw_web/controllers/tx_controller.ex
jyeshe: Yes, for legacy I am keeping the `(a && b) || c` style but for new code and specially here without an `else` was the `if cond, do:`

lib/ae_mdw_web/controllers/tx_controller.ex
jyeshe: and renamed data_txs to response_data

lib/ae_mdw_web/controllers/tx_controller.ex
sborrazas: The `(a && b) || c` code above is new code, that line was added by you

jyeshe: I mean when creating a new function. Without `else` I'd rather stick with`if cond, do:` rather than `(a && b) || nil`.
Nevertheless, some people dislike this implicit `nil` return in Elixir. I am okay with both approaches. We discuss with everyone in the meeting.

lib/ae_mdw_web/controllers/tx_controller.ex
```diff
@@ -58,7 +61,32 @@ defmodule AeMdwWeb.TxController do
...
-  def read_tx_hash(tx_hash) do
+  #
+  # Private functions
+  #
+  defp add_spendtx_details(response_data, %{"account" => account}) do
+    Enum.map(response_data, fn block ->
+      spend_tx_recipient =
+        if block["tx"]["type"] == @type_spend_tx, do: block["tx"]["recipient_id"]
+
+      if nil != spend_tx_recipient and String.slice(spend_tx_recipient, 0..2) == "nm_" do
```
thepiwo, line 72: can't we use the id system from the aeternity node here for getting the name account instead of string manipulation?

thepiwo (commented):
This is a non-breaking change now, right?

lib/ae_mdw_web/controllers/tx_controller.ex
jyeshe: Good point, thanks

jyeshe: DBS.map is widely used. Better continue using it's return as is (with encoded human readable recipient_id) comparing with `nm_`.

lib/ae_mdw_web/controllers/tx_controller.ex
jyeshe: used (a && b) || nil until next refactoring

sborrazas (approved):

thepiwo (approved):

PR #238, opened 2021-09-03
test: move integration tests to a separate directory

sborrazas:
This allows us to focus solely on unit tests, plus, we can tag
integration tests using @moduletag on the module instead of each test
case.


jyeshe (approved):
Tests passing as before (only 1 expected failure) :clap:

wj2016 (approved):

PR #241, opened 2021-09-08
feat: restructure ETS stateful DB streams implementation

sborrazas:
Because we're having problems before with dealing with instance state by using ETS to store the remaining streams, a substantial change was necessary to both get rid of streams and to simplify the way the MDW works by using a more encapsulated and layered coding style.

Changes introduced:

1. **Refactor MDW domain logic to use a layered coding style.**

   Before this change, there weren't really any layers, the `AeMdw`, `AeMdwWeb` and `AeMdwWeb.Db` modules went back and forth calling each other out.

   Now, the layer model consists of the following:

   1. `AeMdwWeb.<Resource>Controller` (e.g. `OracleController`) only deals with getting the request params and displaying the returned data properly.
   2. `AeMdw.<Resource>` (e.g. Oracles) deals with all of the logic related to oracles, which right now includes fetching and
      serializing.
   3. `AeMdw.Mnesia` is just a wrapper around the `:mnesia` operations on the middleware tables.

   Each of these layers can only talk to the layer below, but never the other way round.

2. **Stop using ETS for storing the remaining slices of streams when dealing with pagination. Because of this, a greater refactor such as the one mentioned in 1. was needed.**

   Before this change, in order to get pagination working the entire  list of records from the database was represented as a
   stream. Whenever the user requested for any of these streams, the first few elements of the stream were returned, while the rest of the stream was stored in the instance ETS table, where the key was made out of the pagination options of that particular user for that list of records.

   Because of this, in order to traverse a full list of records doing several requests, the user had to the same server instance throughout all of the pages visited. It is because of this that a load balancer would break pagination by performing requests from the same user to different instances which had different ETS instance state.

By getting rid of streams and no longer storing state in ETS tables, the whole application now becomes stateless, and can be scaled horizontally indefinitely.

This is a PoC

### DIFF
--- a/lib/ae_mdw/mnesia.ex
+++ b/lib/ae_mdw/mnesia.ex
@@ -1,0 +1,139 @@
+defmodule AeMdw.Mnesia do
+  @moduledoc """
+  Mnesia wrapper to provide a simpler API.
+
+  In order to iterate throught collections of records we use the
+  concept of "cursor". A cursor is just a key of any given table that
+  is returned in order the access the next page of records.
+
+  This cursor key is whatever the current key of the table might be
+  (e.g. a tuple, a number). If there is no next page, `nil` is
+  returned instead.
+  """
+
+  @type table() :: atom()
+  @type record() :: tuple()
+  @type key() :: term()
+  @type direction() :: :forward | :backward
+  @type cursor() :: key() | nil
+
+  @end_token :"$end_of_table"
+
+  @spec last_key(table()) :: {:ok, key()} | :none
+  def last_key(tab) do
+    case :mnesia.dirty_last(tab) do
+      @end_token -> :none
+      last_key -> {:ok, last_key}
+    end
+  end
+
+  @spec first_key(table()) :: {:ok, key()} | :none
+  def first_key(tab) do
+    case :mnesia.dirty_first(tab) do
+      @end_token -> :none
+      last_key -> {:ok, last_key}
+    end
+  end
+
+  @spec fetch(table(), key()) :: {:ok, record()} | :not_found
+  def fetch(tab, key) do
+    case :mnesia.dirty_read(tab, key) do
+      [record] -> {:ok, record}
+      [] -> :not_found
+    end
+  end
+
+  @spec fetch!(table(), key()) :: record()
+  def fetch!(tab, key) do
+    {:ok, record} = fetch(tab, key)
+
+    record
+  end
+
+  @spec fetch_keys(table(), direction(), cursor(), non_neg_integer()) :: {[record()], cursor()}
+  def fetch_keys(tab, :forward, nil, limit) do
+    {:atomic, {keys, cursor}} =
+      :mnesia.transaction(fn ->
+        case :mnesia.first(tab) do
+          @end_token -> {[], nil}
+          first_key -> fetch_forward_keys(tab, first_key, limit)
+        end
+      end)
+
+    {keys, cursor}
+  end
+
+  def fetch_keys(tab, :forward, first_key, limit) do
+    {:atomic, {keys, cursor}} =
+      :mnesia.transaction(fn ->
+        case :mnesia.read(tab, first_key) do
+          @end_token -> {[], nil}
+          _first_record -> fetch_forward_keys(tab, first_key, limit)
+        end
+      end)
+
+    {keys, cursor}
+  end
+
+  def fetch_keys(tab, :backward, nil, limit) do
+    {:atomic, {keys, cursor}} =
+      :mnesia.transaction(fn ->
+        case :mnesia.last(tab) do
+          @end_token -> {[], nil}
+          last_key -> fetch_backward_keys(tab, last_key, limit)
+        end
+      end)
+
+    {keys, cursor}
+  end
+
+  def fetch_keys(tab, :backward, last_key, limit) do
+    {:atomic, {keys, cursor}} =
+      :mnesia.transaction(fn ->
+        case :mnesia.read(tab, last_key) do
+          @end_token -> {[], nil}
+          _last_record -> fetch_backward_keys(tab, last_key, limit)
+        end
+      end)
+
+    {keys, cursor}
+  end
+
+  defp fetch_forward_keys(tab, first_key, limit) do
+    keys =
+      Stream.unfold({limit, first_key}, fn
+        {0, _current_key} ->
+          nil
+
+        {n, current_key} ->
+          case :mnesia.next(tab, current_key) do
+            @end_token -> nil
+            next_key -> {next_key, {n - 1, next_key}}
+          end
+      end)
+
+    case Enum.split(keys, limit - 1) do
+      {keys, [next_key]} -> {[first_key | keys], next_key}
+      {keys, []} -> {keys, nil}
+    end
+  end
+
+  defp fetch_backward_keys(tab, last_key, limit) do
+    keys =
+      Stream.unfold({limit, last_key}, fn
+        {0, _current_key} ->
+          nil
+
+        {n, current_key} ->
+          case :mnesia.prev(tab, current_key) do
+            @end_token -> nil
+            prev_key -> {prev_key, {n - 1, prev_key}}
+          end
+      end)
+
+    case Enum.split(keys, limit - 1) do
+      {keys, [prev_key]} -> {[last_key | keys], prev_key}
+      {keys, []} -> {keys, nil}
+    end
+  end
+end
--- a/lib/ae_mdw/mnesia.ex
+++ b/lib/ae_mdw/mnesia.ex
@@ -114,7 +114,7 @@ defmodule AeMdw.Mnesia do
 
     case Enum.split(keys, limit - 1) do
       {keys, [next_key]} -> {[first_key | keys], next_key}
-      {keys, []} -> {keys, nil}
+      {keys, []} -> {[first_key | keys], nil}
     end
   end
 
@@ -133,7 +133,7 @@ defmodule AeMdw.Mnesia do
 
     case Enum.split(keys, limit - 1) do
       {keys, [prev_key]} -> {[last_key | keys], prev_key}
-      {keys, []} -> {keys, nil}
+      {keys, []} -> {[last_key | keys], nil}
     end
   end
 end
--- a/lib/ae_mdw/mnesia.ex
+++ b/lib/ae_mdw/mnesia.ex
@@ -52,8 +52,8 @@ defmodule AeMdw.Mnesia do
 
   @spec fetch_keys(table(), direction(), cursor(), non_neg_integer()) :: {[record()], cursor()}
   def fetch_keys(tab, :forward, nil, limit) do
-    {:atomic, {keys, cursor}} =
-      :mnesia.transaction(fn ->
+    {keys, cursor} =
+      :mnesia.async_dirty(fn ->
         case :mnesia.first(tab) do
           @end_token -> {[], nil}
           first_key -> fetch_forward_keys(tab, first_key, limit)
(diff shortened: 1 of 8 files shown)


lib/ae_mdw/oracles.ex
```diff
@@ -0,0 +1,118 @@
...
+
+    {:ok, {last_gen, -1}} = Mnesia.last_key(AeMdw.Db.Model.Block)
+
+    oracles =
+      exp_keys
+      |> Enum.map(fn {_expiration, oracle_pk} -> fetch_active_oracle(oracle_pk) end)
+      |> serialize_list(last_gen, true)
+
+    {oracles, serialize_cursor(next_cursor)}
+  end
```
jyeshe: beautiful

lib/ae_mdw/oracles.ex
```diff
@@ -0,0 +1,118 @@
...
+        deserialize_cursor(cursor),
+        @limit
+      )
+
+    {:ok, {last_gen, -1}} = Mnesia.last_key(AeMdw.Db.Model.Block)
+
+    oracles =
+      exp_keys
+      |> Enum.map(fn {_expiration, oracle_pk} -> fetch_inactive_oracle(oracle_pk) end)
+      |> serialize_list(last_gen, false)
```
jyeshe, line 59: nitpick: to use serialize terminology for cursors and render for maps to be encoded as json

lib/ae_mdw_web/controllers/oracle_controller.ex
```diff
@@ -51,6 +57,42 @@ defmodule AeMdwWeb.OracleController do
   def oracles(conn, _req),
     do: handle_input(conn, fn -> Cont.response(conn, &json/2) end)
 
+  def active_oracles_v2(%Conn{assigns: assigns} = conn, _params) do
+    direction = Map.get(assigns, :direction, :backward)
```
jyeshe, line 61: pls add a comment on why using :backward as default

jyeshe (commented):
Haven't tested it thoroughly and the coverage is still in progress but code looks very clean and straightforward. 
There is only flag that might be operational, the others are only minors.

lib/ae_mdw/mnesia.ex
```diff
@@ -0,0 +1,91 @@
+defmodule AeMdw.Mnesia do
+  @moduledoc """
+  Mnesia wrapper to provide a simpler API.
+  """
+
+  @type table() :: atom()
+  @type record() :: tuple()
+  @type key() :: term()
+  @type direction() :: :forward | :backward
+  @type cursor() :: key() | nil
```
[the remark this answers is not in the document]
sborrazas: These types are the params and return value types, so it made more sense to make them public

lib/ae_mdw/mnesia.ex
```diff
@@ -0,0 +1,91 @@
+defmodule AeMdw.Mnesia do
+  @moduledoc """
+  Mnesia wrapper to provide a simpler API.
```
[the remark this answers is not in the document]
sborrazas: Done

lib/ae_mdw/mnesia.ex
```diff
@@ -0,0 +1,91 @@
...
+    {:ok, record} = fetch(tab, key)
+
+    record
+  end
+
+  @spec fetch_keys(table(), direction(), cursor(), non_neg_integer()) :: {[record()], cursor()}
+  def fetch_keys(tab, :forward, nil, limit) do
+    case :mnesia.dirty_first(tab) do
+      @end_token -> {[], nil}
+      first_key -> fetch_keys(tab, :forward, first_key, limit)
```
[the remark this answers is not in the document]
sborrazas: Yeah, I might rethink how this is done. Working on it

lib/ae_mdw/mnesia.ex
```diff
@@ -0,0 +1,91 @@
...
+    case :mnesia.dirty_next(tab, current_key) do
+      @end_token -> {acc, nil}
+      next_key -> {acc, next_key}
+    end
+  end
+
+  defp fetch_forward_keys(tab, current_key, limit, acc) do
+    case :mnesia.dirty_next(tab, current_key) do
+      @end_token -> {acc, nil}
+      next_key -> fetch_forward_keys(tab, next_key, limit - 1, [next_key | acc])
```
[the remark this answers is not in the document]
sborrazas: I 100% agree we could use `:mnesia.select` instead. But for now I let it the same way it was done before because I didn't want to change much or break anything. But this should be done as a subsequent implementation. The good thing is that this will only need to be changed in a single place

lib/ae_mdw/oracles.ex
sborrazas: Good call, `render` for user-friendly structures and `serialize` for encoding/decoding into other structures

lib/ae_mdw_web/controllers/oracle_controller.ex
sborrazas: Done

lib/ae_mdw/mnesia.ex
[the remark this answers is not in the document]
jyeshe: nitpick that a part of `cursor()` they aren't used in specs outside of the module but I am ok to leave to be possibly used in other future context modules.

jyeshe (commented):
Some suggestions on testing, stability and parameter/contract analysis (regarding the nil).

lib/ae_mdw/mnesia.ex
```diff
@@ -0,0 +1,107 @@
...
+  def fetch!(tab, key) do
+    {:ok, record} = fetch(tab, key)
+
+    record
+  end
+
+  @spec fetch_keys(table(), direction(), cursor(), non_neg_integer()) :: {[record()], cursor()}
+  def fetch_keys(tab, :forward, nil, limit) do
+    case first_key(tab) do
+      {:ok, first_key} -> fetch_keys(tab, :forward, first_key, limit)
```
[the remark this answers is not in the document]
sborrazas: Yes, it will iterate until limit is 1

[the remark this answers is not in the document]
jyeshe: I was wondering if after last page we should return no records but the client shall be smart enough for that.
Makes sense, no cursor, initialize one.

lib/ae_mdw/oracles.ex
```diff
@@ -0,0 +1,118 @@
...
+  @limit 10
+
+  @table_active AeMdw.Db.Model.ActiveOracle
+  @table_active_expiration Model.ActiveOracleExpiration
+  @table_inactive AeMdw.Db.Model.InactiveOracle
+  @table_inactive_expiration Model.InactiveOracleExpiration
+
+  @spec fetch_active_oracles(Mnesia.direction(), cursor() | nil) :: {[oracle()], cursor() | nil}
+  def fetch_active_oracles(direction, cursor) do
+    {exp_keys, next_cursor} =
```
wj2016, line 26: Does this work for other requests as well? For example, if the client wants to get /txs, it may have to read from different tables.

lib/ae_mdw/mnesia.ex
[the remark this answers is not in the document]
sborrazas: Done, now I'm including every mnesia query inside a transaction, and avoided the recursion by using `Stream.unfold/2`

lib/ae_mdw/mnesia.ex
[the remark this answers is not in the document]
sborrazas: Done, included everything inside a single transaction

test/integration/ae_mdw_web/controllers/oracle_controller_test.exs
```diff
@@ -161,6 +161,48 @@ defmodule Integration.AeMdwWeb.OracleControllerTest do
     end
   end
 
+  describe "oracles_v2" do
+    test "it fetches inactive oracles going forwards", %{conn: conn} do
+      %{
```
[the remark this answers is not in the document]
sborrazas: Had forgotten about it, added them now. thanks

test/integration/ae_mdw_web/controllers/oracle_controller_test.exs
```diff
@@ -161,6 +161,48 @@ defmodule Integration.AeMdwWeb.OracleControllerTest do
     end
   end
 
+  describe "oracles_v2" do
```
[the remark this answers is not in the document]
sborrazas: Added test cases that map the other integration tests we have for the OracleController, and they match 100% in output

lib/ae_mdw/oracles.ex
sborrazas: The idea is that this `Oracles` module will only contain oracle-related logic, if there's anything in common with other context functions that might appear in the future we can think of extracting them into something else. Does that sound good?

sborrazas:
Ran some benchmarks on both versions of the oracles endpoint and got a significant improvement in terms of speed (1.26x) and memory (also 1.26x):

Screenshot from 2021-09-09 16-12-19

Ran some tracing as well to count the amount of times the `:mnesia` module was being called, though not significant, got an improvement there as well:
 
```
$ echo ":mnesia V1 calls: $(grep ' calls :mnesia' tracing_output.v1.log | wc -l)"
:mnesia V1 calls: 535
$ echo ":mnesia V2 calls: $(grep ' calls :mnesia' tracing_output.v2.log | wc -l)"
:mnesia V2 calls: 558
```

In terms of output, also compared between the two with the integration tests that I added and they do match (for both active and inactive oracle endpoints).

lib/ae_mdw/mnesia.ex
```diff
@@ -0,0 +1,139 @@
...
+  def fetch!(tab, key) do
+    {:ok, record} = fetch(tab, key)
+
+    record
+  end
+
+  @spec fetch_keys(table(), direction(), cursor(), non_neg_integer()) :: {[record()], cursor()}
+  def fetch_keys(tab, :forward, nil, limit) do
+    {:atomic, {keys, cursor}} =
+      :mnesia.transaction(fn ->
```
jyeshe, line 56: Took off the hat for the solution of nested contexts. Nice use of `Stream.unfold` :clap: 
Only here the mnesia context could continue to be dirty for lighter fetching with `async_dirty` like in `AeMdw.Db.Util`.

lib/ae_mdw/mnesia.ex
```diff
@@ -0,0 +1,139 @@
...
+          _last_record -> fetch_backward_keys(tab, last_key, limit)
+        end
+      end)
+
+    {keys, cursor}
+  end
+
+  defp fetch_forward_keys(tab, first_key, limit) do
+    keys =
+      Stream.unfold({limit, first_key}, fn
```
jyeshe, line 104: :100:

lib/ae_mdw/mnesia.ex
```diff
@@ -0,0 +1,139 @@
...
+
+        {n, current_key} ->
+          case :mnesia.prev(tab, current_key) do
+            @end_token -> nil
+            prev_key -> {prev_key, {n - 1, prev_key}}
+          end
+      end)
+
+    case Enum.split(keys, limit - 1) do
+      {keys, [prev_key]} -> {[last_key | keys], prev_key}
```
jyeshe, line 135: considering limit = 10, keys has 9 elements after `Enum.split`

lib/ae_mdw/mnesia.ex
sborrazas: Done. With it I reduced the memory usage, but increased the time just a slightly bit. But I think async_dirty/1 is better anyway, nice catch!

Screenshot from 2021-09-10 21-09-22

lib/ae_mdw/mnesia.ex
sborrazas: `keys` will have 9 elements, while `[last_key | key]` will have 10, as requested, so we're good

jyeshe (approved):
Looks good and test passes.

lib/ae_mdw/mnesia.ex
jyeshe: Wondering if `:mnesia.sync_dirty` would be faster for reading (empty log to write) as the `:mnesia.transaction` was faster than `:mnesia.async_dirty`.

sborrazas: Using `:mnesia.sync_dirty/1` reports exactly as much memory as before and doesn't really change much in terms of timing. So I think I'm leaving `async_dirty`

thepiwo (approved):

PR #242, opened 2021-09-08
fix: recipient account is the pointee if name have one

### LINKED ISSUE #226: more SpendTx details on mdw/txs/backward?account=...

jyeshe:
## What

Replace name transfer search with update search. 
When the recipient id is an AENS id name (`nm_...`), the recipient account is the pointee of the name (if name have one). 

## Why
Fixes #226 
The recipient of the name used might not be owner of the name, or in other words, the name might point to an account different from the owner.

## Validate steps

```
elixir --sname aeternity@localhost -S mix test.integration test/integration/ae_mdw_web/controllers/tx_controller_test.exs:582
elixir --sname aeternity@localhost -S mix test.integration test/integration/ae_mdw_web/controllers/tx_controller_test.exs
```

### DIFF
--- a/lib/ae_mdw/db/name.ex
+++ b/lib/ae_mdw/db/name.ex
@@ -1,5 +1,4 @@
 defmodule AeMdw.Db.Name do
-  alias AeMdw.Log
   alias AeMdw.Node, as: AE
   alias AeMdw.Db.{Model, Format}
   alias AeMdw.Validate
@@ -75,18 +74,10 @@ defmodule AeMdw.Db.Name do
   def account_pointer_at(plain_name, time_reference_txi) do
     case locate(plain_name) do
       {nil, _module} ->
-        Log.warn("missing name for plain name #{plain_name}")
         {:error, :name_not_found}
 
       {m_name, _module} ->
-        case pointee_at(m_name, time_reference_txi) do
-          nil ->
-            Log.warn("missing pointee for plain name #{plain_name} on txi #{time_reference_txi}")
-            {:error, :pointee_not_found}
-
-          pointee_pk ->
-            {:ok, pointee_pk}
-        end
+        pointee_at(m_name, time_reference_txi)
     end
   end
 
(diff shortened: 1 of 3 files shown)


lib/ae_mdw/db/name.ex
```diff
@@ -71,6 +72,24 @@ defmodule AeMdw.Db.Name do
...
   end
 
+  def account_pointer_at(plain_name, time_reference_txi) do
+    case locate(plain_name) do
+      {nil, _module} ->
+        Log.warn("missing name for plain name #{plain_name}")
+        {:error, :name_not_found}
+
+      {m_name, _module} ->
+        case pointee_at(m_name, time_reference_txi) do
```
sborrazas, line 82: Perhaps we can make `pontee_at` return `{:ok, pk} | {:error, :pointee_not_found}` and leave the call only in this clause

lib/ae_mdw/db/name.ex
```diff
@@ -233,16 +234,31 @@ defmodule AeMdw.Db.Name do
...
-    %{tx: %{account_id: orig_owner}} = Format.to_raw_map(read_tx!(last_claim_txi))
-    {:id, :account, first_owner} = orig_owner
-    first_owner
+  defp pointee_at(m_name, ref_txi) do
+    m_name
+    |> Model.name(:updates)
+    |> find_update_txi_before(ref_txi)
+    |> case do
+      nil ->
+        Log.warn("name #{elem(m_name, 1)} without pointer at txi #{ref_txi}!")
```
sborrazas, line 243: We can start pattern-matching on records instead of using `elem` or the getter functions like `Model.name(:updates)`.` This can be done like so:

```elixir
def pointee_at(Model.name(updates: updates, index: index), ref_txi)
```

This way you would get compiler errors if the field is wrongly spelled or doesn't exist, plus some added clarity.

sborrazas (dismissed):

lib/ae_mdw/db/name.ex
jyeshe: good idea, eliminates one case.

lib/ae_mdw/db/name.ex
jyeshe: I am pro pattern matching in many cases. This is a good one even though `Model.name/2` also detects typos.

sborrazas (approved):

thepiwo (approved):

PR #243, opened 2021-09-09
ci: add credo to ci

sborrazas:
Since this is the first time adding credo, when diffing it with master it will fail. But from now on it will only print any credo errors ADDED to the codebase, it won't complain about existing unchanged lines.


.github/workflows/test-action.yml
```diff
@@ -39,6 +39,11 @@ jobs:
       - name: Setup dependencies
         run: make docker-deps
 
+      - name: Reset master
```
thepiwo, line 42: why is that needed?

sborrazas: Right now credo only warns for the new errors added, not for the existing ones. To be able to do that it compares the errors with the errors in master. Github does not fetch master by default, so this is why we have to create it before credo runs

thepiwo (approved):

jyeshe (approved):

PR #248, opened 2021-09-15
feat: index inner transactions

### LINKED ISSUE #224: innerTx handling (PayingForTx, GAMeta)
Karol already started working on that back then:
- https://github.com/aeternity/ae_mdw/tree/inner-txs

jyeshe:
## What

This is a feature to index the information of the inner transactions from paying_for_tx and ga_meta_tx.
As tx hash of these inner transactions is not known by the user and the inner txi is the same as the outer, the updated endpoint is only the `/txs`.

## Why

Fixes #224
Inner txs should be searchable by fields from specific transactions (e.g `spender_id` and `recipient_id` for `SpendTx`).

## Validation steps

1. `http://localhost:4000/txs/forward?recipient_id=ak_2RUVa9bvHUD8wYSrvixRjy9LonA9L29wRvwDfQ4y37ysMKjgdQ` does not any GaMetaTx transaction (only one outer SpendTx).
2. `mix migrate_db` to index the inner transactions.
3. Repeat same request of step 1 and it returns the outer SpendTx along with (at least) 3 GaMetaTxs (with inner SpendTx).

Integration test:

image

## Additional Notes

`txi` for inner tx is a negative scope requirement.


lib/ae_mdw/db/sync/inner_tx.ex
```diff
@@ -0,0 +1,15 @@
...
+  """
+
+  @typep wrapper_type() :: :ga_meta_tx | :paying_for_tx
+
+  @spec signed_tx(wrapper_type(), tuple()) :: {atom(), tuple()}
+  def signed_tx(type, wrapper_tx) do
+    case type do
+      :ga_meta_tx -> :aega_meta_tx.tx(wrapper_tx)
+      :paying_for_tx -> :aec_paying_for_tx.tx(wrapper_tx)
+    end
```
sborrazas, line 13: You can patter-match here on the function signature
```elixir
def signed_tx(:ga_meta_tx, wrapper_tx), do: ...
```

lib/ae_mdw/db/sync/transaction.ex
```diff
@@ -133,79 +129,99 @@ defmodule AeMdw.Db.Sync.Transaction do
...
+    else
+      txi + 1
+    end
   end
 
-  ##########
+  #
+  # Private functions
+  #
+  def write_tx(type, txi, tx_hash, block_index, mb_time, wrapper_tx_id) do
```
sborrazas, line 144: This function isn't private

lib/ae_mdw_web/controllers/tx_controller.ex
```diff
@@ -112,14 +112,23 @@ defmodule AeMdwWeb.TxController do
     end
   end
 
-  defp read_tx_hash(tx_hash) do
+  defp read_tx_hash(tx_hash, wrapper_lookup? \\ false) do
```
sborrazas, line 115: From what it looks like this function is always going to be called with a single parameter, so the 2nd parameter will always be `false`. Perhaps we can get rid of it?

jyeshe: It's used to search for inner transaction. 
The sequence is to first try like an outer, most common case (by far). If doesn't find, looks up like an inner transaction.
We could create also a new endpoint for inner transactions or add an additional parameter but avoided because it will be less transparent with one more decision for the client (possibly implying in an additional request).

jyeshe: Not applicable anymore as the /tx/:hash is not part of the scope.

sborrazas (approved):

thepiwo (approved):
looks good, but haven't had the time to test it

jyeshe:
> looks good, but haven't had the time to test it

Thanks for the review. No problem, I've tested with curl and added integration tests. Gonna check closely also in testnet.

PR #252, opened 2021-09-17
style: fix credo alias warning for most files

sborrazas:
The purpose of this is to get rid of warning on subsequent PRs


jyeshe (approved):
The warnings won't go away easily but we can fix it incrementally with specific commits for that at the end of the PRs or with specific ones to solve the other credo complaints. Or you drink some redbull and tackle them all ((:

thepiwo (approved):
looks good, the linter will just fail for code added in the future, right?

PR #256, opened 2021-09-20
refactor: add Collection module to deal with complex pagination

sborrazas:
AwMdw.Collection functions are going to be reused all over the paginated results.

### DIFF
--- a/lib/ae_mdw/collection.ex
+++ b/lib/ae_mdw/collection.ex
@@ -22,7 +22,7 @@ defmodule AeMdw.Collection do
     iex> :mnesia.dirty_all_keys(:table2)
     [:c, :d, :e]
     iex> AeMdw.Collection.concat(:table1, :table2, :forward, nil, 3)
-    {{:table1, [:a, :b]}, {:table2, [:c, :d]}, :d}
+    {{:table1, [:a, :b]}, {:table2, [:c]}, :d}
     iex> AeMdw.Collection.concat(:table1, :table2, :backward, nil, 4)
     {{:table2, [:e, :d, :c, :b]}, {:table1, []}, :a}
     iex> AeMdw.Collection.concat(:table1, :table2, :forward, :d, 4)
--- a/lib/ae_mdw/collection.ex
+++ b/lib/ae_mdw/collection.ex
@@ -22,15 +22,15 @@ defmodule AeMdw.Collection do
     iex> :mnesia.dirty_all_keys(:table2)
     [:c, :d, :e]
     iex> AeMdw.Collection.concat(:table1, :table2, :forward, nil, 3)
-    {[{:a, :table1}, {:b, :table1}, {:c, :table1}], :d}
+    {{:table1, [:a, :b]}, {:table2, [:c, :d]}, :d}
     iex> AeMdw.Collection.concat(:table1, :table2, :backward, nil, 4)
-    {[{:e, :table2}, {:d, :table2}, {:c, :table2}, {:b, :table1}], :a}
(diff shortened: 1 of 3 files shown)


thepiwo (approved):

lib/ae_mdw/collection.ex
```diff
@@ -0,0 +1,69 @@
...
+    {[{:a, :table1}, {:b, :table1}, {:c, :table1}], :d}
+    iex> AeMdw.Collection.concat(:table1, :table2, :backward, nil, 4)
+    {[{:e, :table2}, {:d, :table2}, {:c, :table2}, {:b, :table1}], :a}
+    iex> AeMdw.Collection.concat(:table1, :table2, :forward, :d, 4)
+    {[{:d, :table2}, {:e, :table2}], nil}
+
+  """
+  @spec concat(table(), table(), direction(), cursor(), limit()) ::
+          {[{record(), table()}], cursor()}
+  def concat(first_table, last_table, direction, cursor, limit) do
```
jyeshe, line 34: Moved to a generic module to be used by other pair of tables, right? With the goal and as an exercise for alternatives to try to simplify the input and output, what you think about having functions for specific pair of tables (e.g `concat_oracles`)? Thinking on more specific solution we might possibly not need passing the table atoms on the input and avoid the tuples allocation on the output:
     `{Enum.map(start_keys, &{&1, start_table}) ++ Enum.map(end_keys, &{&1, end_table}), next_key}`

jyeshe (commented):
Good to have names for certain purpose. Thanks

Sharing a reflection whether could we simplify it some more turning it into a more specific solution.

lib/ae_mdw/collection.ex
sborrazas: I would like this module to be domain-agnostic, sort of like a Util kind of module. Business-logic kind of logic would instead be on the context modules (`AeMdw.Oracles`, `AeMdw.Names`, etc). These modules already know the table names, adding them here as well would mean duplicating them

jyeshe: Just checking if there will be another case that will benefit from this generalization. Considering this would be possible, that it would used by AeMdw.Names too for example, could we analyze alternatives to tagging each element of the output? Something like:
`{{start_table, start_keys}, {end_table, end_keys}, next_key}`

jyeshe: Just checking if there will be another case that will benefit from this generalization. Considering this would be possible, that it would used by `AeMdw.Names` too for example, could we analyze alternatives to tagging each element of the output? Something like:
`{{start_table, start_keys}, {end_table, end_keys}, next_key}`

sborrazas: Good idea, done

jyeshe (approved):
Refactor looking good and oracle integration tests passing (with rebase).

lib/ae_mdw/collection.ex
```diff
@@ -0,0 +1,69 @@
...
+  results of one come after the other, depending on the direction.
+
+  ## Examples
+
+    iex> :mnesia.dirty_all_keys(:table1)
+    [:a, :b]
+    iex> :mnesia.dirty_all_keys(:table2)
+    [:c, :d, :e]
+    iex> AeMdw.Collection.concat(:table1, :table2, :forward, nil, 3)
+    {{:table1, [:a, :b]}, {:table2, [:c, :d]}, :d}
```
jyeshe, line 25: approved, only noticed this: `{{:table1, [:a, :b]}, {:table2, [:c]}, :d}`

PR #257, opened 2021-09-23
refactor: add paginated name endpoints without making use of streams

sborrazas:
This includes `/names`, `/names/active`, `/names/inactive` endpoints.

### DIFF
--- a/lib/ae_mdw/auction_bids.ex
+++ b/lib/ae_mdw/auction_bids.ex
@@ -1,0 +1,49 @@
+defmodule AeMdw.AuctionBids do
+  @moduledoc """
+  Context module for dealing with AuctionBids.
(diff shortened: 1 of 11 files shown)


jyeshe (changes requested):
As most of *pre*-existing integration tests rely on `Cont.response_data` which is the same of what the controller does (validating practically only the json encoding and "context" call), it would be important to check if the new modules are retrieving the same data as the old way by previous version of the controller.

One way could be in a separate branch, just for the record, where the new controller endpoints are exposed in routes like a '/v2' just to be able to call old and new controller implementations and compare them.

jyeshe (approved):

thepiwo (approved):

PR #270, opened 2021-10-01
switch localhost:4000 by ae hosted infra for middleware

omar-saadoun:
Replaced http://locahost:4000 for  https://mainnet.aeternity.io/mdw/ on examples all over readme file

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,7 +299,7 @@ Type groups for the transactions listed above are:
 
 `type` parameter:
 ```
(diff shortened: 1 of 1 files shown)


thepiwo:
can you re-word the commit message to be "chore: switch localhost:4000 by hosted infra for mdw" and force push for the commit linter to be satisfied?

thepiwo:
I think we should also add some description in the top, as the hosted infrastructure adds an additional segment of /mdw/ to the path which is not there in local development or default deployment.

Maybe also add some description for the testnet hosted infrastructure.

PR #277, opened 2021-10-06
fix: start :aesync and :app_ctrl_server when initializing app

### LINKED ISSUE #275: mdw incompatibility with node 6.3.0
using mdw with node 6.3.0 seems to fail when using the status endpoint

image

sborrazas:
In this new version of aecore (6.3.0) both the :app_ctrl_server process are required for aecore to start syncing and to respond to
functions such as `:asc_sync.sync_progress()`.

This is rather hack-ish, but I don't know why none of these things are started when aecore starts.

closes #275

### DIFF
--- a/lib/ae_mdw/application.ex
+++ b/lib/ae_mdw/application.ex
@@ -25,6 +25,8 @@ defmodule AeMdw.Application do
     init(:aehttp)
     init_public(:db_state)
     # init(:aesophia)
+    init(:app_ctrl_server)
+    init(:aesync)
 
     children = [
       AeMdw.Sync.Watcher,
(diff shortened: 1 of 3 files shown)


jyeshe (approved):
:raised_hands:

priv/migrations/20210914100800_index_inner_txs.ex
```diff
@@ -28,8 +27,6 @@ defmodule AeMdw.Migrations.IndexInnerTxs do
     if not from_startup? and :ok != Application.ensure_started(:ae_mdw) do
       IO.puts("Ensure sync tables...")
       SyncSup.init_tables()
-      MdwApp.init(:contract_cache)
-      MdwApp.init(:db_state)
```
jyeshe, line 32: Needed by the migration because it replays the sync calling `sync_transaction`.

jyeshe, line 32: Needed by the migration because it replays the sync calling `AeMdw.Db.Sync.Transaction.sync_transaction/4`.

jyeshe (changes requested):

PR #279, opened 2021-10-08
feat: async account aex9 presence

### LINKED ISSUE #268: Sync in parallel the account aex9 presence based on balance
The sync of accounts presence on aex9 contracts based on balance is time consuming.

It can be done in parallel as the mapping is done independently from transaction indexation (txi).

https://github.com/aeternity/ae_mdw/blob/0a7c7d1d606ac584e1054625ba793532e034ec18/lib/ae_mdw/db/contract.ex#L357

jyeshe:
## What

Runs AEX9 account presence sync update asynchronously.

## Why

Blocking calls during Mdw database sync are taking a **long time**.
Fixes #268 

## Problem snapshot
Times in milliseconds for each event log entry of aex9 contract during the sync.

Screenshot from 2021-10-08 00-00-43

## Validation steps

Validation based on async tasks database state:
1. `make shell`
2. `require Ex2ms; require AeMdw.Db.Model; import AeMdw.Db.Util; alias AeMdw.Db.Model;`
3. `{tasks1, _cont} = select(Model.AsyncTasks, Ex2ms.fun do record -> record end, 1000)`
4. Wait 30 seconds or so...
5. `{tasks2, _cont} = select(Model.AsyncTasks, Ex2ms.fun do record -> record end, 1000)`
6. `tasks1 -- tasks2 == tasks1` # should always be true indicating it's consuming all events
7. Repeat 3-6 some times.

PS: This validation relies on the fact that the `AeMdw.Sync.AsyncTasks.Consumer.process/1` code is the same and the on aex9 presence tests found here: `test/ae_mdw/db/contract_test.exs`


lib/ae_mdw/sync/async_tasks/consumer.ex
```diff
@@ -0,0 +1,72 @@
...
+    Enum.each(amounts, fn {{:address, account_pk}, _amount} ->
+      if account_pk == delete_account_pk do
+        Contract.aex9_delete_presence(contract_pk, -1, delete_account_pk)
+      else
+        Contract.aex9_write_new_presence(contract_pk, -1, account_pk)
+      end
+    end)
+  end
+
+  defp sleep_msecs(), do: @base_sleep_msecs + Enum.random(-200..200)
```
sborrazas, line 71: Why would it sleep for negative miliseconds?

lib/ae_mdw/sync/async_tasks/store.ex
```diff
@@ -0,0 +1,78 @@
...
+  def fetch_unprocessed() do
+    any_spec =
+      Ex2ms.fun do
+        record -> record
+      end
+
+    {m_tasks, _cont} = Util.select(Model.AsyncTasks, any_spec, @max_buffer_size)
+
+    Enum.filter(m_tasks, fn Model.async_tasks(index: index) ->
+      [] == :ets.lookup(:async_tasks_processing, index)
```
sborrazas, line 38: We can use `:ets.member(tab, key)` here instead

lib/ae_mdw/sync/async_tasks/producer.ex
```diff
@@ -0,0 +1,71 @@
...
+    {:reply, m_task, new_state}
+  end
+
+  #
+  # Private functions
+  #
+  defp next_state(%{buffer: []} = state) do
+    case Store.fetch_unprocessed() do
+      [] -> {nil, state}
+      [m_task] -> {m_task, state}
```
sborrazas, line 63: This line isn't really needed here

mix.lock
```diff
@@ -18,6 +18,7 @@
   "erlex": {:hex, :erlex, "0.2.6", "c7987d15e899c7a2f34f5420d2a2ea0d659682c06ac607572df55a43753aa12e", [:mix], [], "hexpm", "2ed2e25711feb44d52b17d2780eabf998452f6efda104877a3881c2f8c0c0c75"},
   "ex2ms": {:hex, :ex2ms, "1.6.0", "f39bbd9ff1b0f27b3f707bab2d167066dd8965e7df1149b962d94c74615d0e09", [:mix], [], "hexpm", "0d1ab5e08421af5cd69146efb408dbb1ff77f38a2f4df5f086f2512dc8cf65bf"},
   "file_system": {:hex, :file_system, "0.2.10", "fb082005a9cd1711c05b5248710f8826b02d7d1784e7c3451f9c1231d4fc162d", [:mix], [], "hexpm", "41195edbfb562a593726eda3b3e8b103a309b733ad25f3d642ba49696bf715dc"},
+  "gen_stage": {:hex, :gen_stage, "1.1.1", "78d83b14ca742f4c252770bcdf674d83378ca41579c387c57e2f06d70f596317", [:mix], [], "hexpm", "eb90d2d72609050a66ce42b7d4a69323a60c892a09ead0680d5d8ef16b9a034e"},
```
sborrazas, line 21: I don't see where gen_stage is being used on this PR

mix.lock
```diff
@@ -2,10 +2,10 @@
...
-  "bunt": {:hex, :bunt, "0.2.0", "951c6e801e8b1d2cbe58ebbd3e616a869061ddadcc4863d0a2182541acae9a38", [:mix], [], "hexpm", "7af5c7e09fe1d40f76c8e4f9dd2be7cebd83909f31fee7cd0e9eadc567da8353"},
   "benchee": {:hex, :benchee, "1.0.1", "66b211f9bfd84bd97e6d1beaddf8fc2312aaabe192f776e8931cb0c16f53a521", [:mix], [{:deep_merge, "~> 1.0", [hex: :deep_merge, repo: "hexpm", optional: false]}], "hexpm", "3ad58ae787e9c7c94dd7ceda3b587ec2c64604563e049b2a0e8baafae832addb"},
+  "blankable": {:hex, :blankable, "1.0.0", "89ab564a63c55af117e115144e3b3b57eb53ad43ba0f15553357eb283e0ed425", [:mix], [], "hexpm", "7cf11aac0e44f4eedbee0c15c1d37d94c090cb72a8d9fddf9f7aec30f9278899"},
   "bmark": {:hex, :bmark, "1.0.3", "516e4383fbf6128b580925336ae2ccda50ecb68ce393d72a04b152165ce87df4", [:mix], [], "hexpm", "243b04d2e2431d01d93f442896d5e6ee52396782c161ef0a0e4f51a353393d93"},
+  "bunt": {:hex, :bunt, "0.2.0", "951c6e801e8b1d2cbe58ebbd3e616a869061ddadcc4863d0a2182541acae9a38", [:mix], [], "hexpm", "7af5c7e09fe1d40f76c8e4f9dd2be7cebd83909f31fee7cd0e9eadc567da8353"},
```
sborrazas, line 8: Can we revert these two changes? that way we can make better use of git blame

lib/ae_mdw/sync/async_tasks/consumer.ex
jyeshe: The sleep will be always positive, only the delta might be negative.

lib/ae_mdw/sync/async_tasks/producer.ex
jyeshe: Why not? :thinking:

lib/ae_mdw/sync/async_tasks/consumer.ex
sborrazas: I just realized it was a sum and not a multiplication, thanks for clarifying!

PR #286, opened 2021-10-13
feat: async tasks status

jyeshe:
## What

Add async tasks counters to /status endpoint response.

## Why

Monitor eventual bottlenecks.

## Validation steps

`elixir --sname aeternity@localhost -S mix test test/ae_mdw/sync/async_tasks/stats_test.exs`
`elixir --sname aeternity@localhost -S mix test.integration test/integration/ae_mdw_web/controllers/util_controller_test.exs`


lib/ae_mdw/sync/async_tasks/producer.ex
```diff
@@ -51,14 +54,18 @@ defmodule AeMdw.Sync.AsyncTasks.Producer do
       Store.set_processing(index)
     end
 
+    new_state.buffer
+    |> length()
```
jyeshe, line 58: full list scan but very fast for less than 100 items and few calls

PR #292, opened 2021-10-19
fix: duplicated spend_tx with sender = recipient

### LINKED ISSUE #162: Sending coins to itself makes appear transaction in the list of transactions two times

jyeshe:
## What

Dedup `/txs?account=:aid` response.

## Why

Indexing by `recipient_id` and `sender_id` is needed to continue to have consistent results when filtering `/txs` by these fields.
Fixes #162 

## Validation steps

Specific:
`elixir --sname aeternity@localhost -S mix test test/ae_mdw/db/sync/transaction_test.exs`
`elixir --sname aeternity@localhost -S mix test.integration test/integration/ae_mdw_web/controllers/tx_controller_test.exs:815`

Regression:

### DIFF
--- a/.github/workflows/test-action.yml
+++ b/.github/workflows/test-action.yml
@@ -69,7 +69,7 @@ jobs:
         id: plt-cache
         with:
           path: priv/plts
-          key: plts-v2-${{ hashFiles('mix.lock') }}
+          key: plts-v3-${{ hashFiles('mix.lock') }}
 
       - name: Common tests setup
         uses: ./.github/actions/tests-common


lib/ae_mdw/node.ex
```diff
@@ -116,6 +116,10 @@ defmodule AeMdw.Node do
   end
 
   @spec tx_ids(atom()) :: %{atom() => non_neg_integer()}
+  def tx_ids(:spend_tx) do
+    %{sender_id: 0, recipient_id: 0}
+  end
+
```
sborrazas, line 122: This is going to be overriden by the `SmartGlobal.new(AeMdw.Node, ...)` call on startup, I'm assuming it's just for documentation purposes?

jyeshe: not needed anymore for dialyzer, thanks

PR #296, opened 2021-10-20
fix: fix /txs route handling

sborrazas:


jyeshe (approved):

test/ae_mdw_web/controllers/tx_controller_test.exs
```diff
@@ -0,0 +1,9 @@
+defmodule AeMdwWeb.TxControllerTest do
```
jyeshe, line 1: In the browser works. For CI purposes this can be an integration or `Util.last_gen()` mocked.

PR #318, opened 2021-11-01
feat: add recipient details for /tx and /txi

### LINKED ISSUE #316: Transaction to a name should have info about account on certain height
https://testnet.aeternity.io/mdw/txi/22496032 as a `recipient_id` there is a name hash for such transactions, i want to have an exact account that name had pointer at, at the moment it was send.

jyeshe:
## What

Add recipient details also for `/txi` and `/tx` endpoints.

## Why

Fixes #316 

## Validation steps

Case:

Regression:

### DIFF
--- a/lib/ae_mdw_web/controllers/tx_controller.ex
+++ b/lib/ae_mdw_web/controllers/tx_controller.ex
@@ -106,21 +106,22 @@ defmodule AeMdwWeb.TxController do
   defp handle_spendtx_details(response_data, %{"account" => _account}) do
     response_data
     |> Enum.uniq_by(fn %{"hash" => tx_hash} -> tx_hash end)
-    |> Enum.map(fn %{"tx" => block_tx, "tx_index" => tx_index} = block ->
-      spend_tx_recipient = (block_tx["type"] == @type_spend_tx && block_tx["recipient_id"]) || nil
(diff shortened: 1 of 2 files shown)


lib/ae_mdw_web/controllers/tx_controller.ex
```diff
@@ -106,22 +106,23 @@ defmodule AeMdwWeb.TxController do
...
-    end)
+    |> Enum.map(&maybe_add_spend_tx_details/1)
   end
 
   defp handle_spendtx_details(response_data, _params), do: response_data
 
+  defp maybe_add_spend_tx_details(%{"tx" => block_tx, "tx_index" => tx_index} = block) do
+    recipient_id = block_tx["recipient_id"] || ""
+
+    if block_tx["type"] == @type_spend_tx and String.starts_with?(recipient_id, "nm_") do
```
thepiwo, line 117: do we usually do these checks with String methods? Or can't we use the ID system from the node in binary?

jyeshe: We really need to test the formatted recipient_id here because of multiple sources.

thepiwo (approved):

sborrazas (approved):

PR #320, opened 2021-11-02
feat: add cursor-based pagination to scoped txs

sborrazas:


thepiwo (approved):

test/integration/ae_mdw_web/controllers/tx_controller_test.exs
```diff
@@ -1366,33 +1366,20 @@ defmodule Integration.AeMdwWeb.TxControllerTest do
...
              end)
 
       assert Enum.any?(response["data"], fn %{"tx" => block_tx} ->
-               if block_tx["type"] == "SpendTx" do
-                 assert block_tx["recipient_id"] == id or block_tx["sender_id"] == id
-                 true
-               else
-                 false
-               end
+               block_tx["type"] == "SpendTx"
```
jyeshe, line 1377: missing only the `and (block_tx["recipient_id"] == id or block_tx["sender_id"] == id)`
to validate the account

lib/ae_mdw/txs.ex
```diff
@@ -37,15 +39,27 @@ defmodule AeMdw.Txs do
...
-  def fetch_txs(direction, query, cursor, limit) do
+  def fetch_txs(direction, scope, query, cursor, limit) do
     ids = query |> Map.get(:ids, MapSet.new()) |> MapSet.to_list()
     types = query |> Map.get(:types, MapSet.new()) |> MapSet.to_list()
     cursor = deserialize_cursor(cursor)
 
+    txis_range =
+      case scope do
+        {:gen, %Range{first: first_gen, last: last_gen}} ->
+          {gen_to_txi(first_gen, direction), gen_to_txi(last_gen, opposite_dir(direction))}
```
jyeshe, line 52: I was a little confused by the direction change to the opposite one. Maybe it would be better to have explicit functions:
`first_gen_to_txi` and `last_gen_to_txi` as the direction is used to select the field.

jyeshe (commented):
A single readability opportunity and a minor in test case change.

test/integration/ae_mdw_web/controllers/tx_controller_test.exs
sborrazas: Nice catch, not sure why I removed it.

lib/ae_mdw/txs.ex
sborrazas: Not sure what you meant by this, I think I've done what you asked for, it doesn't really change things much. For next time, please provide code samples, it's a lot less ambiguous than Enligsh! :sunglasses:

jyeshe (approved):

PR #321, opened 2021-11-03
feat: aex9 migration balance timeout

jyeshe:
## What

Adds 1 minute timeout to aex9 balance and saves the contract to be handled separately (in a dets/temporary table).

## Why

Migration was taking more than 10 minutes (sometimes 20 min) because of two very big contracts. 
This change avoids the Mdw (testnet and mainnet) to be out of sync for so long time.

## Result
Total time reduced:
Screenshot from 2021-11-03 11-09-50

Contracts saved:
Screenshot from 2021-11-03 11-09-14


thepiwo (approved):

sborrazas (approved):
LGTM

PR #324, opened 2021-11-04
feat: implement cursor-based pagination for scoped oracles & names

sborrazas:


thepiwo (approved):

lib/ae_mdw/oracles.ex
```diff
@@ -16,38 +16,60 @@ defmodule AeMdw.Oracles do
...
+        nil ->
+          nil
+
+        {:gen, %Range{first: first_gen, last: last_gen}} ->
+          if direction == :forward do
+            {{first_gen, <<>>}, {last_gen, <<>>}}
+          else
+            {{first_gen, nil}, {last_gen, nil}}
+          end
+      end
```
jyeshe, line 42: The naming used in `AeMdw.Oracles` and `AeMdw.Names` looks conflicting with the one in AeMdw.Collection regarding to the `scope()` in a way they appear swapped. Let me try to explain ...

One might expected `gen_range` variable to be a `{:gen, %Range{first: first_gen, last: last_gen}}` while scope to be the `{{first_gen, <<>>}, {last_gen, <<>>}}` or `{{first_gen, nil}, {last_gen, nil}}` because `scope()` type of the `AeMdw.Collection.stream` `scope` argument is a `{key, key}`.

Another factor contributes to make this a good opportunity for a better readability is in regards to definition of `scope()` type of `AeMdw.Collection`. Looking to the :oracle and :tx models, neither of their keys are like `{first_gen, <<>>} or {first_gen, nil}`.

jyeshe (commented):
Can you add a name test case for a name claimed by a ga_meta_tx?

The code looks good to me with one refactoring opportunity.

test/integration/ae_mdw_web/controllers/name_controller_test.exs
```diff
@@ -322,6 +403,54 @@ defmodule Integration.AeMdwWeb.NameControllerTest do
...
+      first = 100_000
+      last = 500_000
+
+      assert %{"data" => data, "next" => next} =
+               conn
+               |> get("/names/gen/#{first}-#{last}")
+               |> json_response(200)
+
+      assert Enum.all?(data, fn %{"info" => %{"expire_height" => kbi}} ->
+               first <= kbi && kbi <= last
```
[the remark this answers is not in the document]
sborrazas: Done

lib/ae_mdw/oracles.ex
sborrazas: Done

jyeshe (commented):
Only missing the test case to paginate a page with a name claimed by a `ga_meta_tx` that caused the first `Internal server error` after the deployment.

jyeshe (approved):
Test and code looking good considering we will do an automation of endpoints testing with dev mode.

PR #336, opened 2021-11-11
test: add local js SDK setup for testing purposes

sborrazas:


thepiwo (commented):
if the goal is to just use it locally lets put it in a separate repo, if we need such thing for unit tests I think we should look for other possibilities, e.g. trying to use the node directly or using the elixir sdk (if still works)

sborrazas:
@thepiwo indeed the Elixir SDK would be ideal. Perhaps we could create a task to check if it works and integrate it for local testing?

thepiwo:
@sborrazas feel free to take a few hours to try it out

marc0olo:
> @sborrazas feel free to take a few hours to try it out

yeah, maybe we could even discuss to make it Iris compatible and feature complete. I would support this

jyeshe (approved):
Looks good for the JS SDK (in case we won't use the Elixir SDK).

thepiwo:
I would propose to close this, do you have different opinions?

sborrazas:
Closing, perhaps next sprint we can start looking into bringing the Elixir SDK back to life.

PR #337, opened 2021-11-12
fix: always display the correct contract_id on contract logs

sborrazas:
closes: #301


thepiwo (approved):
looks good as far as I can see

jyeshe (approved):
Integration test looks good and passes.

PR #340, opened 2021-11-15
feat: long running aex9 balance

### LINKED ISSUE #339: slow mdw sync due to long async task
Mdw sync seems to be stuck around height 408217 and the async tasks stats doesn't decrement.

jyeshe:
## What

Dedicated queue for long running aex9 balance async tasks.

## Why

Fixes #339

## Result

Before (apparently stuck):
Screenshot from 2021-11-15 12-12-13

After:
Screenshot from 2021-11-15 19-18-21

Screenshot from 2021-11-15 19-21-04

3.75 minutes later:
Screenshot from 2021-11-15 19-22-06

In parallel to well-known heavy contract:
```
iex(aeternity@localhost)5> key = :mnesia.dirty_first(Model.AsyncTasks) 
{1636509318043573210, :update_aex9_presence}
iex(aeternity@localhost)6> :mnesia.dirty_read(Model.AsyncTasks, key)
[
  {:async_tasks, {1636509318043573210, :update_aex9_presence},
   [
     <<24, 213, 6, 252, 78, 224, 40, 213, 97, 104, 43, 221, 43, 226, 95, 18,
       156, 159, 128, 80, 156, 196, 127, 22, 200, 236, 251, 246, 166, 45, 133,
       116>>
   ]}
]
iex(aeternity@localhost)7> :aeser_api_encoder.encode(:contract_pubkey, <<24, 213, 6, 252, 78, 224, 40, 213, 97, 104, 43, 221, 43, 226, 95, 18, 156, 159, 128, 80, 156, 196, 127, 22, 200, 236, 251, 246, 166, 45, 133, 116>>)
"ct_BwJcRRa7jTAvkpzc2D16tJzHMGCJurtJMUBtyyfGi2QjPuMVv"
```


lib/ae_mdw/sync/async_tasks/consumer.ex
```diff
@@ -20,51 +41,119 @@ defmodule AeMdw.Sync.AsyncTasks.Consumer do
...
+    timeout = if is_long?, do: @long_timeout_msecs, else: @task_timeout_msecs
+
+    timer_ref = timeout |> :timer.send_after({:timedout, task}) |> ok!
+
+    {task, timer_ref}
+  end
+
+  @spec process(Model.async_tasks_record()) :: :ok
+  def process(Model.async_tasks(index: {_ts, type}, args: args)) do
+    type = if type == 'Elixir.AeMdw.Sync.AsyncTasks.UpdateAex9Presence', do: :update_aex9_presence, else: type
```
jyeshe, line 123: will be removed

lib/ae_mdw/sync/async_tasks/consumer.ex
```diff
@@ -20,51 +41,115 @@ defmodule AeMdw.Sync.AsyncTasks.Consumer do
...
+      else
+        # some still running
+        state
+      end
+
+    {:noreply, new_state}
+  end
+
+  @doc """
+  Just acknowledge (ignore) the DOWN event.
```
jyeshe, line 91: No evidence yet for the need of retry

.iex.exs
```diff
@@ -0,0 +1,3 @@
+alias AeMdw.Db.{Model, Name, Util}
+require Model
+require Ex2ms
```
sborrazas, line 3: Perhaps we can get rid of the abbreviated alias notation, credo won't complain about it because it won't scan this file, but the full notation is always more useful because it can be searched for easily.

lib/ae_mdw/application.ex
```diff
@@ -32,6 +32,7 @@ defmodule AeMdw.Application do
       AeMdw.Sync.Watcher,
       AeMdw.Sync.Supervisor,
       AeMdw.Sync.AsyncTasks.Supervisor,
+      {Task.Supervisor, name: AeMdw.Sync.AsyncTasks.TaskSupervisor},
```
sborrazas, line 35: We can instead add a `child_spec/1` function to the `...AsyncTasks.TaskSupervisor` so we don't have to duplicate the name of the process in two very different places

lib/ae_mdw/sync/async_tasks/store.ex
```diff
@@ -80,4 +81,16 @@ defmodule AeMdw.Sync.AsyncTasks.Store do
       {[], _cont} -> false
     end
   end
+
+  defp dedup_records(tasks) do
```
jyeshe, line 85: As shown by `Store.save_new/2` test case, the code already dedups for new tasks (but some database might have duplicated already saved like mainnet.zip snapshot from Nov 10th).

lib/ae_mdw/application.ex
jyeshe: In this case it's not doable because `...AsyncTasks.TaskSupervisor` is only the name (the module is the `Elixir.Task.Supervisor`). I have moved it to the `AsyncTasks.Supervisor` though in order to shutdown the task when a `AsyncTasks.Consumer` goes down..

.iex.exs
jyeshe: Done

lib/ae_mdw/sync/async_tasks/consumer.ex
```diff
@@ -1,17 +1,40 @@
...
+  alias AeMdw.Sync.AsyncTasks.UpdateAex9Presence
 
   require Model
+  require Logger
+
+  import AeMdw.Util, only: [ok!: 1]
+
+  @base_sleep_msecs 3_000
+  @yield_timeout_msecs 100
+  @task_timeout_msecs 30_000
```
jyeshe, line 21: changed from 20s to 30s as the counter of long tasks reached 5 (different contracts).

lib/ae_mdw/application.ex
sborrazas: You can always define it like so:
```elixir
def child_spec(_args) do
  TaskSupervisor.child_spec(name: __MODULE__) # Or use whatever name you want
end
```
and then just do
```elixir
AeMdw.Sync.AsyncTasks.TaskSupervisor,
```
on the children list

jyeshe: That is interesting but the `,,,AsyncTasks.TaskSupervisor` here does not exist, is just a name. 
Now the spec is private to `,,,AsyncTasks.Supervisor`. May we continue with it to avoid retesting after a creation of new module?

sborrazas: ok

sborrazas (approved):

thepiwo (approved):

PR #345, opened 2021-11-18
feat: consistency check for auction and name expirations

jyeshe:
Uses the node/chain block state at a certain height to validate name state transitions regarding auction and name expiration.


lib/ae_mdw/db/sync/name.ex
```diff
@@ -69,6 +76,13 @@ defmodule AeMdw.Db.Sync.Name do
 
       timeout ->
         auction_end = height + timeout
+        node_auction_end = node_auction_end(plain_name, height + 1) || 0
```
jyeshe, line 79: at the same height the auction doesn't exist yet in the `aens_state_tree`.

lib/ae_mdw/db/sync/name.ex
```diff
@@ -45,6 +47,11 @@ defmodule AeMdw.Db.Sync.Name do
         previous = ok_nil(cache_through_read(Model.InactiveName, plain_name))
         expire = height + :aec_governance.name_claim_max_expiration(proto_vsn)
 
+        if expire != node_name_ttl(plain_name, height + 1) do
```
jyeshe, line 50: first difference caught: `16:59:15.984 [error] [4054] name expire differ for bear.test`
```
iex(aeternity@localhost)8> expire = height + :aec_governance.name_claim_max_expiration(Util.proto_vsn(4055))
54054
iex(aeternity@localhost)9> Name.node_name_ttl("bear.test", 4055)                                            
40054
```

thepiwo (commented):
can you explain a bit what is happening here?

jyeshe:
> can you explain a bit what is happening here?

The idea is to use name system tree, from the respective chain block state (at the height when a name operation happens), to validate the transition the Mdw is performing in the name state.

It's better to keep it as `Draft` though until the behaviour of `:aens_state_tree.get_name` (or eventual alternative) is cleared out since it has not returned the name state for a certain height but with the last state.

lib/ae_mdw/db/sync/name.ex
jyeshe: 54054 is correct. Another micro-block afterwards in the same height changes the ttl to 40054.

jyeshe:
Claim and update happened in the same microblock for `bear.test`:
https://mainnet.aeternity.io/mdw/txi/5800
https://mainnet.aeternity.io/mdw/txi/5801

The `aens_state_tree` belongs to a block_state (`:aec_db.get_block_state(hash) |> :aec_trees.ns()`) which can be retrieved for a key-block or micro-block. When two transactions for the certain name happens in the same micro-block, like `bear.test` case, the block state has the name state only for the last change. For this reason, during a name transaction indexation, the actual name state cannot be validated with the node state.

thepiwo (commented):
whats the state of this?

jyeshe:
> whats the state of this?

This is on hold once it has not provided meaningful validations for the release.
Also these changes were avoided for now to be able to better compare the name sync results.

On next meeting I will provide a more detailed info on the comments above to hear team's opinion if it's valid to continue with it or not.

jyeshe:
Closing after # 344

Issue #346, opened 2021-11-19
investigate sync speed bottlenecks


thepiwo:

sborrazas:
These are the findings I concluded when syncing 250,000 generation forward:

1. Contract call transactions on the when processing the WriteLinksMutation
   This includes the two calls to mnesia:
     a. [Call to fetch the block](https://github.com/aeternity/ae_mdw/blob/master/lib/ae_mdw/db/sync/contract.ex#L36)
     b. [Call to fetch the contract creation txi](https://github.com/aeternity/ae_mdw/blob/master/lib/ae_mdw/db/sync/contract.ex#L38)
2. Contract events processing on the ContractEventsMutation
   This happens when calling [get_txi](https://github.com/aeternity/ae_mdw/blob/master/lib/ae_mdw/db/mutations/contract_events_mutation.ex#L27) to fetch the contract creation txi

All of these mutations processing ends up taking between 40ms and 80ms. There's quite a lot of these transactions, if we tackle these items we can get a much better performance improvement.

jyeshe:
Profiling the reading from generation 0 to 540,000:
- loading blocks: 7h40m
- loading blocks + events: 12h (already with the optimization of #376)

Dry-running the txs to get 2 events in parallel reaches an improvement of 50% but degrades the performance after 300,000.

Since dry-running still represent almost 40% of total read time, an opportunity of optimization is to drop the transactions after contract calls (don't affect the events).
PS: The ones before the contract calls cannot be dropped from dry run because they change the chain state trees that are an input for contract calls dry run.

jyeshe:
Here is a comparison on writing between Mnesia (mnesia_rocksdb) vs RocksDb (direct use of NIF):
```
names = :mnesia.dirty_all_keys(Model.ActiveName)
m_names = names |> Enum.take(100_000) |> Enum.flat_map(fn key -> :mnesia.dirty_read(Model.ActiveName, key) end)
delete = fn -> names |> Enum.take(100_000) |> Enum.each(fn key -> :mnesia.dirty_delete(Model.ActiveName, key) end); end
write100k = fn -> :timer.tc(fn -> :mnesia.sync_transaction(fn -> m_names |> Enum.each(fn m_name -> :mnesia.write(Model.ActiveName, m_name, :write) end) end) end) end
Enum.map(1..10, fn _i -> delete.(); Process.sleep(15_000); res = write100k.(); Process.sleep(15_000); res end) 
```
Result:
[
[...]

PR #348, opened 2021-11-22
feat: async derive_aex9_presence

jyeshe:
## What

Async processing for the `Sync.Contract.aex9_derive_account_presence!/1`.

## Why

Synchronous aex9 balances makes mdw database sync slower. 
It helps with #346

### DIFF
--- a/lib/ae_mdw/db/contract.ex
+++ b/lib/ae_mdw/db/contract.ex
@@ -109,8 +109,6 @@ defmodule AeMdw.Db.Contract do
   @spec logs_write(integer(), integer(), tuple()) :: :ok
   def logs_write(create_txi, txi, call_rec) do
     contract_pk = :aect_call.contract_pubkey(call_rec)
-    is_aex9_contract? = Contract.is_aex9?(contract_pk)
-    aex9_transfer_evt = AeMdw.Node.aex9_transfer_event_hash()
     raw_logs = :aect_call.log(call_rec)
 
     raw_logs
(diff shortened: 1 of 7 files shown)


lib/ae_mdw/db/contract.ex
```diff
@@ -149,14 +147,31 @@ defmodule AeMdw.Db.Contract do
...
   end
 
+  @spec is_aex9_transfer?(binary(), pubkey()) :: boolean()
+  def is_aex9_transfer?(evt_hash, aex9_contract_pk) do
+    aex9_transfer_evt = AeMdw.Node.aex9_transfer_event_hash()
+
+    evt_hash == aex9_transfer_evt and aex9_contract_pk != nil
+  end
+
+  @spec which_aex9_contract_pubkey(pubkey(), pubkey()) :: pubkey()
```
sborrazas, line 165: This function appears to be returning `nil` as well in some cases

lib/ae_mdw/sync/async_tasks/derive_aex9_presence.ex
```diff
@@ -0,0 +1,124 @@
...
+    call_rec
+    |> :aect_call.log()
+    |> Enum.map(fn {addr, [evt_hash | args], _data} ->
+      aex9_contract_pk = DBContract.which_aex9_contract_pubkey(contract_pk, addr)
+
+      if DBContract.is_aex9_transfer?(evt_hash, aex9_contract_pk) do
+        [_from_pk, _to_pk, <<_amount::256>>] = args
+      end
+    end)
+    |> Enum.reject(&(&1 == nil))
```
sborrazas, line 122: Not that it matters, but for future implementations, this can be `&is_nil/1`

lib/ae_mdw/sync/async_tasks/derive_aex9_presence.ex
```diff
@@ -0,0 +1,124 @@
...
+  from a create contract.
+  """
+  @behaviour AeMdw.Sync.AsyncTasks.Work
+
+  alias AeMdw.Node.Db, as: DBN
+
+  alias AeMdw.Contract
+  alias AeMdw.Db.Contract, as: DBContract
+  alias AeMdw.Db.Model
+  alias AeMdw.Db.Util, as: DBU
```
sborrazas, line 13: We could perhaps get rid of renaming aliases, so that we can find all function for any function more easily

jyeshe: It's used through out the code probably to allow accessing `AeMdw.Contract` and `AeMdw.Db.Contract` (same for Util in original module but here I could change for Util).

I got used to the DBN but we can have a refactoring PR to get rid of it through out the code renaming to sth like `AeMdw.Node.NodeDb`.

lib/ae_mdw/sync/async_tasks/derive_aex9_presence.ex
jyeshe: done

lib/ae_mdw/sync/async_tasks/derive_aex9_presence.ex
```diff
@@ -0,0 +1,124 @@
...
+      |> Enum.into(MapSet.new())
+
+    recipients = get_aex9_recipients(contract_pk, kbi, mbi)
+
+    pks =
+      Enum.reduce(recipients, all_pks, fn to_pk, pks ->
+        MapSet.delete(pks, to_pk)
+      end)
+
+    :mnesia.sync_transaction(fn ->
```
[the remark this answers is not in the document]
jyeshe: The `Enum.each/2` and `:aex9_write_presence/3` always return :ok but nice to have that everywhere.

sborrazas (approved):

thepiwo (approved):

PR #349, opened 2021-11-22
fix: contract might not be present

jyeshe:
## What

Handles error for contract init info when `contract_create_tx` is on chain but the contract is not in chain db because its `init` has failed.

## Why

Fix for http://localhost:4000/txs/backward?limit=100&type=contract_create

## Error screenshot

Screenshot from 2021-11-22 17-59-25


sborrazas (approved):
LGTM

thepiwo (approved):

Issue #359, opened 2021-11-25
Fix /transfer routes which aren't returning results properly


sborrazas:
There's a few issues regarding the transfers endpoints:

1. When ordering results backwards, they cannot be filtered by kind
   https://mainnet.aeternity.io/mdw/transfers/gen/20000-19000 -> all transfers are "fee_..."
   https://mainnet.aeternity.io/mdw/transfers/gen/20000-19000?kind=fee -> no "fee_..." transfers
2. when filtering by both account and kind, it does not work if the transactions being filtered are not belonging to the kind of the initial transfers
  https://mainnet.aeternity.io/mdw/transfers/forward?account=ak_JFkVmYeY9iP4gmKexBHx7t1aAj1R6FGvBdWaZRBVLuuzWv83j&limit=10 -> there are a few fee_spend_name transfers there
  https://mainnet.aeternity.io/mdw/transfers/forward?account=ak_JFkVmYeY9iP4gmKexBHx7t1aAj1R6FGvBdWaZRBVLuuzWv83j&limit=10&kind=fee_spend_name -> no fee_spend_name transfers

These are two separate issues which can be resolved by relying on the fact that there's just a few finite kinds, and they don't need to be scoped by prefix:

1. Can be fixed by filtering by each specific kind and then merging results
2. This is a limitation due to the indexes we have, we would need to add a new index to be able to support these features

sborrazas:
Fixed by the two PRs listed here.

PR #360, opened 2021-11-25
fix: allow filtering transfer by kind when backwards direction

sborrazas:
refs #359


thepiwo (approved):

jyeshe (commented):
Code and test look good to me. 

It would good only to add one also for `get(conn, "/transfers/backward?kind=#{kind_prefix}") `

sborrazas:
@jyeshe added a new test case for that situation now

jyeshe (approved):

PR #362, opened 2021-11-29
fix: reindex transfers to be able to filter by account + kind

sborrazas:
Right now, the `TargetIntTransferTx` mnesia table is indexed by `{account_pk, gen_txi, kind, ref_txi}`. This doesn't allow to filter by both account and kind with results sorted by gen_txi.

Instead, the new index created on this migration, named `TargetKindIntTransferTx`, will have `{account_pk, kind, gen_txi, ref_txi}` as a key, which allows to filter by both account and kind on a given range.

refs #359

### DIFF
--- a/lib/ae_mdw/transfers.ex
+++ b/lib/ae_mdw/transfers.ex
@@ -30,8 +30,8 @@ defmodule AeMdw.Transfers do
   @kinds ~w(fee_lock_name fee_refund_name fee_spend_name reward_block reward_dev reward_oracle)
 
   @int_transfer_table Model.IntTransferTx
-  @target_int_transfer_tx Model.TargetIntTransferTx
-  @kind_int_transfer_tx Model.KindIntTransferTx
+  @target_kind_int_transfer_tx_table Model.TargetKindIntTransferTx
+  @kind_int_transfer_tx_table Model.KindIntTransferTx
 
   @spec fetch_transfers(direction(), range(), query(), cursor() | nil, limit()) ::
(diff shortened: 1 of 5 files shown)


jyeshe:
Can you add test cases please?

sborrazas:
@jyeshe done, forgot to handle this case on tests. I have now updated a test-case so that it included this bug

jyeshe:
> @jyeshe done, forgot to handle this case on tests. I have now updated a test-case so that it included this bug

Appreciated. With this feature and migration, the Mdw will be able to filter simultaneously by account and kind on backwards direction too, right? Is so could we have one more case?

lib/ae_mdw/transfers.ex
```diff
@@ -73,50 +73,34 @@ defmodule AeMdw.Transfers do
...
-        nil -> nil
-        {gen_txi, kind, account_pk, ref_txi} -> {account_pk, gen_txi, kind, ref_txi}
-      end
-
-    @target_int_transfer_tx
-    |> Collection.stream(direction, scope, cursor)
-    |> Stream.map(fn {account_pk, gen_txi, kind, ref_txi} ->
-      {gen_txi, kind, account_pk, ref_txi}
-    end)
+    build_stream(%{account_pk: account_pk, kind_prefix: ""}, scope, cursor, direction)
```
jyeshe, line 103: :pray:

jyeshe (approved):

thepiwo (approved):

PR #367, opened 2021-11-30
refactor: extract block rewards syncing into mutation

sborrazas:
This is not the biggest bottleneck when syncing, but it will isolate
the problem and make the mnesia syncing migration faster anyway.


thepiwo (approved):
lgtm

jyeshe (approved):
looks good to me too

PR #373, opened 2021-12-03
refactor: add name/auction expiration mutation when syncing

sborrazas:


thepiwo (approved):
lgtm

jyeshe (dismissed):
Good move to take the reading out of the transaction block.

lib/ae_mdw/db/name.ex
```diff
@@ -37,6 +42,78 @@ defmodule AeMdw.Db.Name do
...
+      end
+
+    expired_names = :mnesia.dirty_select(Model.ActiveNameExpiration, name_mspec)
+    expired_auctions = :mnesia.dirty_select(Model.AuctionExpiration, auction_mspec)
+
+    NamesExpirationMutation.new(height, expired_names, expired_auctions)
+  end
+
+  @spec expire_name(Blocks.height(), Names.plain_name()) :: :ok
+  def expire_name(height, plain_name) do
```
jyeshe, line 64: It would be good to keep this to enforce consistency:
https://github.com/aeternity/ae_mdw/blob/ee3465822fecf346acc564e36f9e209faf8444f8/lib/ae_mdw/db/sync/name.ex#L215

This protects from expiring a name due to `Model.ActiveNameExpiration` that are not the last one. With this protection also we could safely remove the `ensure_no_exsiting_active_name_expiration/1` in regards to name state transition. It would be missing only the analysis if removing it might affect the endpoints that depend on proper `Model.ActiveNameExpiration` records state.

jyeshe (changes requested):

lib/ae_mdw/db/name.ex
```diff
@@ -37,6 +42,78 @@ defmodule AeMdw.Db.Name do
...
+        claims: bids,
+        auction_timeout: timeout,
+        owner: owner,
+        previous: previous
+      )
+
+    m_name_exp = Model.expiration(index: {expire, plain_name})
+    m_owner = Model.owner(index: {owner, plain_name})
+
+    ensure_no_exsiting_active_name_expiration(plain_name)
```
jyeshe, line 101: if not removed, can also be moved to outside the transaction block.

sborrazas: @jyeshe definitely. The idea is to move all of the mutations into separate mutation modules. The next step would be to trim those mutations and move as much code outside of them as possible

lib/ae_mdw/db/name.ex
sborrazas: You're right, I've added it back, thanks!

jyeshe (approved):
:rocket:

PR #374, opened 2021-12-03
feat: preload generations and contract events

jyeshe:
## What

Preloads both generations and contract events sequentially.

## Why

Contribution to #346


jyeshe:
Slower than master when node is already synced after height 150_000.

PR #376, opened 2021-12-06
perf: events only for mbs with contract calls

jyeshe:
## What

Applies transactions to get contract events only when the micro block has a contract call tx.

## Why

Events are grouped by `tx_hash` and are read only when the tx is a `:contract_call_tx`:
https://github.com/aeternity/ae_mdw/blob/ee3465822fecf346acc564e36f9e209faf8444f8/lib/ae_mdw/db/sync/transaction.ex#L82

Contribute to #346 

## Additional notes

Possibly `:contract_create_tx` events and those derived from channel execution should be indexed but are discussed in a separate issue in favor of incremental changes (#377)

## Validation steps

```
defmodule CacheTxs do
  alias AeMdw.Db.Model
  # import AeMdw.Db.Util, only: [read_block!: 1]
  require Model

  def read_events(%Range{first: _from_height, last: _to_height} = range) do
    range
    |> Enum.map(fn height ->
      # IO.inspect Model.block(read_block!({height, -1}), :hash)
      {_key_block, micro_blocks} = AeMdw.Node.Db.get_blocks(height)

      {
        height, 
        micro_blocks
        |> Enum.with_index()
        |> Enum.flat_map(fn {mblock, _mbi} ->
          events = AeMdw.Contract.get_grouped_events(mblock) 
          if events == %{}, do: [], else: [events]
        end)
      }
    end)
  end
  
  def read_filtered_txs_events(%Range{first: _from_height, last: _to_height} = range) do
    range
    |> Enum.map(fn height ->
      # IO.inspect Model.block(read_block!({height, -1}), :hash)
      {_key_block, micro_blocks} = AeMdw.Node.Db.get_blocks(height)

      {
        height, 
        micro_blocks
        |> Enum.with_index()
        |> Enum.flat_map(fn {mblock, _mbi} ->
          has_contract_tx? = 
            mblock
            |> :aec_blocks.txs()
            |> Enum.any?(fn signed_tx -> 
              {mod, _tx} = :aetx.specialize_callback(:aetx_sign.tx(signed_tx))
              mod.type() in [:contract_call_tx, :contract_create_tx]
            end)
          
          if has_contract_tx? do 
            events = AeMdw.Contract.get_grouped_events(mblock) 
            if events == %{}, do: [], else: [events]
          else
            []
          end
        end)
      }
    end)
  end
end

AeMdw.Application.sync(false)
{:ok, table} = :dets.open_file(:cachetxs_res, file: 'cachetxs_res.dets')

spawn(fn ->
  Enum.each(0..5000, fn i -> 
    first = i * 100 
    last = first + 100 
    {t1, _events1} = :timer.tc(fn -> CacheTxs.read_events(first..last) end) 
    {t2, _events2} = :timer.tc(fn -> CacheTxs.read_filtered_txs_events(first..last) end)

    :dets.insert(table, {i, t1, t2})
  end)
end)
```

after completed or by sampling:
```
:dets.match_object(table, :"$1") |> Enum.filter(fn {_i, t1, t2} -> t2 > t1 end)
```

### DIFF
--- a/lib/ae_mdw/contract.ex
+++ b/lib/ae_mdw/contract.ex
@@ -336,20 +336,33 @@ defmodule AeMdw.Contract do
   # Private functions
   #
   defp get_events(micro_block) when elem(micro_block, 0) == :mic_block do
-    header = :aec_blocks.to_header(micro_block)
-    {:ok, hash} = :aec_headers.hash_header(header)
-    consensus = :aec_headers.consensus_module(header)
-    node = {:node, header, hash, :micro}
-    time = :aec_block_insertion.node_time(node)
-    prev_hash = :aec_block_insertion.node_prev_hash(node)
-    prev_key_hash = :aec_block_insertion.node_prev_key_hash(node)
-    {:value, prev_key_header} = :aec_db.find_header(prev_key_hash)
-    {:value, trees_in, _, _, _, _} = :aec_db.find_block_state_and_data(prev_hash, true)
-    trees_in = apply(consensus, :state_pre_transform_micro_node, [node, trees_in])
-    env = :aetx_env.tx_env_from_key_header(prev_key_header, prev_key_hash, time, prev_hash)
     txs = :aec_blocks.txs(micro_block)
-    {:ok, _, _, events} = :aec_block_micro_candidate.apply_block_txs_strict(txs, trees_in, env)
-    events
+
+    if has_contract_call_tx?(txs) do
+      header = :aec_blocks.to_header(micro_block)
+      {:ok, hash} = :aec_headers.hash_header(header)
+      consensus = :aec_headers.consensus_module(header)
+      node = {:node, header, hash, :micro}
+      time = :aec_block_insertion.node_time(node)
+      prev_hash = :aec_block_insertion.node_prev_hash(node)
+      prev_key_hash = :aec_block_insertion.node_prev_key_hash(node)
+      {:value, prev_key_header} = :aec_db.find_header(prev_key_hash)
+      {:value, trees_in, _, _, _, _} = :aec_db.find_block_state_and_data(prev_hash, true)
+      trees_in = apply(consensus, :state_pre_transform_micro_node, [node, trees_in])
+      env = :aetx_env.tx_env_from_key_header(prev_key_header, prev_key_hash, time, prev_hash)
+
+      {:ok, _, _, events} = :aec_block_micro_candidate.apply_block_txs_strict(txs, trees_in, env)
+      events
+    else
+      []
+    end
+  end
+
+  defp has_contract_call_tx?(mb_txs) do
+    Enum.any?(mb_txs, fn signed_tx ->
+      {mod, _tx} = :aetx.specialize_callback(:aetx_sign.tx(signed_tx))
+      mod.type() == :contract_call_tx
+    end)
   end
 
   defp contract_init_args(contract_pk, tx_rec) do


thepiwo (commented):
whats the performance overhead of this change if every microblock **had** a contract call tx?

**ALSO** not only calls can have events, also contract creates and paying for tx

jyeshe:
> whats the performance overhead of this change if every microblock **had** a contract call tx?
> 
> **ALSO** not only calls can have events, also contract creates and paying for tx

The penalty is for the `Enum.any?` which is pretty fast. Supposing a generation with 1 million txs divided into 1000 microblocks:
```
mbs_txs = Enum.map(1..1_000_000, fn _i -> type = to_string(Enum.random(1..20)) |> String.to_atom(); {type, :crypto.strong_rand_bytes(32), :crypto.strong_rand_bytes(32)} end) |> Enum.chunk_every(1000); length(mbs_txs)
```
it takes less than 100ms per generation if all microblocks have the chosen tx type:
```
{tc, having_tx} = :timer.tc(fn -> Enum.map(mbs_txs, fn mb_txs -> Enum.any?(mb_txs, fn {type, _, _} -> type == :"6" end) end) end); {tc, Enum.count(having_tx, &(&1))}
{49657, 1000}
```

Agreed, it's suggested in the additional notes (: For now, due to other investigations on data persistence, it would be good to keep the current behaviour which is to index only for the `:contract_call_tx`

thepiwo:
> Agreed upon other tx types, it's suggested in the additional notes (: For now, due to other investigations on transaction persistence, it would be good to keep the current behaviour which is to index only for the `:contract_call_tx`

are you sure about this? Wasn't the previous logic to check for events in any transaction and this is the performance improvement introduced? I think we should correct it now even if it was incorrect before

jyeshe:
> > Agreed upon other tx types, it's suggested in the additional notes (: For now, due to other investigations on transaction persistence, it would be good to keep the current behaviour which is to index only for the `:contract_call_tx`
> 
> are you sure about this? Wasn't the previous logic to check for events in any transaction and this is the performance improvement introduced? I think we should correct it now even if it was incorrect before

Yes, only for `contract_call_tx` (v1.0.9):
https://github.com/aeternity/ae_mdw/blob/a7175ed73078a378f50647417105acc786c5191b/lib/ae_mdw/db/sync/transaction.ex#L123

image

https://testnet.aeternity.io/mdw/contracts/logs/forward?contract_id=ct_23ip8qhvDpvEyk3tAxnKNTE54Lk2XGbykbcpNUYy5ToJvPcTw6
https://testnet.aeternity.io/mdw/tx/th_24ZcXPfLjjnK2bAHAkAHYaZFj9RcPsqPsxZKVXLx4ytKPfGXyu

In the current version, events from `contract_call_tx` inside a `gameta_tx` or `paying_for` are indexed.

As we are investigating the persistence, from now it would be important if we could only decrease the load in the mnesia transaction to allow better comparison. The fix for #377 would be included in the following minor release (following task after full sync). Could we proceed this way?

jyeshe:
@thepiwo @sborrazas this is a very localized safe change that could be included in the next release, specially for the reason to make mdw mnesia transactions lighter.

Stats of microblocks containing `contract_call_tx` and/or `contract_create_tx`:
```
iex(aeternity@localhost)44> records = :dets.match_object(table, :"$1") |> Enum.sort_by(fn {height, _mbi, _has} -> height end); {List.first(records), List.last(records)}                   
{{1, 0, false}, {500100, 84, false}}
iex(aeternity@localhost)45> :dets.match_object(table, :"$1") |> Enum.count(fn {_h, _mbi, has_ct_tx?} -> has_ct_tx? end)                                       
12784
iex(aeternity@localhost)46> :dets.match_object(table, :"$1") |> Enum.count(fn {_h, _mbi, has_ct_tx?} -> not has_ct_tx? end)                                   
336801
```

sborrazas (approved):

thepiwo (approved):

PR #366, opened 2021-11-30
feat: preload generations and contract events

jyeshe:
## What

Preloads generation blocks and runs contract events with async tasks.

## Why

Contribute to #346 based on loading timings with `CacheTxs.read_txs` and `CacheTxs.read_txs_and_events`:

```
defmodule CacheTxs do
  alias AeMdw.Db.Model

  # import AeMdw.Db.Util, only: [read_block!: 1]
  require Model

  def read_txs(%Range{first: _from_height, last: _to_height} = range) do
      range
      |> Enum.map(fn height ->
        # IO.inspect Model.block(read_block!({height, -1}), :hash)
        {_key_block, micro_blocks} = AeMdw.Node.Db.get_blocks(height)

        micro_blocks
        |> Enum.with_index()
        |> Enum.map(fn {mblock, mbi} ->
          :aec_blocks.txs(mblock)
        end)
      end)
  end

  def read_txs_and_events(%Range{first: _from_height, last: _to_height} = range) do
    range
    |> Enum.map(fn height ->
      # IO.inspect Model.block(read_block!({height, -1}), :hash)
      {_key_block, micro_blocks} = AeMdw.Node.Db.get_blocks(height)

      micro_blocks
      |> Enum.with_index()
      |> Enum.map(fn {mblock, mbi} ->
        %{txs: :aec_blocks.txs(mblock), events: AeMdw.Contract.get_grouped_events(mblock)}
      end)
    end)
  end
end
```

PR #378, opened 2021-12-07
feat: preload blocks when syncing

sborrazas:
Not sure if this is a good alternative to pre-loading chain blocks by using a buffered stream, but we could also use this stream for many other preloading opportunities throughout the code

Thoughts?

### DIFF
--- a/lib/ae_mdw/db/sync/block_index.ex
+++ b/lib/ae_mdw/db/sync/block_index.ex
@@ -20,9 +20,21 @@ defmodule AeMdw.Db.Sync.BlockIndex do
     with true <- max_height >= min_height do
       header = :aec_chain.get_key_header_by_height(max_height) |> ok!
       hash = :aec_headers.hash_header(header) |> ok!
-      syncer = &sync_key_header(~t[block], &1, &2)
+
+      syncer = fn {height, _key_block, _micro_blocks}, acc ->
+        sync_key_header(~t[block], height, acc)
+      end
+
       tracker = Sync.progress_logger(syncer, @log_freq, &log_msg/2)
-      :mnesia.transaction(fn -> max_height..min_height |> Enum.reduce(hash, tracker) end)
+
+      :mnesia.transaction(fn ->
+        max_height..min_height
+        |> Enum.map(fn height ->
+          # Bogus values
+          {height, nil, []}
+        end)
+        |> Enum.reduce(hash, tracker)
+      end)
     end
 
     max_kbi()
--- a/lib/ae_mdw/db/sync/sync.ex
+++ b/lib/ae_mdw/db/sync/sync.ex
@@ -74,9 +74,9 @@ defmodule AeMdw.Db.Sync do
   end
 
   def progress_logger(work_fn, freq, msg_fn) do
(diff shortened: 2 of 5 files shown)


jyeshe (commented):
I would rather use what Erlang and Elixir provides in terms of concurrency and avoid message passing without GenServer.

When the state is stored in the GenServer it's more explicit where the information is coming from and to where it's going (adheres to design by message passing). 

So in terms of design I would be in favor the #374 with the `wait` condition basically as a `GenServer.call` (avoiding custom synchronization). However, I have closed the dratf because after the 150k it became slower than master with the Node previously synced by snapshot. When the Node is busy with crypto and holding transaction locks, this might be different though.

sborrazas:
@jyeshe as per your advice, I've just transformed the semaphore into a genserver and it looks better now (more OTP-ish). I've made two runs from 150_000 to 153_000. With the preload and a buffer size of 100, it took 839s, without the preload it took 940s, so it's a 10% gain, which I think is worth it. We can discuss this in today's meeting

jyeshe:
> 

At the end, benchmark and tests might be used as a criteria. 

Before that, I would like to share few additional comments to detail a personal view on design based on requirements:
- generations loaded sequentially, not in parallel -> need for 1 process (favors out of the box GenServer vs multiple Tasks with max_concurrency > 1 and synchronized to avoid running in parallel)
- younger generations loaded first -> iteration needed (favors out of the box GenServer single explicit state vs stream of multiple Tasks with custom synchronization). 
- buffering -> need for a process state (favors out of the box GenServer vs custom module with custom synchronization)
- producer unbounded by consumer -> produce multiple blocks while consuming one (favors out of the box eager GenServer `:continue` vs concept of Stream which is for lazy processing)

sborrazas:
@jyeshe I've updated the preloading as per your suggestions. Now there's a single task synchronously loading every task until the buffer fills. 

I'd like to introduce this change not only to preload blocks, but also to preload many other things that can be preloaded when syncing (I found one of them can be the microblock events which take a lot to load).

If you have any other suggested improvements, perhaps we can review these changes together, but I feel like we can all benefit from using the BufferedStream. What do you think?

jyeshe:
I think we could put this PR and the other for parallel events (#366) on hold until we have a sync pipeline starting with the chain subscription and replacing the sync process with spawn_link decoupling the listening from loading generations and loading generations from processing them.

jyeshe:
> @jyeshe I've updated the preloading as per your suggestions. Now there's a single task synchronously loading every task until the buffer fills.
> 
> I'd like to introduce this change not only to preload blocks, but also to preload many other things that can be preloaded when syncing (I found one of them can be the microblock events which take a lot to load).
> 
> If you have any other suggested improvements, perhaps we can review these changes together, but I feel like we can all benefit from using the BufferedStream. What do you think?

I think we could put this PR and the other for parallel events (#366) on hold until we have a sync pipeline starting with the chain subscription and replacing the sync process with spawn_link decoupling the listening from loading generations and loading generations from processing them.

sborrazas:
Closing since @jyeshe is working on a different implementation for it

PR #384, opened 2021-12-08
feat: add cursor-based pagination to stats

sborrazas:


sborrazas:
We should merge #383 first

thepiwo (approved):
cool!

lib/ae_mdw/stats.ex
```diff
@@ -0,0 +1,128 @@
...
+    cursor_scope =
+      case deserialize_cursor(cursor) do
+        nil -> nil
+        cursor when direction == :forward -> {cursor, cursor + limit + 1}
+        cursor -> {cursor, cursor - limit - 1}
+      end
+
+    global_scope = if direction == :forward, do: {1, last_gen}, else: {last_gen, 1}
+
+    case intersect_scopes([range_scope, cursor_scope, global_scope], direction) do
```
jyeshe, line 41: We could have a fixed and explicit precedence rule. Besides making clear what is the criteria applied in the code it helps to get more clear results because 
http://localhost:4000/stats/gen/900-2000?cursor=1100&limit=100
uses the range scope while 
http://localhost:4000/stats/gen/1100-2000?cursor=1100&limit=100
uses the cursor.

jyeshe (changes requested):
Code looks good. Besides the explicit precedence rule, most importantly would be to add tests with cursor.

lib/ae_mdw/stats.ex
sborrazas: There's no precedence, it's the intersection. With cursor = 1100 and scope = 900-2000 we get 1100-2000. This way, when you request range 900-2000, on the second page (with limit = 100) you will get cursor = 1000, with the same scope the user requested initially

jyeshe: Isn't it better to replace intersection with fixed criteria. It may cause confusion when someone looks at the URL and need to solve conflicts to know what to expect.

An quick alternative would be to generate the next URL only with the cursor . The range would be only for backwards compatibility (without cursor in the first request) and it would avoid the conflict between path and params that defines two ranges in the same request.

sborrazas: These endpoints assume the user will never (it's not part of the API) modify, or play around with the `cursor` parameter. We've already had this discussion: the user is exclusively allowed to use the `next` URL, whatever it may contain. If we decide tomorrow to name the `cursor` parameter "foo", we should be able to do so without any breaking functionality.

sborrazas: We can also base64 the cursor parameter to make it more evident that it's not something modifiable.

thepiwo: yeah maybe we should make the cursor not so presentable for devs to abuse it

thepiwo: but lets discuss this in a separate issue https://github.com/aeternity/ae_mdw/issues/386

jyeshe: @sborrazas The point here is not about editing but a simple change that would remove conflicting info. The fact that the dev should just program the request to the `next` it as is doesn't imply that we can encode it in any way (for the sake of argument). Added as *suggestion* in the #386.

jyeshe (approved):

PR #385, opened 2021-12-09
chore: remove name expiration cleanup

jyeshe:
## What

Removes cleanup name expiration code that slows down the sync with a full search. 

## Why

This recent validation already protects from old expirations:
https://github.com/aeternity/ae_mdw/blob/72c500109cfb15afb11a17fdaafe7ef920d5a5a5/lib/ae_mdw/db/name.ex#L68

### DIFF
--- a/lib/ae_mdw/db/name.ex
+++ b/lib/ae_mdw/db/name.ex
@@ -104,7 +104,6 @@ defmodule AeMdw.Db.Name do
     m_name_exp = Model.expiration(index: {expire, plain_name})
     m_owner = Model.owner(index: {owner, plain_name})
 
-    ensure_no_exsiting_active_name_expiration(plain_name)
     cache_through_write(Model.ActiveName, m_name)
     cache_through_write(Model.ActiveNameOwner, m_owner)
     cache_through_write(Model.ActiveNameExpiration, m_name_exp)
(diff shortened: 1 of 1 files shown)


lib/ae_mdw/db/name.ex
```diff
@@ -104,7 +104,7 @@ defmodule AeMdw.Db.Name do
     m_name_exp = Model.expiration(index: {expire, plain_name})
     m_owner = Model.owner(index: {owner, plain_name})
 
-    ensure_no_exsiting_active_name_expiration(plain_name)
+    clean_previous_expirations(previous || [])
```
sborrazas, line 107: Isn't `previous` here the result of an `{:ok, inactive_name}` tuple (not a list)?

jyeshe: It returns `nil` when there is not record (with `map_one_nil`).

thepiwo (approved):

lib/ae_mdw/db/name.ex
sborrazas: I was referring to the other case, what happens when it's not `nil`? Is it a list?

jyeshe: correct and exist only when name is reactivated (doesn't cover extensions) so better to remove it and rely on:
https://github.com/aeternity/ae_mdw/blob/72c500109cfb15afb11a17fdaafe7ef920d5a5a5/lib/ae_mdw/db/name.ex#L68

thepiwo (approved):

Issue #387, opened 2021-12-09
Mnesia transaction code removal


sborrazas:
Right now the mnesia transaction deals with logic that can be moved outside of it for efficiency. In particular, we should be able to remove most of the calls that are made to the core modules, which potentially create exceptions that would interfere with mnesia's transaction.

Ideally, the code inside the mnesia transaction should only be a sequence of writes.

jyeshe:
We have been working on it divided into 2 fronts (at least):
i) mutations
ii) preloading the data needed for the mutations

The #366 is a WIP for the preloading.

jyeshe:
I wouldn't consider this an enhancement since it's an issue that affects the commit of the transaction.

Issue #211, opened 2021-08-18
sync AEX-9 transactions instantly


thepiwo:

thepiwo:
relates to #749, should we close one or the other @sborrazas @jyeshe ?

jyeshe:
> relates to #749, should we close one or the other @sborrazas @jyeshe ?

If a best effort approach with a single machine is not considered enough, I would leave it separate because depending on the size of the queue and frequency of updates, in order to get faster results it might be needed to have some cluster dedicated to dry-running.

thepiwo:
so I'll close this in favour of the other task, if we come up with a new approach we can still make a new issue

PR #390, opened 2021-12-10
feat: db transactions per microblock

jyeshe:
## What

Commits sync changes per micro block.

## Why

Mnesia transaction log is not being fully committed.
Contribute to #387
Contribute to #396 

## Additional notes

This paves the way to make instant syncs of AEX9 txs #211. Not sufficient but one step forward to the next step of having syncs per transaction.

### DIFF
--- a/.github/workflows/test-action.yml
+++ b/.github/workflows/test-action.yml
@@ -69,7 +69,7 @@ jobs:
         id: plt-cache
         with:
           path: priv/plts
-          key: plts-v4-${{ hashFiles('mix.lock') }}
+          key: plts-v5-${{ hashFiles('mix.lock') }}
 
       - name: Common tests setup
         uses: ./.github/actions/tests-common
--- a/lib/ae_mdw/db/int_transfer.ex
+++ b/lib/ae_mdw/db/int_transfer.ex
@@ -8,6 +8,7 @@ defmodule AeMdw.Db.IntTransfer do
   alias AeMdw.Db.Model
   alias AeMdw.Mnesia
 
+  require Ex2ms
   require Model
 
   alias AeMdw.Db.BlockRewardsMutation
@@ -24,6 +25,9 @@ defmodule AeMdw.Db.IntTransfer do
 
   @fee_kinds [:lock_name, :spend_name, :refund_name, :earn_oracle]
 
+  @reward_block_kind "reward_block"
+  @reward_dev_kind "reward_dev"
+
   @spec block_rewards_mutation(Blocks.height(), Blocks.key_header(), Blocks.key_hash()) ::
           BlockRewardsMutation.t()
   def block_rewards_mutation(height, key_header, key_hash) do
@@ -37,7 +41,7 @@ defmodule AeMdw.Db.IntTransfer do
       |> :aeu_mtrees.to_list()
       |> Enum.map(fn {target_pk, ser_account} ->
         amount = :aec_accounts.balance(:aec_accounts.deserialize(target_pk, ser_account))
-        kind = (target_pk in dev_benefs && "reward_dev") || "reward_block"
+        kind = (target_pk in dev_benefs && @reward_dev_kind) || @reward_block_kind
 
         {kind, target_pk, amount}
       end)
@@ -69,5 +73,28 @@ defmodule AeMdw.Db.IntTransfer do
     Mnesia.write(Model.IntTransferTx, int_tx)
     Mnesia.write(Model.KindIntTransferTx, kind_tx)
     Mnesia.write(Model.TargetKindIntTransferTx, target_kind_tx)
+  end
+
+  @spec count_block_reward(Blocks.height()) :: pos_integer()
+  def count_block_reward(height) do
+    count_reward(height, @reward_block_kind)
+  end
+
+  @spec count_dev_reward(Blocks.height()) :: pos_integer()
+  def count_dev_reward(height) do
+    count_reward(height, @reward_dev_kind)
+  end
+
+  defp count_reward(height, kind) do
+    height_pos = {height, -1}
+
+    kind_spec =
+      Ex2ms.fun do
+        {:kind_int_transfer_tx, {^kind, ^height_pos, :_, :_}} -> 1
+      end
+
+    Model.KindIntTransferTx
+    |> :mnesia.dirty_select(kind_spec)
+    |> length()
   end
 end
--- a/lib/ae_mdw/db/mutations/stats_mutation.ex
+++ b/lib/ae_mdw/db/mutations/stats_mutation.ex
@@ -4,6 +4,7 @@ defmodule AeMdw.Db.StatsMutation do
   """
 
   alias AeMdw.Blocks
+  alias AeMdw.Db.IntTransfer
   alias AeMdw.Db.Model
   alias AeMdw.Db.Util
   alias AeMdw.Ets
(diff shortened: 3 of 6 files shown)


thepiwo (approved):

thepiwo (commented):
as discussed in standup we can't split transactions for now as our sync is key-block based not micro-block based, right?

jyeshe:
> as discussed in standup we can't split transactions for now as our sync is key-block based not micro-block based, right?

Since the mdw uses a key-value DB most writes operations are idempotent from what I have looked but I will review it again and let it sync to see if there is no `:not_found` issue after it.

jyeshe:
@thepiwo @sborrazas 
The PR is updated with the sync continuing from last microblock synced in case you wanna trigger a new test sync. 

Saving the keyblock at the end fixes the `/blocks/` (shows all microblocks of the generation).  

So from what it's tracked so far, there is only a pending change on async task that needs to be enqueued only after the microblock is commited.

I got only one error on multiple restarts that already existed (and will be fixed in the next commit):
```
{#{'__exception__' => true,'__struct__' => 'Elixir.ArgumentError',message => <<"got empty list">>},[{'Elixir.AeMdw.Util','one!',1,[{file,"lib/ae_mdw/util.ex"},{line,5}]},{'Elixir.AeMdw.Db.Sync.Stat',store_mutation,1,[{file,"lib/ae_mdw/db/sync/stat.ex"},{line,17}]},{'Elixir.AeMdw.Db.Sync.Transaction',sync_generation,2,[{file,"lib/ae_mdw/db/sync/transaction.ex"},{line,162}]},{'Elixir.Enum',reduce_range_inc,4,[{file,"lib/enum.ex"},{line,3371}]},{'Elixir.AeMdw.Db.Sync.Transaction',sync,3,[{file,"lib/ae_mdw/db/sync/transaction.ex"},{line,57}]},{'Elixir.AeMdw.Db.Sync',run_action,1,[{file,"lib/ae_mdw/db/sync/sync.ex"},{line,99}]}]}
```

thepiwo:
what happens in case of micro-forks? e.g. Miner A produces a keyblock then follows with mb1 -> mb2 -> mb3 then Miner B produces a keyblock but only picks up the microblocks from mb2 and reprocesses the txs from mb3 in his new micro block?

jyeshe:
> what happens in case of micro-forks? e.g. Miner A produces a keyblock then follows with mb1 -> mb2 -> mb3 then Miner B produces a keyblock but only picks up the microblocks from mb2 and reprocesses the txs from mb3 in his new micro block?

This is the point I have just commented in the meeting:
https://github.com/aeternity/ae_mdw/blob/ca7f295d9de201c4dd45cdb03dfcddce86b39e21/lib/ae_mdw/db/sync/sync.ex#L33-L35

If there is a running sync process it does not run the fork concurrently and waits for the process with multiple microblocks commit to finish.

When the sync process finishes, then the fork runs here:
https://github.com/aeternity/ae_mdw/blob/ca7f295d9de201c4dd45cdb03dfcddce86b39e21/lib/ae_mdw/db/sync/sync.ex#L44-L45

jyeshe:
WIP: Stats needs a cache reload in case the `last_mbi` is not -1

jyeshe:
Depends on #403

jyeshe:
@thepiwo @sborrazas now with #403 this PR is ready for review and full sync test.
Sum stat reading might still happen if mdw is stopped (not always but stopping and starting multiple times reproduces it).
Investigation and enhancement issue reported [here](https://github.com/aeternity/ae_mdw/issues/404)

thepiwo:
> what happens in case of micro-forks? e.g. Miner A produces a keyblock then follows with mb1 -> mb2 -> mb3 then Miner B produces a keyblock but only picks up the microblocks from mb2 and reprocesses the txs from mb3 in his new micro block?

I still don't understand how that is handled here, what concretely happens to the info and txs that were eventually processed already from mb3 in this example?

jyeshe:
> > what happens in case of micro-forks? e.g. Miner A produces a keyblock then follows with mb1 -> mb2 -> mb3 then Miner B produces a keyblock but only picks up the microblocks from mb2 and reprocesses the txs from mb3 in his new micro block?
> 
> I still don't understand how that is handled here, what concretely happens to the info and txs that were eventually processed already from mb3 in this example?

It means that the sequence of sync and fork invalidation and the fork process itself are kept unchanged. The fork will execute at the same moment as it would before this change.
The only difference will be that instead of the sync process have a big mnesia transaction it will have multiple smaller ones.

thepiwo:
@jyeshe so we are not processing micro-blocks once received, still just when the key-block is final?

jyeshe:
> @jyeshe so we are not processing micro-blocks once received, still just when the key-block is final?

@thepiwo good question, exactly. 

The full sequence is (change only in the 5th step):
1. mdw is notified that there is something to be synced
2. mdw calls the node to get the top height and spawns a process to sync until this top height less 200 generations.
3. for each height until the last to be synced, it loads the key block and all microblocks walking from the last to the first microblock.
4. reverses the list of microblocks and syncs the generation (the key block and all microblocks).
5. while syncing a generation, the mdw commits the txs for each microblock (instead of committing all txs of the generation at once).

jyeshe:
> > @jyeshe so we are not processing micro-blocks once received, still just when the key-block is final?
> 
> @thepiwo good question, exactly.
> 
> The full sequence is (the scope of the PR is in the 5th step):
> 
>     1. mdw is notified that there is something to be synced
> 
>     2. mdw calls the node to get the top height and spawns a process to sync until this top height less 200 generations.
> 
>     3. for each height until the last to be synced, it loads the key block and all microblocks walking from the last to the first microblock.
> 
>     4. reverses the list of microblocks and syncs the generation (the key block and all microblocks).
> 
>     5. while syncing a generation, the mdw commits the txs for each microblock (instead of committing all txs of the generation at once).

PS: 
The top height - 200 is used only when the Mdw starts (`:safe` param):
https://github.com/aeternity/ae_mdw/blob/27d7fbe509ad219460aab13ee26c0c6c5f3274fb/lib/ae_mdw/db/sync/sync.ex#L31 

On the next chain notifications to the subscriber, it gets the top height - 1:
https://github.com/aeternity/ae_mdw/blob/27d7fbe509ad219460aab13ee26c0c6c5f3274fb/lib/ae_mdw/db/sync/sync.ex#L85

jyeshe:
@thepiwo a complement:
The top height - 200 is used only when the Mdw starts (`:safe` param):
https://github.com/aeternity/ae_mdw/blob/27d7fbe509ad219460aab13ee26c0c6c5f3274fb/lib/ae_mdw/db/sync/sync.ex#L31 

On the next chain notifications to the subscriber, it gets the top height - 1:
https://github.com/aeternity/ae_mdw/blob/27d7fbe509ad219460aab13ee26c0c6c5f3274fb/lib/ae_mdw/db/sync/sync.ex#L85

thepiwo:
I understand, makes sense I think

thepiwo (approved):

jyeshe:
rebased for 387-mb-commit-with-stat-fix

jyeshe:
@thepiwo @sborrazas the Model.Block `tx_index` was saved previously after a range of generations. Now it's saved after each generation. With this, the mdw knows if the generation was fully synced. 

The other change was to resync in idempotent manner the generation of the last tx synced (last microblock synced).

sborrazas (approved):
LGTM

PR #391, opened 2021-12-12
test: add contract controller endpoints integration tests

sborrazas:


thepiwo (approved):

test/integration/ae_mdw_web/controllers/contract_controller_test.exs
```diff
@@ -2,124 +2,136 @@ defmodule Integration.AeMdwWeb.ContractControllerTest do
...
+               conn |> get("/contracts/logs") |> json_response(200)
+
+      call_txis =
+        logs
+        |> Enum.map(fn %{"call_txi" => call_txi} -> call_txi end)
+        |> Enum.reverse()
 
-      assert response = json_response(conn, 200)
-      assert Enum.count(response["data"]) == @default_limit
-      assert Jason.encode!(response["data"]) == get_contract_logs_json(@contract0)
```
jyeshe, line 32: This asserts that the logs are correct. Could be safely removed though when tests are added to the `AeMdw.Contracts`.

test/integration/ae_mdw_web/controllers/contract_controller_test.exs
```diff
@@ -2,124 +2,136 @@ defmodule Integration.AeMdwWeb.ContractControllerTest do
...
+      assert @default_limit = length(call_txis)
+      assert ^call_txis = Enum.sort(call_txis)
```
jyeshe, line 22: extra assert for order :+1:

test/integration/ae_mdw_web/controllers/contract_controller_test.exs
```diff
@@ -2,124 +2,136 @@ defmodule Integration.AeMdwWeb.ContractControllerTest do
...
-      assert response_next["data"] != response["data"]
+      next_create_txis = Enum.map(next_logs, fn %{"contract_txi" => create_txi} -> create_txi end)
 
-      Enum.each(response_next["data"], fn log_next ->
-        Enum.each(response["data"], fn log -> assert log_next["call_txi"] < log["call_txi"] end)
-      end)
+      assert @default_limit = length(next_create_txis)
+      assert ^next_create_txis = Enum.sort(next_create_txis)
+      assert Enum.all?(next_logs, &match?(%{"contract_id" => ^contract_id}, &1))
+      assert Enum.at(create_txis, @default_limit - 1) <= Enum.at(next_create_txis, 0)
```
jyeshe, line 77: really next :pray:

jyeshe (commented):
Nice test coverage. Only would keep `the get_contract_logs_json` or add tests to the module that retrieves the logs (and calls) to assure that the paginated info is complete and correct (by filter and other fields).

test/integration/ae_mdw_web/controllers/contract_controller_test.exs
sborrazas: We're not doing this for any other resource in the codebase, perhaps you can provide a sample test with all of the mocking that `AeMdw.Contracts` requires?

test/integration/ae_mdw_web/controllers/contract_controller_test.exs
```diff
@@ -131,27 +143,205 @@ defmodule Integration.AeMdwWeb.ContractControllerTest do
...
+
+      assert length(data_call_txis) > 0
+      assert ^data_call_txis = Enum.sort(data_call_txis)
+      assert Enum.all?(logs, fn %{"data" => data} -> String.starts_with?(data, data_prefix) end)
     end
   end
 
   describe "calls" do
-    test "it gets calls from a contract", %{conn: conn} do
+    test "it get logs backwards without any filters", %{conn: conn} do
```
jyeshe, line 196: small description typo: logs -> calls

test/integration/ae_mdw_web/controllers/contract_controller_test.exs
jyeshe: True, I have added because the controller was doing only pagination validation and not validating the data fields values.

In theory, we should have for logs validated per:
```
         {:ok, create_txi} <- deserialize_cursor_int(create_txi_bin),
         {:ok, call_txi} <- deserialize_cursor_int(call_txi_bin),
         {:ok, event_hash} <- deserialize_cursor_string(event_hash_bin),
         {:ok, log_idx} <- deserialize_cursor_int(log_idx_bin),
         {:ok, data} <- deserialize_cursor_data(data_bin) do
```
and calls per:
```
        {:ok, call_txi} <- deserialize_cursor_int(call_txi_bin),
         {:ok, local_idx} <- deserialize_cursor_int(local_idx_bin),
         {:ok, create_txi} <- deserialize_cursor_int(create_txi_bin),
         {:ok, pk} <- deserialize_cursor_string(pk_bin),
         {:ok, fname} <- deserialize_cursor_string(fname_bin),
         {:ok, pos} <- deserialize_cursor_int(pos_bin) do
```
but some of these filters are already tested (like event_hash) so maybe we could add tests for the remaining filters in this controller test. If there are no urgent tasks, would add for the `AeMdw.Contracts` only two simple test cases validating the structure:
1. fetch_logs
2. fetch_calls

sborrazas: I still don't understand what it is that you would like to test, can you please provide some **sample code** to understand better?

jyeshe: 1. For `/contracts/logs/*`
There is this test to filter logs by event hash:
https://github.com/aeternity/ae_mdw/blob/38595177b7329126fe05c9c6b40f1afc2a9042fa/test/integration/ae_mdw_web/controllers/contract_controller_test.exs#L151
this by data:
https://github.com/aeternity/ae_mdw/blob/38595177b7329126fe05c9c6b40f1afc2a9042fa/test/integration/ae_mdw_web/controllers/contract_controller_test.exs#L176

Could we have test cases for the other kind of cursors please? 
```
         {:ok, create_txi} <- deserialize_cursor_int(create_txi_bin), # this maybe is already covered by contract_id param
         {:ok, call_txi} <- deserialize_cursor_int(call_txi_bin),
         {:ok, log_idx} <- deserialize_cursor_int(log_idx_bin),
```
from https://github.com/aeternity/ae_mdw/blob/b2a82e08e29bac432d10271db94f685eed087bf6/lib/ae_mdw/contracts.ex#L451-L455

2. For `/contracts/calls/*`
To have test cases for each kind of cursor decoded here:
https://github.com/aeternity/ae_mdw/blob/b2a82e08e29bac432d10271db94f685eed087bf6/lib/ae_mdw/contracts.ex#L479-L484

3. For `AeMdw.Contracts`
The tests would validate the logs and calls data (filtering and pagination not strictly needed as they are tested by the controller). The pseudo-code for fields validation:
i) for logs
```
create 1 log
read the log
check that it was rendered with all fields and these fields have the correct data
```
ii) same sequence for calls

jyeshe: I'm sorry too, now I am seeing that the cursor is different this time, with multiple fields even though not used in the original search because the log record key is a complex one (all these fields).

The tests for filtering `contract/calls` would be complete too. Thanks.

To conclude and move on, could we add only for `/contract/logs/backward` and `/contracts/calls/backward`?
I understand without the word `backward` it works and it's tested but just to cover the path param.

sborrazas: Done

test/integration/ae_mdw_web/controllers/contract_controller_test.exs
sborrazas: Done

jyeshe (approved):

PR #392, opened 2021-12-13
feat: add cursor-based pagination to contract logs/calls

sborrazas:


sborrazas:
Requires #391 to be merged first

thepiwo (approved):

jyeshe (changes requested):
Code looks good but safer to be validated (and continue to work without regression) with unit tests for the module `AeMdw.Contracts`.

jyeshe (approved):

PR #395, opened 2021-12-13
feat: index contract init events and internal calls

### LINKED ISSUE #377: index `:contract_create_tx` events
Currently the events are logged only for contract calls transactions:
https://github.com/aeternity/ae_mdw/blob/ee3465822fecf346acc564e36f9e209faf8444f8/lib/ae_mdw/db/sync/transaction.ex#L82

jyeshe:
## Why
Resolve #377 

## Validation steps
elixir --sname aeternity@localhost -S mix test.integration test/integration/ae_mdw_web/controllers/contract_controller_test.exs

### DIFF
--- a/lib/ae_mdw/contract.ex
+++ b/lib/ae_mdw/contract.ex
@@ -306,13 +306,18 @@ defmodule AeMdw.Contract do
     end
   end
 
-  def get_init_call_details(contract_pk, tx_rec, block_hash) do
+  def get_init_call_rec(contract_pk, tx_rec, block_hash) do
     create_nonce = :aect_create_tx.nonce(tx_rec)
 
     tx_rec
     |> :aect_create_tx.owner_pubkey()
     |> :aect_call.id(create_nonce, contract_pk)
     |> call_rec_from_id(contract_pk, block_hash)
+  end
+
+  def get_init_call_details(contract_pk, tx_rec, block_hash) do
+    contract_pk
+    |> get_init_call_rec(tx_rec, block_hash)
     |> :aect_call.serialize_for_client()
     |> Map.drop(["gas_price", "height", "caller_nonce"])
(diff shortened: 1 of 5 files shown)


lib/ae_mdw/db/sync/transaction.ex
```diff
@@ -82,11 +82,16 @@ defmodule AeMdw.Db.Sync.Transaction do
...
+      case type do
+        :contract_create_tx -> :aect_create_tx.contract_pubkey(tx)
+        :contract_call_tx -> :aect_call_tx.contract_pubkey(tx)
+        _other_tx -> nil
+      end
+
     contract_events_mutation =
-      if type == :contract_call_tx do
-        ct_pk = :aect_call_tx.contract_pubkey(tx)
+      if not is_nil(ct_pk) do
```
sborrazas, line 93: I think an `if ct_pk` should do here

lib/ae_mdw/contract.ex
```diff
@@ -306,13 +306,18 @@ defmodule AeMdw.Contract do
     end
   end
 
-  def get_init_call_details(contract_pk, tx_rec, block_hash) do
+  def get_init_call_rec(contract_pk, tx_rec, block_hash) do
```
sborrazas, line 309: We should probably add typespecs to every new public function we create

sborrazas (changes requested):

lib/ae_mdw/db/sync/transaction.ex
jyeshe: glad you prefer shorter too

lib/ae_mdw/contract.ex
jyeshe: added for this and remaining functions of the module to enable credo/spec check

thepiwo (approved):

sborrazas (approved):

PR #400, opened 2021-12-15
chore: prev_stat is not used

jyeshe:
## What / Why

`prev_stat` is read but all fields are overwritten (which is correct because only sum stat depends on previous sum).

## Additional notes

Does not fix the `store_mutation` error but clears the path to continue working on it.


lib/ae_mdw/db/mutations/stats_mutation.ex
```diff
@@ -32,13 +35,11 @@ defmodule AeMdw.Db.StatsMutation do
   @spec mutate(t()) :: :ok
   def mutate(%__MODULE__{
         height: height,
-        prev_stat: prev_stat,
         prev_sum_stat: prev_sum_stat,
         token_supply_delta: token_supply_delta
       }) do
     stat =
-      Model.stat(prev_stat,
```
jyeshe, line 40: all fields are set (no field used from prev_stat)

lib/ae_mdw/db/mutations/stats_mutation.ex
```diff
@@ -5,25 +5,28 @@ defmodule AeMdw.Db.StatsMutation do
...
 
   @opaque t() :: %__MODULE__{
             height: Blocks.height(),
-            prev_stat: Model.stat(),
             prev_sum_stat: Model.sum_stat(),
             token_supply_delta: integer()
           }
 
-  @spec new(Blocks.height(), Model.stat(), Model.sum_stat(), integer()) :: t()
-  def new(height, prev_stat, prev_sum_stat, token_supply_delta) do
```
sborrazas, line 23: My reasoning behind not doing **any** sort of computation inside the mutation was to be able in the future to run in-memory mutations without having to mock the core or any other external function. IMO, we should try to reach a state where mutations only perform mnesia writes

lib/ae_mdw/db/mutations/stats_mutation.ex
sborrazas: Good catch, thanks!

lib/ae_mdw/db/mutations/stats_mutation.ex
jyeshe: Good goal. Here it cannot be done because it needs the mutations to be executed before getting the results from the cache.

jyeshe: But with microblocks transaction it will be (:

sborrazas: Sounds good!

sborrazas (approved):

thepiwo (approved):

PR #403, opened 2021-12-16
feat: read stats from db when resuming mb sync

jyeshe:
## What

Calculates the stats using database info when sync is resumed not from the first microblock of the generation.

## Why

In this case the cache doesn't have data from the microblocks that were already synced (only in the first `AeMdw.Db.Sync.Transaction.sync_generation` call).

## Additional notes

Base branch for #390 (transactions per microblock) requires this change to save stats properly. 

@thepiwo @sborrazas Feel free to merge this if you prefer to review this together with base branch code.

### DIFF
--- a/lib/ae_mdw/db/mutations/stats_mutation.ex
+++ b/lib/ae_mdw/db/mutations/stats_mutation.ex
@@ -4,6 +4,7 @@ defmodule AeMdw.Db.StatsMutation do
   """
 
   alias AeMdw.Blocks
+  alias AeMdw.Db.IntTransfer
   alias AeMdw.Db.Model
   alias AeMdw.Db.Util
   alias AeMdw.Ets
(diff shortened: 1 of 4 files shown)


thepiwo (approved):
looks good, haven't checked the code

lib/ae_mdw/db/mutations/stats_mutation.ex
```diff
@@ -4,79 +4,115 @@ defmodule AeMdw.Db.StatsMutation do
...
-  @spec new(Blocks.height()) :: t()
-  def new(height) do
-    token_supply_delta = AeMdw.Node.token_supply_delta(height + 1)
-
-    prev_sum_stat = get_sum_stat(height)
+  @spec new(Blocks.height(), boolean()) :: t()
+  def new(height, all_cached?) do
+    m_stat = make_stat(height + 1, all_cached?)
+    Model.stat(block_reward: inc_block_reward, dev_reward: inc_dev_reward) = m_stat
+    m_sum_stat = make_sum_stat(height + 1, inc_block_reward, inc_dev_reward)
```
sborrazas, line 26: Can we instead, move these computations outside of the `new` function and into some private function in the `Transaction` module? This way, `new` function only does some very simple struct building. Mutations should only be have the logic for writing to mnesia on the `mutate` function

sborrazas (changes requested):

lib/ae_mdw/db/mutations/stats_mutation.ex
jyeshe: Sure, for a shallower and lighter `new`, a factory was added as requested.
Regarding the writing, this PR already moves reading logic out of the `mutate/1` leaving it to perform only `Mnesia` writes as by design.

sborrazas (approved):

Issue #388, opened 2021-12-09
Syncing process failure recovery


sborrazas:
Right now, for every exception thrown in the syncing process, the middleware completely stops syncing. Most of these exceptions are gone when retrying, so we should look into mechanisms to do so without having to manually restart the server.

We should also try to implement as much of #387 as possible, so that we reduce the amount of exceptions that happen inside the mnesia transaction.

jyeshe:
Leaving the #387 with it's own scope, this is in progress with a refactoring of the sync pipeline.

jyeshe:
Leaving the #387 with it's own scope, this is in progress with a refactoring of the sync pipeline so that when sync process ends with a reason other than `:normal` doesn't put the subscriber down and reprocessing the height makes the sync to continue from last synced (until max_height).

jyeshe:
Sync is not waiting for the migrations:
```
2022-03-14 11:34:02.642 [info] syncing block index at 340000
2022-03-14 11:34:09.258 [info] Indexed 144799 records in 80s
2022-03-14 11:34:09.272 [info] applied version 20220223222120
2022-03-14 11:34:09.290 [debug] tell_controllers(:app_running, :ae_mdw, :aeternity@localhost)
2022-03-14 11:34:15.220 [info] syncing block index at 330000
```

PR #406, opened 2021-12-17
refactor: remove dep from chain subscriber

jyeshe:
## What

1. Removes dependency from `AeMdw.Db.Sync` (renamed to `AeMdw.Db.Sync.ChainSubscriber`) in such a way that when `AeMdw.Db.Sync.Transaction` and `AeMdw.Db.Sync.BlockIndex` receives the `max_height` it's fully resolved/calculated.
2. Simplify subscriber moving out functions that are not strictly related to the subscription notifications (likewise in a pipeline, what happens next stays out of the subscriber).

## Why

This is a base refactor to introduce sync auto-restart in order to fix sync issues that doesn't depend on mdw database state.
refs #388 

## Additional notes

Easier to see changes on subscriber with (one of the cases when git doesn't track changes after a `git mv`): 
`git diff master:lib/ae_mdw/db/sync/sync.ex lib/ae_mdw/db/sync/chain_subscriber.ex`


thepiwo (approved):
lgtm

sborrazas (approved):

PR #421, opened 2021-12-23
test: refactor name controller integration tests

sborrazas:
The purpose of this is to test pagination properly, with the added
benefi that many of the tests would fail when there's name
inconsistencies in the DB.

PR #423, opened 2021-12-24
test: restructure oracles integration tests

sborrazas:
The purpose of this is to test pagination properly, with the added
benefit that many of the tests would fail when there's name
inconsistencies.

Requires #421 to be merged first.

### DIFF
--- a/.github/workflows/test-action.yml
+++ b/.github/workflows/test-action.yml
@@ -69,7 +69,7 @@ jobs:
         id: plt-cache
         with:
           path: priv/plts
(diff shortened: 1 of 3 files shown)


thepiwo (approved):

jyeshe (changes requested):
Is there any failing right now for you? If so could you comment which ones and why they fail please?

sborrazas:
@jyeshe no failures on oracles. Only still getting failures on names.

jyeshe (approved):
all tests passed

PR #426, opened 2021-12-28
perf: create ContractCreateMutation

sborrazas:
This allows to compute all computation needed before the mnesia
transaction.

refs #346


thepiwo (approved):

lib/ae_mdw/db/mutations/origin_mutation.ex
```diff
@@ -0,0 +1,41 @@
...
+  def new(tx_type, pubkey, txi, tx_hash) do
+    %__MODULE__{tx_type: tx_type, pubkey: pubkey, txi: txi, tx_hash: tx_hash}
+  end
+
+  @spec mutate(t()) :: :ok
+  def mutate(%__MODULE__{tx_type: tx_type, pubkey: pubkey, txi: txi, tx_hash: tx_hash}) do
+    m_origin = Model.origin(index: {tx_type, pubkey, txi}, tx_id: tx_hash)
+    m_rev_origin = Model.rev_origin(index: {txi, tx_type, pubkey})
+    Mnesia.write(Model.Origin, m_origin)
+    Mnesia.write(Model.RevOrigin, m_rev_origin)
```
jyeshe, line 33: This seems like an opportunity for reusing `AeMdw.Db.MnesiaWriteMutation` (with `origin_mutations` creating one for `Model.Origin` and another for `Model.RevOrigin`).

lib/ae_mdw/db/sync/transaction.ex
```diff
@@ -223,6 +226,61 @@ defmodule AeMdw.Db.Sync.Transaction do
...
+      {:ok, contract_info} ->
+        call_rec = Contract.get_init_call_rec(contract_pk, tx, block_hash)
+
+        aex9_meta_info =
+          if Contract.is_aex9?(contract_info) do
+            Contract.aex9_meta_info(contract_pk)
+          end
+
+        [
+          mutations,
```
jyeshe, line 258: This is harmless because of the flatten happening outside but to decoupling might be better to append to the mutations (append to a small list of only 2 or 3 items).
I was going to propose putting it to the head of `mutations` list (like line 244) but `ContractCreateMutation.mutate` has a `DBContract.logs_write` call which depends on `Origin.tx_index` so better keep the same order as it's already done.

lib/ae_mdw/db/sync/transaction.ex
```diff
@@ -223,6 +226,61 @@ defmodule AeMdw.Db.Sync.Transaction do
...
+         _block_index,
+         block_hash,
+         mb_events
+       ) do
+    contract_pk = :aect_create_tx.contract_pubkey(tx)
+    owner_pk = :aect_create_tx.owner_pubkey(tx)
+    events = Map.get(mb_events, tx_hash, [])
+
+    mutations = [
+      ContractEventsMutation.new(contract_pk, events, txi)
```
jyeshe, line 244: Would leave this outside of `tx_mutations` for being specific for `:contract_create_tx` and `:contract_call_tx` (to help decreasing the signature and make it clear which kinds of txs depend on the events).

jyeshe (changes requested):
Nice move to decrease the risk of exceptions during `:mnesia` transactions. 
Few refactor opportunities.

lib/ae_mdw/db/mutations/origin_mutation.ex
sborrazas: Done

lib/ae_mdw/db/sync/transaction.ex
sborrazas: Done

lib/ae_mdw/db/sync/transaction.ex
sborrazas: Not sure what you meant by this? `code >>>>> words`

jyeshe (approved):

lib/ae_mdw/db/sync/transaction.ex
jyeshe: The proposal is to shorten the number of params of `tx_mutations` (sorry old saying from C, function signature) and keep the handling of `ContractEventsMutation` where it was. With also the removal of `signed_tx`, in code terms it would become like this:
```
tx_mutations(
         :contract_create_tx,
         tx,
         txi,
         tx_hash,
         _block_index,
         block_hash
       )
```
Beyond making the function call simpler for other code readers, most importantly, it keeps it clear where the `mb_events` and `signed_tx` are used by being explicit (avoid searching in the different mutations how/when it's used).

PS: From preexisting code, following this same reasoning, all microblocks shouldn't be passed to a `transaction_mutations`. A better parameter would be `tx_events` as a tx doesn't look for events from other tx.

jyeshe:
for future reference, this was a refactor to change code organization not for perf.

PR #427, opened 2021-12-28
test: restructure tx_controller integration tests

sborrazas:
The purpose of this is to be able to test pagination properly on all
tx endpoints.

### DIFF
--- a/test/integration/ae_mdw_web/controllers/tx_controller_test.exs
+++ b/test/integration/ae_mdw_web/controllers/tx_controller_test.exs
@@ -6,11 +6,8 @@ defmodule Integration.AeMdwWeb.TxControllerTest do
   alias AeMdw.Db.Model
   alias AeMdw.Db.Name
   alias AeMdw.Db.Util
-  alias AeMdwWeb.Continuation
   alias AeMdwWeb.TxController
   alias :aeser_api_encoder, as: Enc
-  alias AeMdwWeb.Util, as: WebUtil
-  alias Plug.Conn
 
   require Model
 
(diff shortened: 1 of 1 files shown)


thepiwo (approved):

test/integration/ae_mdw_web/controllers/tx_controller_test.exs
```diff
@@ -164,78 +161,45 @@ defmodule Integration.AeMdwWeb.TxControllerTest do
...
-      Enum.each(response["data"], fn data ->
-        assert data["tx_index"] in 0..(limit - 1)
-      end)
-
-      {:ok, data, _has_cont?} =
-        Continuation.response_data({TxController, :txs, %{}, build_scope(conn.assigns), 0}, limit)
+      txis = Enum.map(txs, fn %{"tx_index" => tx_index} -> tx_index end)
 
-      assert ^data = response["data"]
+      assert ^txis = Enum.to_list(0..(limit - 1))
```
jyeshe, line 170: Could we add an assertion for the first hashes too? It would helpful to detect the sync is performed equally across multiple resyncs.

jyeshe: For the first 99 txs maybe (as there was problem once with one txi around 80) and I wouldn't say for the last 99 because they change but assert the same tx hash for the txs between 500_000 and 500_100.

jyeshe (commented):
Nice chore. Only two comments that might help us across multiple resyncs. 
But it can be automated with a separate script if you prefer.

test/integration/ae_mdw_web/controllers/tx_controller_test.exs
sborrazas: These are additional changes that won't undo the changes introduced in this PR. Perhaps we can create a separate issue to add what you're asking for?

jyeshe: Sure #433

jyeshe (approved):

PR #428, opened 2021-12-30
perf: create ContractCallMutation

sborrazas:
#426  needs to be merged first

### DIFF
--- a/lib/ae_mdw/contract.ex
+++ b/lib/ae_mdw/contract.ex
@@ -38,6 +38,7 @@ defmodule AeMdw.Contract do
   # fcode or aevm info
   @type ct_info :: {:fcode, map(), list(), any()} | list()
   @type function_hash :: <<_::32>>
+  @type fun_arg_res() :: map()
 
   @typep tx :: tuple()
   @typep block_hash :: <<_::256>>
@@ -248,7 +249,7 @@ defmodule AeMdw.Contract do
     |> call_rec_from_id(contract_pk, block_hash)
   end
 
-  @spec call_tx_info(tx(), DBN.pubkey(), block_hash(), fun()) :: {map(), call()}
+  @spec call_tx_info(tx(), DBN.pubkey(), block_hash(), fun()) :: {fun_arg_res(), call()}
   def call_tx_info(tx_rec, contract_pk, block_hash, format_fn) do
     {:ok, ct_info} = get_info(contract_pk)
     call_id = :aect_call_tx.call_id(tx_rec)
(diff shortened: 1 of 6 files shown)


thepiwo (approved):

lib/ae_mdw/contract.ex
```diff
@@ -310,6 +314,24 @@ defmodule AeMdw.Contract do
     Enum.group_by(get_events(micro_block), fn {_event_name, %{tx_hash: tx_hash}} -> tx_hash end)
   end
 
+  @spec get_txi(pubkey()) :: integer()
+  def get_txi(contract_pk) do
+    case :ets.lookup(:ct_create_sync_cache, contract_pk) do
```
jyeshe, line 319: As this `get_txi` function is related to sync, it would be better to keep in previous module (`AeMdw.Db.Sync.Contract`)

lib/ae_mdw/db/sync/transaction.ex
```diff
@@ -264,6 +254,29 @@ defmodule AeMdw.Db.Sync.Transaction do
...
+       ) do
+    contract_pk = :aect_call_tx.contract_pubkey(tx)
+    create_txi = Contract.get_txi(contract_pk)
+    events = Map.get(mb_events, tx_hash, [])
+
+    {fun_arg_res, call_rec} =
+      Contract.call_tx_info(tx, contract_pk, block_hash, &Contract.to_map/1)
+
+    [
+      ContractEventsMutation.new(contract_pk, events, txi),
```
jyeshe, line 275: Same as https://github.com/aeternity/ae_mdw/pull/426#discussion_r777557249

jyeshe (changes requested):

jyeshe:
@sborrazas I have removed the design request. It would be missing now only the `get_txi` and to provide some performance indicator please.

lib/ae_mdw/contract.ex
sborrazas: I think overall all of the resource modules that are under the `AeMdw.Db.Sync.*` should be removed, and instead moved to the regular modules `AeMdw.Db.*` (or maybe even `AeMdw.*`). Right now both `Db.Sync.*` and `Db.*` contain similar functions for all resources, and you never know where to place them. 

What do you think? Shouldn't they all belong to a single `Contract/Tx/Oracle/etc` module? If not, how do you know where to place them?

jyeshe: I think the current division can be improved specially regarding the AEX9 but for scope of this PR we shouldn't change it. Ideally making it more clear what is specific for Sync and for Endpoints (besides separating what is specific for AEX9).

jyeshe (commented):
Comment on `get_txi` and some performance indicator please (or rename/consider as a `refactor`).

sborrazas:
@jyeshe done

jyeshe (approved):

PR #429, opened 2021-12-30
refactor: extract channel_create_tx syncing to Sync.Transaction

sborrazas:
Requires #428 to be merged

### DIFF
--- a/lib/ae_mdw/db/sync/transaction.ex
+++ b/lib/ae_mdw/db/sync/transaction.ex
@@ -276,6 +276,21 @@ defmodule AeMdw.Db.Sync.Transaction do
     ]
   end
 
+  defp tx_mutations(
+         :channel_create_tx,
+         _tx,
+         signed_tx,
+         txi,
+         tx_hash,
+         _block_index,
+         _block_hash,
+         _mb_events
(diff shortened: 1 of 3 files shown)


thepiwo (approved):

lib/ae_mdw/db/sync/transaction.ex
```diff
@@ -277,6 +277,21 @@ defmodule AeMdw.Db.Sync.Transaction do
     ]
   end
 
+  defp tx_mutations(
+         :channel_create_tx,
+         _tx,
+         signed_tx,
```
jyeshe, line 283: This param is only used by `:channel_create_tx`(and no other tx will use it). Could we move it outside of `tx_mutations` too? (following same reasoning as https://github.com/aeternity/ae_mdw/pull/426#discussion_r777557249)

jyeshe (changes requested):

jyeshe (approved):
The `signed_tx` param is only used by `:channel_create_tx` (and no other tx will use it). 
Since this is crossing 3 PRs and it's a design issue that is more subject, from my side I leave it optional request for now or later.

PR #430, opened 2021-12-30
perf: extract OracleRegister transaction into mutation

sborrazas:
Needs #429 to be merged first

PR #431, opened 2021-12-31
perf: extract name claim mutation

sborrazas:
Requires #430 to be merged first

### DIFF
--- a/lib/ae_mdw/db/mutations/name_claim_mutation.ex
+++ b/lib/ae_mdw/db/mutations/name_claim_mutation.ex
@@ -1,0 +1,181 @@
+defmodule AeMdw.Db.NameClaimMutation do
+  @moduledoc """
+  Processes name_claim_tx.
+  """
+
+  alias AeMdw.Blocks
+  alias AeMdw.Db.Format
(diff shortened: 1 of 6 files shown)


thepiwo (approved):

jyeshe (commented):
I intend to hold the approval of this for a while until reproducing the name expiration issue.

thepiwo:
@jyeshe @sborrazas can we merge those set of changes soon, or are there remaining blockers?

jyeshe:
> @jyeshe @sborrazas can we merge those set of changes soon, or are there remaining blockers?

The `:mnesia.sync_transaction` has resulted in consistent auction and active name expirations for 3 out of 3 syncs. The sample is not big but as there was an intermittent expiration issue that happened 2 out 3 syncs I think there is a base for other improvements. 

I imagine now it depends only if there were syncs and endpoints validations with the branch that includes all these refactorings. My vps is currently allocated on sync after #452 bugfix. However if @sborrazas have had one full sync and validated the `/names` and `/oracles` endpoints we could merge it and make more tests with your vps @thepiwo .

jyeshe (approved):

PR #438, opened 2022-01-04
feat: aex9 presence on calls other than transfer

jyeshe:


.dialyzer_ignore.exs
```diff
@@ -5,5 +5,7 @@
   # Following warnings are due to bad implementation of riverside.
   {"lib/phoenix/router.ex", :pattern_match, 402},
   {"lib/riverside.ex", :pattern_match, 876},
-  {"lib/ae_mdw_web/websocket/socket_handler.ex", :overlapping_contract, 2}
+  {"lib/ae_mdw_web/websocket/socket_handler.ex", :overlapping_contract, 2},
+  {"lib/ae_mdw/db/mutations/contract_call_mutation.ex", :pattern_match, 58},
+  {"lib/ae_mdw/db/mutations/contract_call_mutation.ex", :pattern_match, 59}
```
sborrazas, line 10: What specific errors do you get here? I'd like to avoid ignoring type errors

lib/ae_mdw/node.ex
```diff
@@ -37,7 +37,8 @@ defmodule AeMdw.Node do
           | :channel_client_reconnect_tx
           | :paying_for_tx
 
-  @type signed_tx() :: term()
+  @type signed_tx :: term()
+  @type pubkey :: <<_::256>>
```
sborrazas, line 41: This is found on `AeMdw.Node.Db`, perhaps we should use just one of them?

.dialyzer_ignore.exs
jyeshe: dialyzer is not following the short circuit condition on:
`Contract.is_successful_call?(fun_arg_res) and not Contract.is_non_stateful_aex9_function?(fun_arg_res.function)`
it complains because `fun_arg_res` is of the type `fun_arg_res_or_error()` (when a contract fails it's an error tuple) but fun_arg_res.function is a safe access because of `Contract.is_successful_call?`.

I've discarded the other way doing the parttern mathing of the whole fun_args_res_error that because it looses orthogonality. The function would become:
`Contract.is_non_stateful_aex9_successful_call?`

jyeshe: would you have another idea?

lib/ae_mdw/db/mutations/contract_call_mutation.ex
```diff
@@ -30,13 +43,45 @@ defmodule AeMdw.Db.ContractCallMutation do
...
+    if Contract.is_aex9?(contract_pk) and Contract.is_successful_call?(fun_arg_res) and
+         not Contract.is_non_stateful_aex9_function?(fun_arg_res.function) do
+      update_aex9_presence(
+        contract_pk,
+        caller_pk,
+        txi,
+        fun_arg_res.function,
+        fun_arg_res.arguments
+      )
+    end
```
sborrazas, line 65: Perhaps something like:
```elixir
with true <- Contract.is_aex9(contract_pk),
  {:ok, fname, args} <- Contract.extract_non_stateful_function(fun_arg_res) do
  
  update_aex9_presence(...)
end
```

Dialyzer is always going to complain about these cases (and any decent statically typed language will do too), so I would try to avoid getting into a point where we would need to add ignores for every one of them

jyeshe: I would avoid saying any, always, never, none in our profession (: it's not a NP-hard problem.

Good naming solution, thanks.

jyeshe: Please note it won't cover semantically these cases but we need to move on so I am removing the dialyzer_ignore. 
```
def extract_non_stateful_aex9_function(%{function: _fname, result: %{error: _error}}), do: :not_found
def extract_non_stateful_aex9_function(%{function: _fname, result: %{abort: _error}}), do: :not_found
```

sborrazas: > I would avoid saying any, always, never, none in our profession (: it's not a NP-hard problem.
> 
> Good naming solution, thanks.

"always", "never", "none", "all" are _**simplifier**_ words. "sometimes", "if", "when", "some" are _**complecting**_ words. Our job is to find the simplest solution (the hardest thing to do), and avoid circumventing the problem so that others have to pay for it in the future

jyeshe: "any decent statically typed language will do too" is not a strong statement? ok :+1: 
next problem with any compiler I will call you (:

sborrazas (approved):

jyeshe:
@thepiwo presence kept after a burn or swap f6b4457

lib/ae_mdw/db/contract.ex
```diff
@@ -62,6 +53,22 @@ defmodule AeMdw.Db.Contract do
     end
   end
 
+  @spec aex9_delete_presence(pubkey(), pubkey()) :: :ok
+  def aex9_delete_presence(contract_pk, account_pk) do
```
thepiwo, line 57: why is delete presence still needed?

lib/ae_mdw/db/mutations/contract_call_mutation.ex
```diff
@@ -30,13 +43,47 @@ defmodule AeMdw.Db.ContractCallMutation do
...
+  defp update_aex9_presence(contract_pk, caller_pk, txi, method_name, method_args) do
+    account_pk =
+      if method_name in ["burn", "swap"] do
+        caller_pk
+      else
+        Contract.get_aex9_destination_address(method_name, method_args)
+      end
+
+    if account_pk do
+      DBContract.aex9_write_presence(contract_pk, txi, account_pk)
```
thepiwo, line 83: this writes just if a account is present or not right?

lib/ae_mdw/db/mutations/contract_call_mutation.ex
```diff
@@ -30,13 +43,47 @@ defmodule AeMdw.Db.ContractCallMutation do
...
+
+    :ok
+  end
+
+  #
+  # Private functions
+  #
+  defp update_aex9_presence(contract_pk, caller_pk, txi, method_name, method_args) do
+    account_pk =
+      if method_name in ["burn", "swap"] do
```
thepiwo, line 76: both require presence before being used, so can be ignored for further presence finding

lib/ae_mdw/db/mutations/contract_call_mutation.ex
```diff
@@ -30,13 +43,47 @@ defmodule AeMdw.Db.ContractCallMutation do
...
+        caller_pk
```
thepiwo, line 77: only swap receiver would be interesting, but not part of the standard

lib/ae_mdw/db/mutations/contract_call_mutation.ex
```diff
@@ -30,13 +43,47 @@ defmodule AeMdw.Db.ContractCallMutation do
...
         txi: txi,
         fun_arg_res: fun_arg_res,
         call_rec: call_rec
       }) do
     DBContract.call_write(create_txi, txi, fun_arg_res)
     DBContract.logs_write(create_txi, txi, call_rec)
+
+    with true <- Contract.is_aex9?(contract_pk),
+         {:ok, method_name, method_args} <-
+           Contract.extract_non_stateful_aex9_function(fun_arg_res) do
```
thepiwo, line 58: is there a reason to use functions over events?

lib/ae_mdw/db/contract.ex
jyeshe: It's used on a migration to fix the `create_txi= -1`. With this signature turned out to be used only in tests, so it was moved to a new section for test helpers to be used in other test cases (e.g one present in `437-async-task-index-cache`).

lib/ae_mdw/db/mutations/contract_call_mutation.ex
jyeshe: exactly

lib/ae_mdw/db/mutations/contract_call_mutation.ex
jyeshe: The mdw is tracking the presence in two ways: 
- on creation (`create_txi`) 
- and per change basis (`call_txi`)

This would be for the `call_txi`.

lib/ae_mdw/db/mutations/contract_call_mutation.ex
jyeshe: Could you explain why the `burn` should be ignored? Asking because they are in the same section of the standard.

lib/ae_mdw/db/mutations/contract_call_mutation.ex
jyeshe: The reason is to update aex9 presence based on balance after contract calls that may transfer funds and are out of the standard.

lib/ae_mdw/db/mutations/contract_call_mutation.ex
thepiwo: what gets "changed" for burn and swap?

lib/ae_mdw/db/mutations/contract_call_mutation.ex
thepiwo: both don't need to be tracked

lib/ae_mdw/db/mutations/contract_call_mutation.ex
thepiwo: good!

lib/ae_mdw/db/mutations/contract_call_mutation.ex
jyeshe: I understand your concern and you know better than me that the balance of the caller is burned partially or completely. 

The thing is that even though the presence is already mapped, the model is `{account_pk, create_or_call_txi, contract_pk}` and the txi of the last change (last `call_txi`) is shown in the [/aex9/balances/account/:account](https://mainnet.aeternity.art/mdw/aex9/balances/account/ak_rLoCtHE3NK9dKyCNonJFYWkEEfeAsDUWa887GsCKqV1rhSuT6) 

The block_hash, tx_hash, tx_type and height are also derived from the last `call_txi`.

jyeshe: One improvement that could be done in the model would be change it to:
`{account_pk, contract_pk, create_txi, call_txi}` because only the last `call_txi` matters. 

It's retrieved only by `aex9_search_contract` that is called by `account_balances_reply` that filters by:
```
        call_txi = List.last(txi_list)
        {amount, create_txi, call_txi, contract_pk}
```
PS: Before this enhancement, with the current model, it could check if already exists before writing a new but it might degrade because it needs to iterate throughout all call_txis until find one that matches the contract_pk.
Opening an enhancement issue for this ...

jyeshe: Please find it here #459

lib/ae_mdw/db/mutations/contract_call_mutation.ex
jyeshe: For the current model it would be needed for the same reason as:
https://github.com/aeternity/ae_mdw/pull/438#discussion_r783353246

Unless your request aims to omit this particular changes (after a burn or swap) from /aex9/balances/account/. 

Could you detail it please?

lib/ae_mdw/db/mutations/contract_call_mutation.ex
thepiwo: okay, looks good then :)

thepiwo (approved):

PR #440, opened 2022-01-05
test: fix oracles and tx_controller integration tests

sborrazas:
@jyeshe what do you think about moving these `sync/oracle_test.exs` tests to unit tests instead?


jyeshe (approved):
LGTM. @sborrazas for me it's fine. It will need a little extra setup but I think it's a good move since `AeMdw.Db.Oracle.expire_oracle` doesn't depend on chain state.

PR #444, opened 2022-01-06
refactor: extract oracle extand/response mutations

sborrazas:
Requires #431 to be merged first


thepiwo (approved):

jyeshe (approved):
Nice separation of prepare and write state.

PR #445, opened 2022-01-06
refactor: split ga_attach_tx mutation to use FieldMutation instead

sborrazas:


lib/ae_mdw/db/sync/transaction.ex
```diff
@@ -285,6 +285,25 @@ defmodule AeMdw.Db.Sync.Transaction do
...
+         tx_hash,
+         _block_index,
+         _block_hash,
+         _mb_events
+       ) do
+    contract_pk = :aega_attach_tx.contract_pubkey(tx)
+    :ets.insert(:ct_create_sync_cache, {contract_pk, txi})
+    AeMdw.Ets.inc(:stat_sync_cache, :contracts)
+
+    [
```
jyeshe, line 302: harmless list nesting but not needed (like `:channel_create_tx`)

sborrazas: From a functional programming perspective, functions are simpler returning the same type always. E.g.:
```elixir
@spec tx_mutations(...) :: [Mutation.t()]
# is simpler than
@spec tx_mutations(...) :: [Mutation.t()] | Mutation.t()
```

It would actually very hack-ish to do in any proper statically typed programming language

jyeshe: Please compare with the return of `:channel_create_tx`:
https://github.com/aeternity/ae_mdw/blob/e3d4ab7cdb1cb98390fc32d622bcb13ff3683604/lib/ae_mdw/db/sync/transaction.ex#L285

`origin_mutations` already returns a list of mutations

sborrazas: You're right, I understood your first comment differently. Fixed now, sorry!

jyeshe (approved):

thepiwo (approved):

Issue #450, opened 2022-01-10
`/totalstats` and `/stats` broken


marc0olo:
affected version: `v1.4.0`

## current response
```json
{"data":[{"height":542490,"sum_block_reward":0,"sum_dev_reward":0,"total_token_supply":243746023899258298727537613},{"height":542489,"sum_block_reward":0,"sum_dev_reward":0,"total_token_supply":243746023899258298727537613},{"height":542488,"sum_block_reward":0,"sum_dev_reward":0,"total_token_supply":243746023899258298727537613},{"height":542487,"sum_block_reward":0,"sum_dev_reward":0,"total_token_supply":243746023899258298727537613},{"height":542486,"sum_block_reward":0,"sum_dev_reward":0,"total_token_supply":243746023899258298727537613},{"height":542485,"sum_block_reward":0,"sum_dev_reward":0,"total_token_supply":243746023899258298727537613},{"height":542484,"sum_block_reward":0,"sum_dev_reward":0,"total_token_supply":243746023899258298727537613},{"height":542483,"sum_block_reward":0,"sum_dev_reward":0,"total_token_supply":243746023899258298727537613},{"height":542482,"sum_block_reward":0,"sum_dev_reward":0,"total_token_supply":243746023899258298727537613},{"height":542481,"sum_block_reward":0,"sum_dev_reward":0,"total_token_supply":243746023899258298727537613}],"next":"/totalstats?cursor=542480&limit=10"}
```

## problem:
- values are zero and total supply doesn't seem to change

----------

additionally I am wondering if there was a breaking change in `/stats` because the https://aenalytics.org webseite doesn't display the number of auctions, names, contracts & oracles anymore

marc0olo:
update: problem is also existent for `/stats`

jyeshe:
`mnesia_rocksdb` bug affects `/totalstats`:
```
iex(aeternity@localhost)1> :mnesia.table_info(Model.ActiveName, :size) 
0
iex(aeternity@localhost)2> :mnesia.dirty_first(Model.ActiveName)
"0.chain"
```

marc0olo:
seems to be fixed for `/totalstats` in `v1.5.0`

still missing data for `/stats`

marc0olo:
apparently we had a different interpretation of that endpoint and somehow we introduced breaking changes here lately.

[...]

PR #452, opened 2022-01-10
fix: execute block_rewards mutation before stats mutation

sborrazas:
refs #450


jyeshe (dismissed):

jyeshe:
@sborrazas here is a complementary fix (needed also for `all_cached?=false` condition):
https://github.com/aeternity/ae_mdw/blob/efa876446ac386f433efea553121dc75a8b9a905/lib/ae_mdw/db/int_transfer.ex#L93

jyeshe (approved):
Nice catch. I haven't validated if this along with #453 is sufficient but both are needed.

thepiwo (approved):

Issue #381, opened 2021-12-08
fetch historical oracle information


marc0olo:
at the moment I see no way to get historical oracle information. from explorer perspective it would be cool to be able to show all historical queries and responses of an oracle.

ideally the query and responses are returned in one object. the response can be empty as the oracle doesn't necessarily respond.

I am not sure if this is an overkill but ideally I would also be able to fetch information about the oracle and the corresponding query/response by `query_id` / `response_id`.

thepiwo:
@sborrazas @jyeshe can you add your feedback, how much work would that be?

marc0olo:
@sborrazas can we close this? I just saw your comment in the issue of the frontend. seems like it's already doable. do the responses mentioned there already include all the historical oracle queries and responses?

we need a bit of a logic in the frontend to connect the responses to the queries. but I think that's ok

sborrazas:
@marc0olo to include all oracle-related txs you need to use `type_group=oracle` query filter, like so: https://mainnet.aeternity.io/mdw/txs/forward?type_group=oracle&oracle_id=ok_g5vQK6beY3vsTJHH7KBusesyzq9WMdEYorF8VyvZURXTjLnxT

This will include both query and responses. I think we can close this issue then

marc0olo:
yeah, awesome! can be closed =)

marc0olo:
@sborrazas it seems like we are unable to correlate the respond tx and the query tx with the grouped request. the `query_id` seems to be missing for the `OracleQueryTx` object :-(

`{"fee":20000,"nonce":3,"oracle_id":"ok_g5vQK6beY3vsTJHH7KBusesyzq9WMdEYorF8VyvZURXTjLnxT","query":"YmluYXJ5","query_fee":20000,"query_ttl":{"type":"delta","value":1000},"response_ttl":{"type":"delta","value":1000},"sender_id":"ak_g5vQK6beY3vsTJHH7KBusesyzq9WMdEYorF8VyvZURXTjLnxT","type":"OracleQueryTx","version":1}`

thepiwo:
@marc0olo @sborrazas I think this can be closed now, right?

marc0olo:
it will pop up again if there is still an issue. would create a new issue then or re-open if necessary. when @sborrazas thinks it can be closed I am fine. no time to check now 😅

PR #455, opened 2022-01-11
feat: add encoded query_id on query txs

sborrazas:
This is because core serialization doesn't return the query_id by
default: https://github.com/aeternity/aeternity/blob/master/apps/aeoracle/src/aeo_query_tx.erl#L254

ref #381
ref #60


marc0olo:
wow, that was quick! thanks =)

thepiwo (approved):

jyeshe (approved):
query_id returned on `http://localhost:4000/txs/forward?type=oracle_query`

Issue #437, opened 2022-01-04
update aex9 presence after other calls besides transfer


jyeshe:
Preferably exclude the non stateful method calls from the `:update_aex9_presence`.

jyeshe:
Calls that are explicit regarding the destination of the transfer funds will be used for mapping the account presence in an AEX9.
For other calls, the AEX9 balance will be used to continue tracking the up to date account presence.

PR #456, opened 2022-01-11
perf: async task type and args as a key

jyeshe:
## What

Indexes the async tasks by task type and `args` (these two attributes define the work to be peformed).

## Why
Currently, for lower volume of `UpdateAex9Presence` and `DeriveAex9Presence`, the task `args` (e.g `contract_pk`) is used to check if an async task already exists to avoid duplicate processing. However, the task `args` is not a key/index for async tasks.

With #437 and DEX, the volume of async tasks will increase justifying another index (in memory only).

PS: The current async task index is used for ordering to avoid task processing starvation.

## Validation steps

`elixir --sname aeternity@localhost -S mix test.integration test/integration/ae_mdw/sync/async_tasks/producer_consumer_test.exs`

### DIFF
--- a/lib/ae_mdw/db/contract.ex
+++ b/lib/ae_mdw/db/contract.ex
@@ -76,14 +76,33 @@ defmodule AeMdw.Db.Contract do
     :ok
   end
 
-  @spec aex9_presence_exists?(pubkey(), pubkey(), integer() | nil) :: boolean()
-  def aex9_presence_exists?(contract_pk, account_pk, txi \\ nil) do
-    txi_search_prev = (txi && txi + 1) || nil
+  @spec aex9_presence_exists?(pubkey(), pubkey(), integer()) :: boolean()
+  def aex9_presence_exists?(contract_pk, account_pk, txi) do
+    txi_search_prev = txi + 1
 
     case :mnesia.prev(Model.Aex9AccountPresence, {account_pk, txi_search_prev, contract_pk}) do
       {^account_pk, _txi, ^contract_pk} -> true
       _other_key -> false
     end
+  end
+
+  @spec aex9_presence_exists?(pubkey(), pubkey()) :: boolean()
+  def aex9_presence_exists?(contract_pk, account_pk) do
+    0
+    |> Stream.unfold(fn
+      :found ->
+        nil
+
+      txi_search_next ->
+        case :mnesia.next(Model.Aex9AccountPresence, {account_pk, txi_search_next, contract_pk}) do
+          {^account_pk, _txi, ^contract_pk} -> {true, :found}
+          {^account_pk, next_txi, _contract_pk} -> {false, next_txi}
+          _not_found -> nil
+        end
+    end)
+    |> Enum.to_list()
+    |> List.last()
+    |> Kernel.||(false)
   end
 
   @spec call_write(integer(), integer(), Contract.fun_arg_res_or_error()) :: :ok
--- a/lib/ae_mdw/sync/async_tasks/store.ex
+++ b/lib/ae_mdw/sync/async_tasks/store.ex
@@ -78,7 +78,7 @@ defmodule AeMdw.Sync.AsyncTasks.Store do
   #
   # Private functions
   #
-  def cache_tasks_by_args() do
+  defp cache_tasks_by_args() do
     {:atomic, indexed_args_records} =
       :mnesia.transaction(fn ->
         args_spec =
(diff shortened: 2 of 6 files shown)


thepiwo (approved):
good as far as I understand

lib/ae_mdw/sync/async_tasks/store.ex
```diff
@@ -53,44 +60,39 @@ defmodule AeMdw.Sync.AsyncTasks.Store do
...
-      end
+  defp cache_tasks_by_args() do
+    {:atomic, indexed_args_records} =
+      :mnesia.transaction(fn ->
+        args_spec =
+          Ex2ms.fun do
+            Model.async_tasks(index: {_ts, task_type}, args: args) -> {{task_type, args}}
+          end
+
+        :mnesia.select(Model.AsyncTasks, args_spec)
```
sborrazas, line 89: Wouldn't it be more efficient here if we used `:mnesia.all_keys` instead, and then map every value to obtain the `{task_type, args}` tuples?

lib/ae_mdw/db/contract.ex
```diff
@@ -76,16 +76,35 @@ defmodule AeMdw.Db.Contract do
...
+        case :mnesia.next(Model.Aex9AccountPresence, {account_pk, txi_search_next, contract_pk}) do
+          {^account_pk, _txi, ^contract_pk} -> {true, :found}
+          {^account_pk, next_txi, _contract_pk} -> {false, next_txi}
+          _not_found -> nil
+        end
+    end)
+    |> Enum.to_list()
+    |> List.last()
+    |> Kernel.||(false)
+  end
```
sborrazas, line 106: I thought of a more brief version of it, taking advantage that streams are lazy, and not all values will be computed:
```elixir
0
|> Stream.unfold(fn
  txi_search_next ->
    case :mnesia.next(Model.Aex9AccountPresence, {account_pk, txi_search_next, contract_pk}) do
      {^account_pk, next_txi, contract_pk} -> {contract_pk, next_txi}
      _not_found -> nil
    end
end)
|> Enum.member?(contract_pk)
```

jyeshe: The `{^account_pk, _txi, ^contract_pk} -> {true, :found}` is helping to stop on first `contract_pk` match. Without this check, it iterates through all presences of the account.

The index gonna be refactored though to avoid these potential longer iterations #459

lib/ae_mdw/sync/async_tasks/store.ex
jyeshe: profiled now ... `:mnesia.select` too only 100ms for 10k items (runs once on app start).

lib/ae_mdw/db/contract.ex
sborrazas: > The `{^account_pk, _txi, ^contract_pk} -> {true, :found}` is helping to stop on first `contract_pk` match. Without this check, it iterates through all presences of the account.
> 
> The index gonna be refactored though to avoid these potential longer iterations #459

That's not how streams work, the first match with `member?` will stop the stream from computing the rest

sborrazas (approved):

lib/ae_mdw/db/contract.ex
jyeshe: Please notice that the size of the list is determined by `Stream.unfold` (it doesn't use a condition outside of the inlined function).

sborrazas: The benefit and elegance of streams, is that it could be infinite and not necessarily compute in its entirety. Manually handling the end of the stream adds complexity (e.g. having to handle end values like `:found` and adding a lot more code into it), but I guess it's a matter of taste, sorry for bring it up!

jyeshe: No problem, I like discussing readability too. Only in this case it's a behaviour of `Stream.unfold`: "The enumeration finishes if it returns nil."

Maybe a case for a new Enum or Stream function to pick the last based not on a list but on unfolded result like a `reduce` but receiving a seed like `unfold`.

sborrazas: "The enumeration finishes if it returns `nil`" doesn't mean that it will continue executing until it returns `nil`, all streams are lazy and will only execute if the value is needed. Example:
```elixir
Stream.unfold(0, fn 
  5 -> nil
  n -> IO.inspect(n + 1); {n + 1, n + 1} 
end) |> Enum.member?(3)
# prints 1, 2, 3
# returns true
```

As you can see in this example, the values 4 and 5 weren't computed at all

PR #461, opened 2022-01-13
feat: /names/owned_by for inactive names

### LINKED ISSUE #157: Enpoint returning inactive names for the account
We've need such thing to use in the wallet, to make user experience more user friendly.

jyeshe:
## What

Add/handle `active` parameter on `/name/owned_by` to return active or inactive names (last ownership of expired or revoked names).

## Why

Resolves #157 

## Validation steps

`elixir --sname aeternity@localhost -S mix test.integration test/integration/ae_mdw_web/controllers/name_controller_test.exs:544`
`elixir --sname aeternity@localhost -S mix test.integration test/integration/ae_mdw_web/controllers/name_controller_test.exs`

### DIFF
--- a/lib/ae_mdw/db/name.ex
+++ b/lib/ae_mdw/db/name.ex
@@ -30,10 +30,16 @@ defmodule AeMdw.Db.Name do
     |> map_ok!(&Validate.id!/1)
   end
 
-  def owned_by(owner_pk) do
+  def owned_by(owner_pk, _active? = true) do
     %{
-      actives: collect_vals(Model.ActiveNameOwner, owner_pk),
+      names: collect_vals(Model.ActiveNameOwner, owner_pk),
       top_bids: collect_vals(Model.AuctionOwner, owner_pk)
+    }
+  end
+
+  def owned_by(owner_pk, _active? = false) do
+    %{
+      names: collect_vals(Model.InactiveNameOwner, owner_pk)
     }
   end
 
@@ -73,6 +79,7 @@ defmodule AeMdw.Db.Name do
     if Model.name(m_name, :expire) == height do
       owner = Model.name(m_name, :owner)
       cache_through_write(Model.InactiveName, m_name)
+      cache_through_write(Model.InactiveOwnerName, {owner, plain_name})
       cache_through_write(Model.InactiveNameExpiration, m_exp)
       cache_through_delete(Model.ActiveName, plain_name)
       cache_through_delete(Model.ActiveNameOwner, {owner, plain_name})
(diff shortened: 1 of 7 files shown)


lib/ae_mdw_web/controllers/name_controller.ex
```diff
@@ -97,7 +95,8 @@ defmodule AeMdwWeb.NameController do
   def owned_by(conn, %{"id" => owner} = params),
     do:
       handle_input(conn, fn ->
-        owned_by_reply(conn, Validate.id!(owner, [:account_pubkey]), expand?(params))
+        active? = params |> Map.get("active", "true") |> String.to_existing_atom()
```
sborrazas, line 98: This will return a 500 for any value that does not exist as an `atom`. We should probably use `Map.get("active", "true") == "true"` instead

lib/ae_mdw/db/name.ex
```diff
@@ -30,13 +30,19 @@ defmodule AeMdw.Db.Name do
...
     %{
-      actives: collect_vals(Model.ActiveNameOwner, owner_pk),
+      names: collect_vals(Model.ActiveNameOwner, owner_pk),
       top_bids: collect_vals(Model.AuctionOwner, owner_pk)
     }
   end
 
+  def owned_by(owner_pk, _active? = false) do
+    %{
+      names: collect_vals(Model.InactiveNameOwner, owner_pk)
```
sborrazas, line 42: I don't think we should ever return unlimited lists of results in a single page, this is easily exploitable. But we were already doing it for active names so I'll leave it up to you. Perhaps we could think of a new endpoint?

sborrazas (changes requested):

lib/ae_mdw_web/controllers/name_controller.ex
jyeshe: done

priv/migrations/20220113112001_inactive_name_owner.ex
```diff
@@ -0,0 +1,38 @@
...
+  alias AeMdw.Db.Model
+  alias AeMdw.Log
+
+  require Ex2ms
+  require Model
+
+  @doc """
+  Writes {account_pk, create_txi, contract_pk} aex9 presence and deletes old ones with txi = -1.
+  """
+  @spec run(boolean()) :: {:ok, {pos_integer(), pos_integer()}}
```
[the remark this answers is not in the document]
jyeshe: right, done.

sborrazas (approved):

thepiwo (approved):
cool!

PR #463, opened 2022-01-14
fix: inactive name owner table name for sync

jyeshe:


thepiwo (approved):

sborrazas (approved):

lib/ae_mdw/db/name.ex
```diff
@@ -79,7 +79,7 @@ defmodule AeMdw.Db.Name do
     if Model.name(m_name, :expire) == height do
       owner = Model.name(m_name, :owner)
       cache_through_write(Model.InactiveName, m_name)
-      cache_through_write(Model.InactiveOwnerName, {owner, plain_name})
+      cache_through_write(Model.InactiveNameOwner, {owner, plain_name})
```
sborrazas, line 82: This fails locally for me as well, because the 2nd parameter of a `:mnesia.write/3` has to be the full record, not just the key of the record

sborrazas (changes requested):

lib/ae_mdw/db/name.ex
jyeshe: probably should leave more time syncing... thank you for finding this!

sborrazas (approved):

PR #464, opened 2022-01-14
perf: drop txs not used for events

jyeshe:
## What

Drop micro-block txs after contract calls for less dry-running load. The events of these txs (when produced) are not read so avoids dry-running these txs.

The ones before the contract calls are preserved/not dropped from dry run because they change the chain state trees that are an input for contract calls dry run.

## Why

Profiling the reading from generation 0 to 540,000:

    loading blocks: 7h40m
    loading blocks + events: 12h20m
    loading blocks + events dropping unused txs: 11h20m

Loading blocks here means only reading them from the Node and events to dry run the txs of these blocks.

refs #346

### DIFF
--- a/lib/ae_mdw/contract.ex
+++ b/lib/ae_mdw/contract.ex
@@ -437,8 +437,9 @@ defmodule AeMdw.Contract do
 
   defp get_events(micro_block) when elem(micro_block, 0) == :mic_block do
     txs = :aec_blocks.txs(micro_block)
+    txs_taken = txs_until_last_contract_tx(txs)
 
-    if has_contract_tx?(txs) do
+    if txs_taken != [] do
       header = :aec_blocks.to_header(micro_block)
       {:ok, hash} = :aec_headers.hash_header(header)
       consensus = :aec_headers.consensus_module(header)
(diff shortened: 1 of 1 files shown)


lib/ae_mdw/contract.ex
```diff
@@ -451,18 +452,32 @@ defmodule AeMdw.Contract do
...
   end
 
-  defp has_contract_tx?(mb_txs) do
-    Enum.any?(mb_txs, fn signed_tx ->
-      {mod, _tx} = :aetx.specialize_callback(:aetx_sign.tx(signed_tx))
-      mod.type() in [:contract_create_tx, :contract_call_tx]
-    end)
+  defp txs_until_last_contract_tx(mb_txs) do
+    {tx_pos, last_ct_tx_pos} =
+      Enum.reduce(mb_txs, {1, -1}, fn signed_tx, {tx_pos, last_index} ->
```
jyeshe, line 466: traverse the whole list of txs but compensates and it was already done most of the cases with `Enum.any?`

thepiwo (approved):

sborrazas (approved):

Issue #425, opened 2021-12-25
index aex9 contracts with chain.clone and chain.create


jyeshe:

thepiwo:
should definitely be done for consistency

jyeshe:
Gonna add aex9 info and we will have the feature done.
```
if aex9_meta_info do
      DBContract.aex9_creation_write(aex9_meta_info, contract_pk, owner_pk, txi)
end
```

jyeshe:
> if aex9_meta_info do
>       DBContract.aex9_creation_write(aex9_meta_info, contract_pk, owner_pk, txi)
> end

Not enough as there is no `Chain.create` nor `Chain.clone` for `ct_skLwjzYhBH4J9ETuayxiVRaoYJmgdMfZJcrc581fpjEm84ykm` from #507

PR #474, opened 2022-01-20
fix: index the origin of contracts created via Chain.clone/create

sborrazas:
refs #425

### DIFF
--- a/lib/ae_mdw/contracts.ex
+++ b/lib/ae_mdw/contracts.ex
@@ -411,8 +411,8 @@ defmodule AeMdw.Contracts do
     pk = Validate.id!(contract_id)
 
     case Origin.tx_index({:contract, pk}) do
-      nil -> raise ErrInput.Id, value: contract_id
-      txi -> txi
+      {:ok, txi} -> txi
+      :not_found -> raise ErrInput.Id, value: contract_id
     end
   end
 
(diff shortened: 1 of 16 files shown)


thepiwo (approved):

jyeshe (commented):
Code looks good and all tests passed except this one that looks like could be deleted:
1) test logs when contract log doesn't have a creation txi, contract_id shouldn't be nil (Integration.AeMdwWeb.ContractControllerTest)
     test/integration/ae_mdw_web/controllers/contract_controller_test.exs:180
     ** (RuntimeError) expected response with status 200, got: 400, with body:
     "{\"error\":\"invalid id: ct_eJhrbPPS4V97VLKEVbSCJFpdA4uyXiZujQyLqMFoYV88TzDe6\"}"

In this PR or following one after the release we could delete the Model.Origin table.

sborrazas:
@jyeshe that test doesn't fail for me, and it looks like what it's testing is correct (special hard-coded contracts have negative values for the creation txi). But perhaps we should probably just store the negative value, and not display it to the user, display "nil" instead. Thoughts?

jyeshe:
> @jyeshe that test doesn't fail for me, and it looks like what it's testing is correct (special hard-coded contracts have negative values for the creation txi). But perhaps we should probably just store the negative value, and not display it to the user, display "nil" instead. Thoughts?

The tests only had run with the `1.5.0-07cf3e4` database skipping two migrations. With the #481 and rerunning the migrations the test passed. Better to keep the test as you're saying because of the contracts from the fork.

jyeshe:
Gonna do a final validation after the #481 merge (or similar fix).

jyeshe:
```
test events_mutations/4 creates name update mutation (Integration.AeMdw.Db.Sync.ContractTest)
     test/integration/ae_mdw/db/sync/contract_test.exs:45
test events_mutations/4 creates name transfer mutation (Integration.AeMdw.Db.Sync.ContractTest)
     test/integration/ae_mdw/db/sync/contract_test.exs:16
```
After the merge of #481 and changing the `Sync.Contract.events_mutations` calls on the tests above, I think we are good to go and have all contracts indexed :pray:

jyeshe (approved):
tests passed

PR #475, opened 2022-01-21
feat: rocksdb without mnesia

jyeshe:
## What

This is a foundational PoC to overcome remaining consistency problems and improve [write performance](https://github.com/aeternity/ae_mdw/issues/346#issuecomment-1017769284) closely to 60%. With the help of existing `Mnesia` module and by generalizing it to `Database`, the Mdw code can be easily changed to use RocksDB directly without `:mnesia` and `:mnesia_rocksdb`.

It enables actual database transactions managed by RocksDB via a single RocksDB database and
multiple column families allowing to write in multiple tables on the same transaction. 

The read during transactions might be also dirty to get states of one chain transaction after another before the microblock commit.

```
AeMdw.Db.RocksDbCF.put(Model.Block, Model.block(index: {123, -1}, tx_index: 234))
:ok
AeMdw.Db.RocksDbCF.fetch(Model.Block, {123, -1})                                           
:not_found
AeMdw.Db.RocksDbCF.dirty_fetch(Model.Block, {123, -1})     
{:ok, {:block, {123, -1}, 234, ""}}
AeMdw.Db.RocksDb.commit()                             
:ok
AeMdw.Db.RocksDbCF.fetch(Model.Block, {123, -1})      
{:ok, {:block, {123, -1}, 234, ""}}
```

## Why

Step to #529 

## Validation steps

Unit tests that can be run anytime (and validated by CI):
`elixir --sname aeternity@localhost -S mix test test/ae_mdw/db/rocksdbcf_test.exs`

After syncing for 4 hours with this branch, these tests compare the result with the response of mainnet.aeternity.io:
```
export INTEGRATION_TEST=1 
elixir --sname aeternity@localhost -S mix test.integration test/integration/ae_mdw_web/controllers/tx_controller_test.exs:35
elixir --sname aeternity@localhost -S mix test.integration test/integration/ae_mdw_web/controllers/tx_controller_test.exs:87
```

## Additional info

This PR applies the persistence with RocksDB initially to the Model.Block and Model.Tx tables.There is no performance gain noticed as we have analysed that loading the blocks and dry-running the events take most of the share of the processing. However there is the benefit of having database transaction managed by RocksDB, with writes to multiple column families in the same transaction.

### DIFF
--- a/lib/ae_mdw/database.ex
+++ b/lib/ae_mdw/database.ex
@@ -16,8 +16,6 @@ defmodule AeMdw.Database do
   alias AeMdw.Db.RocksDb
   alias AeMdw.Db.RocksDbCF
 
-  require Model
-
   @type table() :: atom()
   @type record() :: tuple()
   @type key() :: term()
@@ -185,13 +183,13 @@ defmodule AeMdw.Database do
     match?({:ok, _record}, fetch(tab, key))
   end
 
-  @spec delete(table(), key()) :: :ok | {:error, any()}
+  @spec delete(table(), key()) :: :ok
   def delete(tab, key) when use_rocksdb?(tab) do
     :ok = RocksDbCF.delete(tab, key)
   end
 
   def delete(table, key) do
-    :ok = :mnesia.delete(table, key, :write)
+    :mnesia.delete(table, key, :write)
   end
 
   @spec read(table(), key()) :: [record()]
@@ -213,11 +211,11 @@ defmodule AeMdw.Database do
 
   @spec write(table(), record()) :: :ok
   def write(tab, record) when use_rocksdb?(tab) do
-    :ok = RocksDbCF.put(tab, record)
+    RocksDbCF.put(tab, record)
   end
 
   def write(table, record) do
-    :ok = :mnesia.write(table, record, :write)
+    :mnesia.write(table, record, :write)
   end
 
   @spec commit() :: :ok
--- a/lib/ae_mdw/database.ex
+++ b/lib/ae_mdw/database.ex
@@ -212,12 +212,8 @@ defmodule AeMdw.Database do
   end
 
   @spec write(table(), record()) :: :ok
-  def write(Model.Block, record) do
-    :ok = RocksDbCF.put(Model.Block, record)
-  end
-
-  def write(Model.Tx, record) do
-    :ok = RocksDbCF.put(Model.Tx, record)
+  def write(tab, record) when use_rocksdb?(tab) do
+    :ok = RocksDbCF.put(tab, record)
   end
 
   def write(table, record) do
--- a/lib/ae_mdw/database.ex
+++ b/lib/ae_mdw/database.ex
@@ -212,8 +212,12 @@ defmodule AeMdw.Database do
   end
 
   @spec write(table(), record()) :: :ok
-  def write(tab, record) when use_rocksdb?(tab) do
-    :ok = RocksDbCF.put(tab, record)
+  def write(Model.Block, record) do
+    :ok = RocksDbCF.put(Model.Block, record)
+  end
+
+  def write(Model.Tx, record) do
+    :ok = RocksDbCF.put(Model.Tx, record)
   end
 
   def write(table, record) do
--- a/lib/ae_mdw/database.ex
+++ b/lib/ae_mdw/database.ex
@@ -88,7 +88,7 @@ defmodule AeMdw.Database do
 
   @spec last_key(table()) :: {:ok, key()} | :none
   def last_key(tab) when use_rocksdb?(tab) do
-    case RocksDbCF.last_key(tab, false) do
+    case RocksDbCF.last_key(tab) do
       {:ok, last_key} -> {:ok, last_key}
       :not_found -> :none
     end
@@ -187,11 +187,11 @@ defmodule AeMdw.Database do
 
   @spec delete(table(), key()) :: :ok | {:error, any()}
   def delete(tab, key) when use_rocksdb?(tab) do
-    RocksDbCF.delete(tab, key)
+    :ok = RocksDbCF.delete(tab, key)
   end
 
   def delete(table, key) do
-    :mnesia.delete(table, key, :write)
+    :ok = :mnesia.delete(table, key, :write)
   end
 
   @spec read(table(), key()) :: [record()]
@@ -213,20 +213,20 @@ defmodule AeMdw.Database do
 
   @spec write(table(), record()) :: :ok
   def write(Model.Block, record) do
-    RocksDbCF.put(Model.Block, record)
+    :ok = RocksDbCF.put(Model.Block, record)
   end
 
   def write(Model.Tx, record) do
-    RocksDbCF.put(Model.Tx, record)
+    :ok = RocksDbCF.put(Model.Tx, record)
   end
 
   def write(table, record) do
-    :mnesia.write(table, record, :write)
+    :ok = :mnesia.write(table, record, :write)
   end
 
   @spec commit() :: :ok
   def commit do
-    RocksDb.commit()
+    :ok = RocksDb.commit()
   end
 
   @spec transaction([Mutation.t()]) :: :ok
--- a/lib/ae_mdw/database.ex
+++ b/lib/ae_mdw/database.ex
@@ -11,7 +11,11 @@ defmodule AeMdw.Database do
   returned instead.
   """
 
+  alias AeMdw.Db.Model
   alias AeMdw.Db.Mutation
+  alias AeMdw.Db.RocksDbCF
+
+  require Model
 
   @type table() :: atom()
   @type record() :: tuple()
@@ -21,6 +25,12 @@ defmodule AeMdw.Database do
   @type limit() :: pos_integer()
 
   @end_token :"$end_of_table"
+
+  defmacro use_rocksdb?(tab) do
+    quote do
+      unquote(tab) == Model.Block or unquote(tab) == Model.Tx
+    end
+  end
 
   @spec dirty_all_keys(table()) :: [key()]
   def dirty_all_keys(table) do
@@ -76,6 +86,13 @@ defmodule AeMdw.Database do
   end
 
   @spec last_key(table()) :: {:ok, key()} | :none
+  def last_key(tab) when use_rocksdb?(tab) do
+    case RocksDbCF.last_key(tab, false) do
+      {:ok, last_key} -> {:ok, last_key}
+      :not_found -> :none
+    end
+  end
+
   def last_key(tab) do
     case :mnesia.dirty_last(tab) do
       @end_token -> :none
@@ -92,6 +109,13 @@ defmodule AeMdw.Database do
   end
 
   @spec first_key(table()) :: {:ok, key()} | :none
+  def first_key(tab) when use_rocksdb?(tab) do
+    case RocksDbCF.first_key(tab) do
+      {:ok, first_key} -> {:ok, first_key}
+      :not_found -> :none
+    end
+  end
+
   def first_key(tab) do
     case :mnesia.dirty_first(tab) do
       @end_token -> :none
@@ -100,6 +124,13 @@ defmodule AeMdw.Database do
   end
 
   @spec prev_key(table(), key()) :: {:ok, key()} | :none
+  def prev_key(tab, key) when use_rocksdb?(tab) do
+    case RocksDbCF.prev_key(tab, key) do
+      {:ok, prev_key} -> {:ok, prev_key}
+      :not_found -> :none
+    end
+  end
+
   def prev_key(tab, key) do
     case :mnesia.dirty_prev(tab, key) do
       @end_token -> :none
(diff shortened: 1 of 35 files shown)


lib/ae_mdw/db/rocksdb.ex
```diff
@@ -0,0 +1,258 @@
...
+  def iterator(table, read_options \\ []) do
+    {db_ref, cf_ref} = cf_refs(table)
+
+    :rocksdb.iterator(db_ref, cf_ref, read_options)
+  end
+
+  @doc """
+  Delete a key-value from a column family.
+  """
+  @spec delete(table(), binary()) :: boolean()
```
jyeshe, line 170: this will adjusted to be write through the transaction

lib/ae_mdw/db/rocksdb.ex
```diff
@@ -0,0 +1,258 @@
...
+  def delete(table, key) do
+    {db_ref, cf_ref} = cf_refs(table)
+
+    case :rocksdb.delete(db_ref, cf_ref, key, []) do
```
jyeshe, line 174: this will adjusted to be write through the transaction

jyeshe:
@sborrazas @thepiwo a suggestion is to leave the Mdw syncing from scratch until height 135000, which takes about 4 hours (with the Node already synced), and then run these commands:
```
export INTEGRATION_TEST=1 
elixir --sname aeternity@localhost -S mix test.integration test/integration/ae_mdw_web/controllers/tx_controller_test.exs:35
elixir --sname aeternity@localhost -S mix test.integration test/integration/ae_mdw_web/controllers/tx_controller_test.exs:87
```

jyeshe:
rebased and ready for final validation

lib/ae_mdw/db/sync/transaction.ex
```diff
@@ -174,6 +172,7 @@ defmodule AeMdw.Db.Sync.Transaction do
         if mbi > last_mbi do
           {mutations, acc} = micro_block_mutations(mblock, txi_acc)
           Database.transaction(mutations)
+          Database.commit()
```
jyeshe, line 175: In a following PR the commits will be done only through mutations.

lib/ae_mdw/db/util.ex
```diff
@@ -239,11 +260,11 @@ defmodule AeMdw.Db.Util do
 
   @spec gen_to_txi(Blocks.height()) :: Txs.txi()
   def gen_to_txi(gen) do
-    case Database.fetch(Model.Block, {gen, -1}) do
```
jyeshe, line 242: Changed not needed anymore (same result). This will be reverted in a following PR.

lib/ae_mdw/db/util.ex
```diff
@@ -253,11 +274,11 @@ defmodule AeMdw.Db.Util do
 
   @spec txi_to_gen(Txs.txi()) :: Blocks.height()
   def txi_to_gen(txi) do
-    case Database.fetch(Model.Tx, txi) do
```
jyeshe, line 256: same here

thepiwo (approved):
looks good to me, but can't judge the code in detail

sborrazas:
As per our discussion on today's meeting, these are the changes that we agreed on:

- Reverting `DatabaseWriteMutation` change
- Not exposing the `index()` concept outside Database
- `defp encode_record_value(table, {record_name, key, nil}), do: ""`
- `defp decode_value(table, ""), do: {}`
- `@opaque transaction()`
- `@opaque iterator()`
- `RocksDbCF` should not use `:rocksdb`
- Clean `RocksDbCF` so that it doesn't know what a block/tx (or what ETS is). Move it to `Util`
- Every function in `Database` should validate the return value from `RocksDb` (`:ok = ..., etc`)

Additionally we discussed adding these in subsequent PRs:

- `DatabaseWriteMutation -> WriteMutation`
- Implement V3 and remove `:persistent_term`

V3:
```elixir
{:ok, transaction} = RocksDb.open()

Enum.each(mutations, &(Mutation.mutation(transaction, &1))

:ok = RocksDb.close(transaction)
```

lib/ae_mdw/db/util.ex
```diff
@@ -30,7 +31,7 @@ defmodule AeMdw.Db.Util do
     do: read_tx(txi) |> one!
 
   def read_block({_, _} = bi) do
-    case Database.fetch(~t[block], bi) do
+    case RocksDbCF.dirty_fetch(~t[block], bi) do
```
jyeshe, line 34: Direct use of `RocksDbCF` as read through an ongoing transaction is required (different from the way `Database.dirty_read/2` currently works for other tables). So this avoids same name for different behaviors.

jyeshe:
@sborrazas done

Please find the changes for each item after 7384743. 

The cache needed to stay in `RocksDbCF` for now because it needs to be a write through operation (moving outside would risk to have stale data and without the cache it's much slower). After the refactoring of `AeMdw.Db.Util`, we can add a restriction so that all operations regarding Model.Tx and Model.Block are made with it.

lib/ae_mdw/db/util.ex
```diff
@@ -11,31 +12,42 @@ defmodule AeMdw.Db.Util do
...
+    end
+  end
 
   def read_tx!(txi),
     do: read_tx(txi) |> one!
 
-  def read_block({_, _} = bi),
-    do: :mnesia.async_dirty(fn -> Database.read(~t[block], bi) end)
+  def read_block({_, _} = bi) do
+    case RocksDbCF.read_block(bi) do
```
jyeshe, line 34: Direct access to `RocksDbCF` as `Database.dirty_read` is a different operation (behaves differently because previously there was no real transaction). After refactoring the Database removing the Mnesia code, `Database.dirty_read` can have a uniform behavior of reading through the transaction (read uncommitted then if not read committed).

jyeshe:
The height for these [integration tests](https://github.com/aeternity/ae_mdw/pull/475#issuecomment-1040800726) was decreased to 44495 (after 1 million txs)

lib/ae_mdw/database.ex
```diff
@@ -11,7 +11,12 @@ defmodule AeMdw.Database do
   returned instead.
   """
 
+  alias AeMdw.Db.Model
   alias AeMdw.Db.Mutation
+  alias AeMdw.Db.RocksDb
+  alias AeMdw.Db.RocksDbCF
+
+  require Model
```
sborrazas, line 19: We don't need this `require` anymore

lib/ae_mdw/database.ex
```diff
@@ -127,31 +185,23 @@ defmodule AeMdw.Database do
...
 
-  @spec next_key(table(), direction(), cursor()) :: {:ok, key()} | :none
-  def next_key(tab, :forward, nil), do: first_key(tab)
-
-  def next_key(tab, :forward, cursor) do
-    case :mnesia.dirty_next(tab, cursor) do
-      @end_token -> :none
-      record -> {:ok, record}
-    end
+  @spec delete(table(), key()) :: :ok | {:error, any()}
```
sborrazas, line 188: Wrong typespec

lib/ae_mdw/database.ex
```diff
@@ -162,8 +212,17 @@ defmodule AeMdw.Database do
   end
 
   @spec write(table(), record()) :: :ok
+  def write(tab, record) when use_rocksdb?(tab) do
+    :ok = RocksDbCF.put(tab, record)
```
sborrazas, line 216: No need to pattern-match the result here

lib/ae_mdw/database.ex
```diff
@@ -162,8 +212,17 @@ defmodule AeMdw.Database do
...
+  end
+
   def write(table, record) do
-    :mnesia.write(table, record, :write)
+    :ok = :mnesia.write(table, record, :write)
```
sborrazas, line 220: No need to pattern-match the result here

lib/ae_mdw/db/rocksdb.ex
```diff
@@ -0,0 +1,264 @@
...
+  Puts changes into an optimistic transaction.
+
+  RocksDB raises an exception on conflict which is helpful to identify unwanted parallelism in the future when
+  it doesn't respect dependency.
+  """
+  @spec put(table(), binary(), binary()) :: :ok | :error
+  def put(table, key, value) do
+    with {_db_ref, cf_ref} <- cf_refs(table),
+         {:ok, t_ref} <- get_transaction() do
+      :ok = :rocksdb.transaction_put(t_ref, cf_ref, key, value)
```
sborrazas, line 93: No need to pattern-match the result here

lib/ae_mdw/db/rocksdb.ex
```diff
@@ -0,0 +1,264 @@
...
+  @doc """
+  Puts changes into an optimistic transaction.
+
+  RocksDB raises an exception on conflict which is helpful to identify unwanted parallelism in the future when
+  it doesn't respect dependency.
+  """
+  @spec put(table(), binary(), binary()) :: :ok | :error
+  def put(table, key, value) do
+    with {_db_ref, cf_ref} <- cf_refs(table),
+         {:ok, t_ref} <- get_transaction() do
```
sborrazas, line 92: `get_transaction` will always return a transaction, no need for `{:ok, transaction}` or to include it inside the `with` statement

lib/ae_mdw/db/rocksdb.ex
```diff
@@ -0,0 +1,264 @@
...
+
+  @doc """
+  Puts changes into an optimistic transaction.
+
+  RocksDB raises an exception on conflict which is helpful to identify unwanted parallelism in the future when
+  it doesn't respect dependency.
+  """
+  @spec put(table(), binary(), binary()) :: :ok | :error
+  def put(table, key, value) do
+    with {_db_ref, cf_ref} <- cf_refs(table),
```
sborrazas, line 91: `cf_refs` will always return a value of the form `{db_ref, cf_ref}`, no need to include it inside the `with` statement

lib/ae_mdw/db/rocksdb.ex
```diff
@@ -0,0 +1,264 @@
...
+         {:ok, t_ref} <- get_transaction() do
+      :ok = :rocksdb.transaction_put(t_ref, cf_ref, key, value)
+    else
+      error ->
```
sborrazas, line 95: All operations here don't return any error, so there no need to handle it

lib/ae_mdw/db/rocksdb.ex
```diff
@@ -0,0 +1,264 @@
...
+
+  @doc """
+  Commits changes made with put/3
+  """
+  @spec commit() :: :ok | :error
+  def commit() do
+    case get_existing_transaction() do
+      {:ok, t_ref} ->
+        :ok = :rocksdb.transaction_commit(t_ref)
+        :ok = :persistent_term.put({__MODULE__, :transaction}, nil)
```
sborrazas, line 109: No need to pattern-match the result here

lib/ae_mdw/db/rocksdb.ex
```diff
@@ -0,0 +1,264 @@
...
+  Creates a transaction and commits a list of mutations.
+  """
+  @spec commit([Mutation.t()]) :: :ok | :error
+  def commit(mutation_list) do
+    {:ok, t_ref} = new_transaction()
+
+    Enum.each(mutation_list, &Mutation.mutate/1)
+
+    :ok = :rocksdb.transaction_commit(t_ref)
+    :ok = :persistent_term.put({__MODULE__, :transaction}, nil)
```
sborrazas, line 126: No need to pattern-match the result here

lib/ae_mdw/db/rocksdb.ex
```diff
@@ -0,0 +1,264 @@
...
+          {:ok, binary()} | :not_found | {:error, {:corruption, charlist()}} | {:error, any()}
+  def get(table, key) do
+    {db_ref, cf_ref} = cf_refs(table)
+    :rocksdb.get(db_ref, cf_ref, key, [])
+  end
+
+  @doc """
+  Iterator for a column family.
+  """
+  @spec iterator(table(), Keyword.t()) :: {:ok, iterator()} | {:error, any()}
```
sborrazas, line 159: :pray:

lib/ae_mdw/db/rocksdb.ex
```diff
@@ -0,0 +1,264 @@
...
+
+  defp get_transaction() do
+    case :persistent_term.get({__MODULE__, :transaction}, nil) do
+      nil -> new_transaction()
+      t_ref -> {:ok, t_ref}
+    end
+  end
+
+  defp new_transaction() do
+    case :rocksdb.transaction(db_ref(), sync: true) do
```
sborrazas, line 250: This function will always return `{:ok, transaction_handle}`

sborrazas (commented):
Just some minor comments regarding typespecs, rest LGTM. Thanks!

lib/ae_mdw/database.ex
jyeshe: yes, some of the returns were already checked by `RocksDbCF` but it was a good thing we review it.

lib/ae_mdw/db/rocksdb.ex
jyeshe: It was added for not existing table but changed to let it crash (since it would be a dev issue). thanks!

jyeshe:
@sborrazas done thanks for the type findings, it made the code cleaner.

sborrazas (approved):

Issue #485, opened 2022-01-24
aens transfer doesn't extend expiration


jyeshe:
This is a blocker for the syncing process:
```
iex(aeternity@localhost)1> 14:35:28.865 [error] Error in process <0.2105.0> on node aeternity@localhost with exit value:
[...]

PR #488, opened 2022-01-26
fix: name and oracle int calls filtering

jyeshe:
not yet the fix for sync issue (#485) as there are no `AENS.claim` internal calls but this fix is needed


sborrazas (approved):
Nice catch!

thepiwo (approved):

Issue #386, opened 2021-12-09
obfuscate cursor to not encourage setting it manually


thepiwo:

jyeshe:
Suggestion: remove conflicting info of cursor and ranges.

thepiwo:
1. isn't the range included to limit the results?
2. could be, but I think its good to keep

jyeshe:
1. The limit of the results is performed by `limit` param for `n` items before or after the cursor. 
2. Okay, that's fine too. `cursor` stays an index (encoded or not) to a record of a table.

jyeshe:
Providing an example on why the range is redundant:
http://localhost:4000/stats/gen/1100-2000?cursor=1100&limit=100
(while the actual range is cursor 1100 + 100)

thepiwo:
if it includes gen 1100 to 2000, what keeps it from going over the borders if we don't know the generation any more?

jyeshe:
The proposal is to remove the 1100 to 2000. `cursor` + `limit` is the actual range.

thepiwo:
how would `/stats/forward?cursor=1100&limit=100` know to stop at gen 2000?

jyeshe:
`/stats/gen/1100-2000?cursor=1100&limit=100` doesn't stop at 2000 (:
the range is ignored and `cursor` + `limit` is applied.

thepiwo:
hmm it should probably do stop if the gen is specified like this, when we have 1100- it can be unlimited

jyeshe:
@thepiwo It stops before the 2000 because it uses the `cursor + limit`. I would stop at `1100 + 100`.
I don't consider an operational issue but only a design one (to remove part of the path that is not used).

thepiwo:
so currently `/stats/gen/1100-2000/forward?cursor=1100&limit=100` then next and next, ..., would not be limited by the `gen/1100-2000`? That sounds like a bug to me and should be fixed.

thepiwo:
@sborrazas @jyeshe any update on this?

PR #499, opened 2022-01-28
refactor: extract range independently of the direction requested

sborrazas:
This simplifies pagination code a bit for us be able to fetch the same
stream of resources with the same scope, independently of the
direction.

### DIFF
--- a/lib/ae_mdw/collection.ex
+++ b/lib/ae_mdw/collection.ex
@@ -33,7 +33,12 @@ defmodule AeMdw.Collection do
   """
   @spec stream(table(), direction(), scope(), cursor()) :: Enumerable.t()
   def stream(tab, direction, scope, cursor) do
-    {first, last} = scope || {nil, nil}
+    {first, last} =
+      case {scope, direction} do
+        {nil, _dir} -> {nil, nil}
+        {s, :forward} -> s
+        {{last, first}, :backward} -> {first, last}
+      end
 
     case fetch_first_key(tab, direction, first, cursor) do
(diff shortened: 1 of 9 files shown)


thepiwo (approved):

jyeshe (commented):
This is an option that is good to simplify the pagination. 

An alternative would be removing the range as it holds the same info as the combination of cursor and limit params. 

Related to the comments of this issue (indepedently of the obfuscation or the cursor):
https://github.com/aeternity/ae_mdw/issues/386

sborrazas:
@jyeshe the range holds more information than the cursor + limit. Since you can't know up to what point you should continue retrieving data.

jyeshe:
> @jyeshe the range holds more information than the cursor + limit. Since you can't know up to what point you should continue retrieving data.

This is the current `next` for `/stats`:
/stats/gen/1100-2000?cursor=1100&limit=100

The proposal would be to remove the "1100 to 2000" as it's ignored by the endpoint. Instead currently the actual range is `cursor` (1100) until `cursor + limit` (1200).

So this would be the proposed `next`:
/stats/forward?cursor=1100&limit=100

where keeping the direction would have the benefit of standardizing across multiple paths that doesn't have range but will always have direction nonetheless.

jyeshe (approved):
My comment applies only to `/stats`, `/totalstats` and `/blocks`, for the other endpoints the range would be a better fit (more flexible filtering).

Integration tests passed except for known:
test/integration/ae_mdw_web/controllers/name_controller_test.exs:166
test/integration/ae_mdw_web/controllers/oracle_controller_test.exs:182

involving oracle and inactive name expirations.

PR #505, opened 2022-02-03
feat: /aex9/by_contract search

### LINKED ISSUE #489: new /aex9/by_contract endpoint path
Same response fields as /aex9/by_name (except for contract_id that could be omitted since it's passed as param) but filtering by contract_id

jyeshe:
Resolves #489

### DIFF
--- a/lib/ae_mdw_web/controllers/aex9_controller.ex
+++ b/lib/ae_mdw_web/controllers/aex9_controller.ex
@@ -15,6 +15,10 @@ defmodule AeMdwWeb.Aex9Controller do
 
   @max_range_length 10
 
+  @spec by_contract(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def by_contract(conn, %{"id" => contract_id}),
+    do: handle_input(conn, fn -> by_contract_reply(conn, contract_id) end)
+
   @spec by_names(Plug.Conn.t(), map()) :: Plug.Conn.t()
   def by_names(conn, params),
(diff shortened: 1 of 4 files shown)


thepiwo (approved):

lib/ae_mdw_web/controllers/aex9_controller.ex
```diff
@@ -177,6 +181,16 @@ defmodule AeMdwWeb.Aex9Controller do
   #
   # Private functions
   #
+  defp by_contract_reply(conn, contract_id) do
+    entry =
+      case Contract.aex9_search_contract_by_id(contract_id) do
+        {:ok, rev_aex9_key} -> Format.to_map(rev_aex9_key, Model.RevAex9Contract)
+        :not_found -> %{}
```
sborrazas, line 188: Wouldn't we want to return 404 here?

jyeshe: We could make it different on v2 even though it's not a rest api. 
For now it would be better like this because the other kinds of aex9 searches return 200 with empty result:
https://mainnet.aeternity.io/mdw/aex9/by_name?exact=AAAB

sborrazas (approved):

lib/ae_mdw_web/controllers/aex9_controller.ex
sborrazas: You're right, for backwards compat we should keep this behavior

PR #514, opened 2022-02-05
feat: aex9 contract created by call tx

### LINKED ISSUE #425: index aex9 contracts with chain.clone and chain.create

jyeshe:
## What

Index AEX9 contracts created by a `:contract_call_tx`

## Why

Resolves #425

## Validation steps

1. setup testnet
2. `elixir --sname aeternity@localhost -S mix test.integration test/integration/ae_mdw/sync/contract_call_mutation_test.exs:149`

### DIFF
--- a/lib/ae_mdw/db/sync/contract.ex
+++ b/lib/ae_mdw/db/sync/contract.ex
@@ -19,6 +19,7 @@ defmodule AeMdw.Db.Sync.Contract do
   alias AeMdw.Node.Db
   alias AeMdw.Sync.AsyncTasks
   alias AeMdw.Txs
+  alias AeMdw.Validate
 
   require Model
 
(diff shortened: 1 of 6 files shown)


lib/ae_mdw/db/sync/contract.ex
```diff
@@ -52,6 +53,42 @@ defmodule AeMdw.Db.Sync.Contract do
     :ok
   end
 
+  @spec child_contract_mutations(
+          boolean(),
+          Contract.fun_arg_res_or_error(),
+          Txs.txi(),
+          Txs.tx_hash()
+        ) ::
+          {[Mutation.t()], Contract.aex9_meta_info() | nil}
```
sborrazas, line 62: Perhaps a smaller type signature for this would be `{:ok, [Mutation.t()], Contract.aex9_meta_info()} | :not_found`, since there's no case for `{:ok, [Mutation.t()], Contract.aex9_meta_info()}` where `[Mutation.t()]` isn't empty

jyeshe: The problem is that the aex9 might be present or not when the origin mutations is not empty. 
from 
`{:ok, [Mutation.t()], Contract.aex9_meta_info()} | :not_found`
we would need
`{:ok, [Mutation.t()], Contract.aex9_meta_info() | nil} | :not_found`
that would look more complex than:
`{[Mutation.t()], Contract.aex9_meta_info() | nil}`

sborrazas: I hadn't realized about that case where there could be mutations and the info would still be `nil`. Thanks for clearing that up!

sborrazas (approved):

thepiwo (approved):

PR #526, opened 2022-02-10
feat: add prev link cursor on paginated endpoints

sborrazas:


lib/ae_mdw/collection.ex
```diff
@@ -4,28 +4,53 @@ defmodule AeMdw.Collection do
...
+      direction
+      |> stream_fn.()
+      |> Stream.take(limit + 1)
+      |> Enum.split(limit)
+      |> case do
+        {records, []} -> {prev_cursor, records, nil}
+        {records, [next_cursor]} -> {prev_cursor, records, {next_cursor, false}}
+      end
+
+    # IO.inspect(["PAGINATION", p, res])
```
jyeshe, line 44: nit pick for merging

lib/ae_mdw_web/controllers/block_controller.ex
```diff
@@ -43,42 +43,26 @@ defmodule AeMdwWeb.BlockController do
...
-      if next_cursor do
-        %URI{
-          path: path,
-          query: URI.encode_query(%{"cursor" => next_cursor, "limit" => limit})
-        }
-        |> URI.to_string()
-      end
-
-    json(conn, %{"data" => blocks, "next" => uri})
+    paginate(conn, prev_cursor, blocks, next_cursor)
```
jyeshe, line 52: :100:

jyeshe (changes requested):
I like the general idea and the integration tests looks trustworthy.

There are some issues with opportunity for higher coverage:
1. http://localhost:4000/txs/forward
then go to next
http://localhost:4000/txs/forward?cursor=10
then go to prev
/txs/forward?cursor=9&rev=1 
gives  `error: "invalid id: 1"`
2. `prev` could be nil for all paths on first hit without cursor likewise `/blocks/forward`
3. similar cases to 1. are (possibly not be limited to)
http://localhost:4000/contracts/calls/forward
http://localhost:4000/contracts/logs/forward

thepiwo (commented):
we should probably document that in the readme?

jyeshe (commented):
Code looks good to me and `/txs/forward` going next and prev works.
Missing only the same `invalid rev` fix for:
http://localhost:4000/contracts/calls/forward
http://localhost:4000/contracts/logs/forward

sborrazas:
@jyeshe done, thanks

sborrazas:
> we should probably document that in the readme?

I've now updated the README to include the `prev` parameter in the instructions. Thanks for reminding me!

jyeshe (approved):

Issue #498, opened 2022-01-28
plan for introduction of breaking changes


thepiwo:
- make breaking changes routes `/v2/` prefixed
 - deploy 1.X.X and 2.X.X
 
 discuss route versioning

sborrazas:
I think the approach we should take when it comes to routes versioning should be by prefixing the version of the API to all the routes. Similar to how many APIs do it, including [the Twitter API](https://developer.twitter.com/en/docs/twitter-api).

If we use this mechanism I would suggest using the following approach for each of these cases:
1. If a new route if needed, we simply add it to the current (latest) version
2. If a route needs to be removed we simply remove it up until the next version
3. If a route's request/response needs to be changed then:
  A. If it's an additive change (e.g. adding new attributes to the paginated results map) we can simply update it on the current version
  B. If it's a non-additive (breaking) change we either add a new route with a new path and keep the previous one and deprecate the previous one on the next version OR we simply update it on the next version
  
The effort involved to get there now would be to simply add a `/v2/` prefix to all of the current routes that we have, while simultaneously keeping the previous ones. We should also start thinking about a process on HOW to notify users of deprecations which will happen when users update to the next version (say `/v3/`). We should also think about WHEN we stop  supporting the previous versions (after a year perhaps?)

thepiwo:
As discussed we want to remove the deprecated "v1" endpoints, for this we will check usage on the hosted infrastructure to see how widely its still used and coordinate with internal and external teams to move to v2

jyeshe:
> As discussed we want to remove the deprecated "v1" endpoints, for this we will check usage on the hosted infrastructure to see how widely its still used and coordinate with internal and external teams to move to v2

Removal of v1 was initiated and is scheduled to 1st May.

thepiwo:
any update on this?

sborrazas:
@thepiwo there's still the old explorer that's publicly exposed which uses v1 endpoints. This is the reason we can't get rid of v1 yet

PR #530, opened 2022-02-14
feat: add /v2 routes to support versioning

sborrazas:
From now on, all new routes should be added only to the latest
version, which will be released at some point in the future.

Older routes will be announced for deprecation and deprecated
accordingly.

Refs #498

### DIFF
--- a/README.md
+++ b/README.md
@@ -306,7 +306,7 @@ Type groups for the transactions listed above are:
 
 `type` parameter:
 ```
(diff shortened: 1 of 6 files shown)


thepiwo (commented):
this will also need to be adjusted in the docs, right?

jyeshe (commented):
We have tests for a /v2-only path, could we have one or two also for a shared path that is requested with /v2?

jyeshe (approved):

thepiwo (approved):

PR #531, opened 2022-02-15
fix: fix missing streams errors 

sborrazas:


jyeshe (changes requested):
Solved the test case that was not passing and refactoring looks good.
Only one detail on error handling.

lib/ae_mdw/blocks.ex
```diff
@@ -73,6 +73,19 @@ defmodule AeMdw.Blocks do
     hash
   end
 
+  @spec fetch_txis_from_gen(height()) :: Enumerable.t()
+  def fetch_txis_from_gen(height) do
+    with Model.block(tx_index: tx_index_start)
+         when is_integer(tx_index_start) and tx_index_start >= 0 <-
+           Database.fetch!(@table, {height, -1}),
```
jyeshe, line 80: part of the 500 trap

thepiwo (approved):

jyeshe (approved):
Approving as I need this fix to conclude the rocksdb PR. The 500 happens only when the height is not yet present on Mdw so we might have a minor fix on a following PR if you please.

sborrazas:
@jyeshe fixed, thank you. Merging now

Issue #536, opened 2022-02-18
/aex9/by_name internal server error


jyeshe:
```
[error] CRASH REPORT Process <0.4096.0> with 0 neighbours exited with reason: {{{{#{'__exception__' => true,'__struct__' => 'Elixir.ArgumentError',message => <<"argument error">>},[{erlang,iolist_to_binary,[nil],[]},{crypto,hash,2,[{file,"crypto.erl"},{line,604}]},{aeser_api_encoder,check_str,1,[{file,"/home/rogerio/Workspace/ae/aeternity/_build/default/lib/aeserialization/src/aeser_api_encoder.erl"},{line,156}]},{aeser_api_encoder,base58_check,1,[{file,"/home/rogerio/Workspace/ae/aeternity/_build/default/lib/aeserialization/src/aeser_api_encoder.erl"},{line,148}]},...]},...},...},...}
```

thepiwo:
only on testnet, still under investigation

jyeshe:
[25855475](https://testnet.aeternity.io/mdw/txi/25855475) is a contract call that creates `ct_MW1jv3erANqbwLxncgdD15GuHTQpppRpgPq51ntgE97V6Wx8G`

```
iex(aeternity@localhost)2> AeMdw.Validate.id! "ct_MW1jv3erANqbwLxncgdD15GuHTQpppRpgPq51ntgE97V6Wx8G"
<<46, 140, 90, 150, 66, 91, 192, 187, 176, 56, 15, 126, 65, 1, 21, 203, 189, 63,
  147, 234, 239, 209, 153, 5, 179, 228, 185, 187, 35, 56, 82, 235>>
```

jyeshe:
Affects also [/aex9/by_contract/](https://testnet.aeternity.io/mdw/aex9/by_contract/ct_MW1jv3erANqbwLxncgdD15GuHTQpppRpgPq51ntgE97V6Wx8G)

PR #540, opened 2022-02-21
fix: raise exception when aex9 contract doesn't exist

sborrazas:
refs #536

### DIFF
--- a/lib/ae_mdw/db/format.ex
+++ b/lib/ae_mdw/db/format.ex
@@ -116,7 +116,7 @@ defmodule AeMdw.Db.Format do
       symbol: symbol,
       decimals: decimals,
       contract_txi: txi,
-      contract_id: :aeser_id.create(:contract, Origin.pubkey({:contract, txi}))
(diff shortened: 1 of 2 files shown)


thepiwo (approved):

jyeshe (changes requested):
The problem currently is an Internal Server Error. With another exception raised the endpoint will continue to be unavailable.

sborrazas:
@jyeshe this won't fix the problem, but it will make the error more explicit and actually uses properly typed functions

jyeshe:
> @jyeshe this won't fix the problem, but it will make the error more explicit and actually uses properly typed functions

On the other banging behaviour (`Txs.fetch!` for instance) the Mdw is not raising an error, it just let the Beam raise its own error. When it raises an exception on endpoints, it uses `AeMdw.Error` to send a message to the user regarding a wrong input.

But if we wanna add other kind of exceptions, it's ok for me.

jyeshe (approved):

PR #548, opened 2022-02-23
chore: withhold non-migrated v2 routes

sborrazas:
refs #498


jyeshe (commented):
README partial revert would be good too.

jyeshe (approved):

PR #551, opened 2022-02-24
feat: aex9 transfers pagination

### LINKED ISSUE #545: add pagination to aex9 transfers
https://mainnet.aeternity.io/mdw/aex9/transfers/from/ak_fCCw1JEkvXdztZxk8FRGNAkvmArhVeow89e64yX4AxbCPrVh5
was giving timeout due to the 45k entries for this sender.

jyeshe:
## What

Adds pagination to `/aex9/transfer` within `/v2` (`/v1` result is without `data`)

This allows navigating backwards or forward upon limit based on call txi for the 3 kinds of transfers:
- from sender
- to recipient
- from-to pair 

There is a migration that reindexes the aex9 transfers.

## Why

Fixes #545

## Validation steps

```
INTEGRATION_TEST=1 elixir --sname aeternity@localhost -S mix test.integration test/integration/ae_mdw_web/controllers/aex9_controller_test.exs

Finished in 2.8 seconds
24 tests, 0 failures
```


lib/ae_mdw/aex9.ex
```diff
@@ -0,0 +1,138 @@
...
+    cursor_key =
+      deserialize_cursor(cursor) || default_pair_cursor(direction, sender_pk, recipient_pk)
+
+    paginate_transfers(
+      pagination,
+      Model.Aex9PairTransfer,
+      fn {s, r, _txi, _amount, _idx} -> s == sender_pk && r == recipient_pk end,
+      cursor_key
+    )
+  end
```
sborrazas, line 58: Perhaps this logic can be better achieved using the scope and avoid to manually fetch the first element of the table as a default cursor:
```elixir
scope = {
  {sender_pk, receiver_pk, Util.min_int(), nil, nil}, 
  {sender_pk, receiver_pk, Util.max_int(), nil, nil}
}

fn direction ->
  Collection.stream(Model.Aex9Pairtransfer, direction, scope, cursor)
end
|> Collection.paginate(pagination)
```

This way we can also avoid inspecting the pagination tuple and having to use a key checker function

jyeshe: Let me divide the problem in two:
1) the fetching of the first key
2) the use of scopes

1) For this I have preserved the style of existing `Collection.stream` which is to include the first_key in the stream while unfolding. 
new:
https://github.com/aeternity/ae_mdw/blob/8389238209755e1b779c68adc0576975bca42b1b/lib/ae_mdw/collection.ex#L79
existing:
https://github.com/aeternity/ae_mdw/blob/8389238209755e1b779c68adc0576975bca42b1b/lib/ae_mdw/collection.ex#L123 
that happens after
https://github.com/aeternity/ae_mdw/blob/8389238209755e1b779c68adc0576975bca42b1b/lib/ae_mdw/collection.ex#L66
We may change both but it would be better wait for a more opportune time so I can get back faster to RocksDB transition.

2) The main reason I have avoided using scope was because part of it is already represented in the pagination and the other part was not applicable to aex9 transfers.

As per definition, conceptually speaking in the context of Mdw, we know that "Scope specifies the time period to look for transactions matching the criteria, as well as direction":

- forward   - from beginning (genesis) to the end
- backward  - from end (top of chain) to the beginning
- gen/A-B   - from generation A to B (forward if A < B, backward otherwise)
- txi/A-B   - from transaction index A to B (forward if A < B, backward otherwise)

So doing the mapping, `forward/backward` was already present in pagination (preserving the same structure as other modules like `AeMdw.Txs`, `AeMdw.Blocks`, ...) and gen/A-B or txi/A-B not applicable to /aex9/transfers.

I could create a scope like boundaries that is not a scope as above but avoided mixing the concept.
   
Gladly, without any scope, it allowed the code to be much simpler for this endpoint.

Code without scope:
```
def stream(table, direction, cursor_key, upper_bound_key) do
    Stream.unfold(
      cursor_key,
      fn key ->
        with true <- key != :none,
             {:ok, next_key} <- Database.next_key(table, direction, key) do
          {key, next_key}
        else
          false -> nil
          :none -> {key, :none}
        end
      end
    )
  end
```

Pre-existing code with scope (that has it's purpose for blocks, txs, ...):
```
def stream(tab, direction, scope, cursor) do
    {first, last} =
      case {scope, direction} do
        {nil, _dir} -> {nil, nil}
        {s, :forward} -> s
        {{last, first}, :backward} -> {first, last}
      end

    case fetch_first_key(tab, direction, first, cursor) do
      {:ok, first_key} -> unfold_stream(tab, direction, first_key, last)
      :none -> []
    end
  end

  # followed by
  defp unfold_stream(tab, direction, first_key, last_key) do
    stream =
      Stream.unfold(first_key, fn
        :end_keys ->
          nil

        key ->
          case Database.next_key(tab, direction, key) do
            {:ok, next_key} -> {key, next_key}
            :none -> {key, :end_keys}
          end
      end)

    if last_key do
      Stream.take_while(stream, fn key ->
        if direction == :forward, do: key <= last_key, else: key >= last_key
      end)
    else
      stream
    end
  end
```

The idea is to let the `build_streamer` have the responsibility to define what will be streamed. The `Collection.stream` was more generic, it streams keys from a starting point, without filtering. The benefit from this will be to have more flexibility to add new filters (by amount for instance). Another thought was to make it explicit what kind of keys are expected to be returned, instead of saying I want something between this and that, it says give me anything like this.

jyeshe: @sborrazas 
Done the item 1) in this PR for the sake of opaque pagination.
For the item 2) even though the previous `build_streamer` is more explicit on what is actually streamed and the `Collection.stream/3` was simpler while not dealing with scopes, the concept of scope was generalized to key boundary in order to reuse the existing `Collection.stream/4`.

jyeshe: But again, not a criticism on existing functions, they have their purpose for blocks, txs, ... and could be easily adapted to aex9 transfers.

jyeshe:
Finished in 33.8 seconds

Top 3 slowest (31.9s), 94.2% of total time:

  * test transfers_from_v1 from a sender with many transfers (30985.2ms) (Integration.AeMdwWeb.Aex9ControllerTest)
  * test balances_range gets balances on a contract for range of generations (609.7ms) (Integration.AeMdwWeb.Aex9ControllerTest)
  * test balances gets all accounts balances for a contract (322.6ms) (Integration.AeMdwWeb.Aex9ControllerTest)

24 tests, 0 failures

lib/ae_mdw/collection.ex
```diff
@@ -54,10 +54,10 @@ defmodule AeMdw.Collection do
   @doc """
   Builds a stream from records from a table starting from the initial_key given.
   """
-  @spec stream(table(), direction(), scope(), cursor()) :: Enumerable.t()
-  def stream(tab, direction, scope, cursor) do
+  @spec stream(table(), direction(), key_boundary(), cursor()) :: Enumerable.t()
+  def stream(tab, direction, key_boundary, cursor) do
     {first, last} =
-      case {scope, direction} do
+      case {key_boundary, direction} do
```
sborrazas, line 60: Love the renaming, we can now refer to scope in the context of generations/txis and boundaries in the context of database page boundaries!

sborrazas (approved):

thepiwo (approved):

PR #553, opened 2022-02-28
refactor: mutations derive to default impl

jyeshe:
## What/Why

Make declarative implementations for the `Mutation` and `TxnMutation` protocols to simplify the transition to transaction mutations when it's resumed.

## Validation steps

Resync and see that these endpoints work for height 10k:
http://localhost:4000/blocks/forward
http://localhost:4000/txs/forward
http://localhost:4000/names


thepiwo (approved):

sborrazas (approved):
Didn't know that you could do this. Looks better now! :muscle:

PR #563, opened 2022-03-03
feat: add aex9 v2 endpoints

sborrazas:


thepiwo (approved):

lib/ae_mdw/aex9.ex
```diff
@@ -85,6 +105,159 @@ defmodule AeMdw.Aex9 do
     )
   end
 
+  @spec fetch_tokens(pagination(), query(), order_by(), cursor() | nil) ::
+          {:ok, cursor() | nil, [aex9_token()], cursor() | nil} | {:error, Error.t()}
+  def fetch_tokens(pagination, query, order_by, cursor) do
+    try do
```
jyeshe, line 111: fallback_controller would be helpful to standardize wherever an exception happens it's raised immediately in place of being returned as a struct.

jyeshe: okay the other way around like this.

PR #564, opened 2022-03-03
feat: allow mdw sync module to restart after a failure

sborrazas:
Also, remove the Dynamic syncing supervisor from the application
startup. The ability to start/stop the mdw syncing at runtime is not
really needed and complicates an the processes communication.

refs #388


thepiwo (approved):

lib/ae_mdw/application.ex
```diff
@@ -218,6 +225,18 @@ defmodule AeMdw.Application do
...
   defp init(:aesync), do: Application.ensure_all_started(:aesync)
 
+  defp init(:tables) do
+    :ets.new(:tx_sync_cache, [:named_table, :ordered_set, :public])
+    :ets.new(:name_sync_cache, [:named_table, :ordered_set, :public])
+    :ets.new(:oracle_sync_cache, [:named_table, :ordered_set, :public])
+    :ets.new(:aex9_sync_cache, [:named_table, :ordered_set, :public])
+    :ets.new(:derive_aex9_presence_cache, [:named_table, :duplicate_bag, :public])
+    :ets.new(:ct_create_sync_cache, [:named_table, :ordered_set, :public])
+    :ets.new(:stat_sync_cache, [:named_table, :ordered_set, :public])
```
jyeshe, line 235: with restarts the `:stat_sync_cache` need to resume counting from the point of the exception.

jyeshe (changes requested):

lib/ae_mdw/application.ex
sborrazas: Once the exception occurs, the new process is going to start syncing from the last generation. And whenever a new generation is synced, the cache tables are cleared. So it wouldn't have the wrong values associated to them

jyeshe: true, it's already reset and recomputed in case of resume, nice! as we've discussed only a minor change is needed for `:name_sync_cache` and `:oracle_sync_cache`

sborrazas: Done

jyeshe (approved):

PR #567, opened 2022-03-04
chore: added check script for readme routes

### LINKED ISSUE #566: automated readme navigator

thepiwo:
resolves #566

### DIFF
--- a/check_endpoints.sh
+++ b/check_endpoints.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+BASE_URL="https://mainnet.aeternity.io/mdw/"
+
+ERROR=0
+SUCCESS=0
+
+while read LINE; do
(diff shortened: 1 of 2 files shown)


check_endpoints.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+BASE_URL="https://mainnet.aeternity.io/mdw/"
+
+cat README.md | grep -q "$REGEX"
```
jyeshe, line 5: what you think about a default `REGEX`? would it supposed to be the `BASE_URL`?

jyeshe (commented):
Nice combination of curl and bash tricks.

check_endpoints.sh
thepiwo: was from debugging

jyeshe (approved):

sborrazas (approved):

PR #578, opened 2022-03-09
chore: use in-memory database when dealing with latest changes

sborrazas:
This is a draft which can be improved in many ways, but ultimately
this is the intended goal:

* Keep changes from the last X generations or microblocks
  in-memory (could be 1 microblock, or more)
* Easily revert these changes which weren't yet committed
* To be able to asynchronously write these in-memory changes to the
  database without having to wait for these changes to be done to
  continue syncing
* Transform the Mutation.execute/2 function into a 100% stateless
  function, which would only return the new state which contains the
  new changes to be migrated. Leaving only the State.commit/1 function
  to actually perform these changes (asynchronously)

Pending improvements:
* Instead of keeping the state of the un-committed changes in the
  State genserver state, use :persistent_term to allow other
  processes (e.g. webserver/controller processes) to inspect these
  changes
* Read these un-committed changes from the context modules and from
  the Collection module

Attached there's a diagram of the communication between the layers of the system:

image

### DIFF
--- a/lib/ae_mdw/application.ex
+++ b/lib/ae_mdw/application.ex
@@ -43,6 +43,7 @@ defmodule AeMdw.Application do
     :ok = AeMdw.Db.RocksDb.open()
 
     children = [
+      AeMdw.Db.State,
       AeMdw.Sync.Watcher,
       AeMdw.Sync.AsyncTasks.Supervisor,
       AeMdwWeb.Supervisor,
--- a/lib/ae_mdw/database.ex
+++ b/lib/ae_mdw/database.ex
@@ -13,7 +13,6 @@ defmodule AeMdw.Database do
 
   alias AeMdw.Db.Model
   alias AeMdw.Db.Mutation
-  alias AeMdw.Db.TxnMutation
   alias AeMdw.Db.RocksDb
   alias AeMdw.Db.RocksDbCF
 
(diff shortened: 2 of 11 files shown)


jyeshe:
This is an approach, one that consider the latest blocks as separate feed/data source which requires specific implementations to `insert`, `fetch`, `merge` with database data, `persist/commit` and `invalidate`. Even though there would be a lot of new code to implement, an important positive aspect is the simplicity on invalidation (second bullet) and the easiness to discriminate which data is from the last blocks helping to debug during development and bug fixes. Still there would be a lot of new code to implement the separate feed and the persistence of it in a timely manner with following blocks.

Another approach would be to consider the latest changes as a regular database transaction, without differentiating latest blocks from the other ones, having a single data source. The `insert`, `fetch`, `persist` would be the same as current one, there would be no need to merge the data while fetching on endpoints and the current invalidate code would deal with latest invalidation after being adapted to delete through transactions (first delete on transaction, if not present on transaction, then delete persisted). It looks like the change would be localized only on the invalidation and the `Collection` in a way replacing read operations by dirty read ones, i.e read from latest db transaction besides persisted data.

One 3rd alternative is a hybrid one, the `insert`, `fetch`, `merge` with database data and invalidation would be done as the first approach and the persistence is dealt like the 2nd and actually by sync process as it is currently. When the latest key block becomes not the latest, it is saved by current flow (`AeMdw.Db.Sync` Listener -> `AeMdw.Db.Sync.Transaction` sync process).

From this 3, I would be more inclined to refine the 2nd, check if it would really work testing it for one transaction type (contract call let's say) but the 3rd (which is based on the 1st) has also its benefits in case the scope of the data to be ingested into memory and exposed to endpoints is made of few tables.

thepiwo:
all sounds really good for me!

> * Keep changes from the last X generations or microblocks
>   in-memory (could be 1 microblock, or more)

I would suggest to aim for a minimum of 6 key-blocks, if working and performant maybe up to 18-100 key-blocks, maybe even configurable.

Issue #529, opened 2022-02-14
Use RocksDB direclty without :mnesia


jyeshe:
https://github.com/aeternity/ae_mdw/issues/346#issuecomment-1017769284

jyeshe:
## Benchmarking RocksDb vs Mnesia

### Setup
warmup: 2 s
time: 5 s
memory time: 1 s
parallel: 1
Elixir 1.10.4
Erlang 22.3.4.24

### 100 writes
```
Benchmarking mensia_write_100...
Benchmarking rocksdb_txn_100...

Name                       ips        average  deviation         median         99th %
mensia_write_100        189.45        5.28 ms   ±266.88%        4.68 ms       10.25 ms
rocksdb_txn_100         113.70        8.79 ms   ±347.66%        5.61 ms       38.06 ms

Comparison: 
mensia_write_100        189.45
rocksdb_txn_100         113.70 - **1.67x slower** +3.52 ms
Elixir 1.10.4
Erlang 22.3.4.24
```

### 1000 writes
```
Benchmarking mensia_write_1k...
Benchmarking rocksdb_txn_1k...

Name                      ips        average  deviation         median         99th %
rocksdb_txn_1k          58.09       17.21 ms   ±212.68%       14.57 ms       36.74 ms
mensia_write_1k         21.00       47.62 ms     ±7.76%       47.76 ms       57.97 ms

Comparison: 
rocksdb_txn_1k          58.09
mensia_write_1k         21.00 - **2.77x slower** +30.41 ms

Memory usage statistics:

Name                    average  deviation         median         99th %
rocksdb_txn_1k          0.70 MB     ±0.00%        0.70 MB        0.70 MB
mensia_write_1k         2.85 MB     ±0.00%        2.85 MB        2.85 MB

Comparison: 
rocksdb_txn_1k          0.70 MB
mensia_write_1k         2.85 MB - **4.06x memory usage** +2.15 MB
```

### 10000 writes
```
Benchmarking mensia_write_10k...
Benchmarking rocksdb_txn_10k...

Name                       ips        average  deviation         median         99th %
rocksdb_txn_10k           6.96      143.58 ms    ±99.31%      107.87 ms      717.65 ms
mensia_write_10k          2.08      480.18 ms    ±10.21%      489.23 ms      523.02 ms

Comparison: 
rocksdb_txn_10k           6.96
mensia_write_10k          2.08 - **3.34x slower** +336.60 ms

Memory usage statistics:

[...]

PR #590, opened 2022-03-15
refactor: chain and name tables persisted with rocksdb

jyeshe:
refs #529

### DIFF
--- a/lib/ae_mdw/collection.ex
+++ b/lib/ae_mdw/collection.ex
@@ -69,6 +69,22 @@ defmodule AeMdw.Collection do
     end
   end
 
+  @spec stream(table(), direction(), key()) :: Enumerable.t()
+  def stream(table, direction, boundary_first_key) do
+    Stream.unfold(
+      boundary_first_key,
+      fn key ->
+        with true <- key != :none,
+             {:ok, next_key} <- Database.next_key(table, direction, key) do
+          {key, next_key}
+        else
+          false -> nil
+          :none -> {key, :none}
+        end
+      end
+    )
+  end
+
   @doc """
   Merges any given stream of keys into a single stream, in sorted order and without dups.
   """
--- a/lib/ae_mdw/collection.ex
+++ b/lib/ae_mdw/collection.ex
@@ -69,17 +69,17 @@ defmodule AeMdw.Collection do
     end
   end
 
-  @spec stream(table(), direction(), key()) :: Enumerable.t()
-  def stream(table, direction, boundary_first_key) do
+  @doc """
+  Streams forward a table seeking the iterator to a boundary start key.
+  """
+  @spec stream(table(), key()) :: Enumerable.t()
+  def stream(table, boundary_start_key) do
     Stream.unfold(
-      boundary_first_key,
+      boundary_start_key,
       fn key ->
-        with true <- key != :none,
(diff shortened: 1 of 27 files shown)


jyeshe:
@sborrazas @thepiwo in case you wanna sync in advance, I have made two syncs until 80k and /names endpoint is working.
There are only test cases to refactor (exclude from mnesia) and a full sync will be made to run the integration tests.

lib/ae_mdw/collection.ex
```diff
@@ -69,6 +69,22 @@ defmodule AeMdw.Collection do
...
+  def stream(table, boundary_start_key) do
+    Stream.unfold(
+      boundary_start_key,
+      fn key ->
+        case Database.next_key(table, key) do
+          {:ok, next_key} -> {next_key, next_key}
+          :none -> nil
+        end
+      end
+    )
```
sborrazas, line 85: If the first key is the same as the boundary start key this function will not return it. Perhaps we can reuse the `unfold_stream` function which is more general?

lib/ae_mdw/db/model.ex
```diff
@@ -131,16 +152,42 @@ defmodule AeMdw.Db.Model do
...
+          )
+
   # owner: (updated via name claim/transfer)
   #     index = {pubkey, entity},
   @owner_defaults [index: nil, unused: nil]
   defrecord :owner, @owner_defaults
 
+  @type owner() ::
+          record(:owner,
+            index: {Db.pubkey(), binary()},
```
sborrazas, line 176: Wouldn't this `binary()` be `Names.plain_name()` ?

lib/ae_mdw/db/mutations/update_ids_counts_mutation.ex
```diff
@@ -0,0 +1,32 @@
...
+
+  @spec new([update_key()]) :: t()
+  def new(ids_counts) do
+    %__MODULE__{ids_counts: ids_counts}
+  end
+
+  @spec execute(t(), Database.transaction()) :: :ok
+  def execute(%__MODULE__{ids_counts: ids_counts}, txn) do
+    Enum.each(ids_counts, fn {id_count_key, delta} ->
+      IdCounter.update_count(txn, id_count_key, -delta)
```
sborrazas, line 29: Why is the delta negated here? Feels rather counter-intuitive

thepiwo (approved):

lib/ae_mdw/collection.ex
jyeshe: Exactly this is to replace the `:mnesia.select` on a simpler way when we don't have the first key and care only for a pattern like here:
```
    |> Collection.stream({height_pos, kind, <<>>, -1})
    |> Stream.take_while(fn
      {^height_pos, ^kind, _target, _ref} -> true
      _other_height_kind -> false
    end)
```
I will fetch the first key though like the boundary function.

lib/ae_mdw/db/mutations/update_ids_counts_mutation.ex
jyeshe: It's called by the invalidation that deletes some objects/ids. The logic is the same as before, it was only migrated to a mutation avoiding bigger changes on invalidation code because we can have a specific PR for that.

thepiwo (approved):

sborrazas (approved):

PR #594, opened 2022-03-18
refactor: contract tables persisted with rocksdb

jyeshe:
Closes #529 

## Additional Notes
1. Expiration inconsistencies fixes gonna be validated after a full sync.
2. After replacing `:mnesia`/`:mnesia_rocksdb` with `:rocksdb`/`RocksDb`, all chain txs are saved through RocksDB optimistic transactions which has [lower performance](https://github.com/aeternity/ae_mdw/issues/529#issuecomment-1044503185) when writing close to 100 records (key-values). This affects the sync speed for the first 3 million txis (until height 134252), specially for txi between 2M and 3M, since the commit is per microblock and most microblocks has a single tx.

```
iex(aeternity@localhost)1> bi_txs1 = 0..1_000_000 |> Enum.map(&Database.fetch!(Model.Tx, &1)) |> Enum.group_by(fn Model.tx(block_index: bi) -> bi end) |> Enum.map(fn {bi, list} -> {bi, length(list)} end)

iex(aeternity@localhost)2> bi_txs2 = 1_000_001..2_000_000 |> Enum.map(&Database.fetch!(Model.Tx, &1)) |> Enum.group_by(fn Model.tx(block_index: bi) -> bi end) |> Enum.map(fn {bi, list} -> {bi, length(list)} end)

iex(aeternity@localhost)3> bi_txs3 = 2_000_001..3_000_000 |> Enum.map(&Database.fetch!(Model.Tx, &1)) |> Enum.group_by(fn Model.tx(block_index: bi) -> bi end) |> Enum.map(fn {bi, list} -> {bi, length(list)} end)

iex(aeternity@localhost)4> txis = 0..1_000_000; bi_txs = bi_txs1; %{txis: txis, last_bi: Enum.max_by(bi_txs, &elem(&1, 0)), num_mbs: length(bi_txs), num_mbs_single_tx: Enum.count(bi_txs, fn {_bi, len} -> len == 1 end)}                                          
%{
  last_bi: {{44495, 43}, 1},
  num_mbs: 351512,
  num_mbs_single_tx: 199930,
  txis: 0..1000000
}
iex(aeternity@localhost)5> txis = 1_000_001..2_000_000; bi_txs = bi_txs2; %{txis: txis, last_bi: Enum.max_by(bi_txs, &elem(&1, 0)), num_mbs: length(bi_txs), num_mbs_single_tx: Enum.count(bi_txs, fn {_bi, len} -> len == 1 end)}
%{
[...]

PR #598, opened 2022-03-21
fix: missing InactiveNameOwner record

jyeshe:
## What

Single function to deactivate names so that `InactiveNameOwner` (along with `InactiveName` and `InactiveNameExpiration`) is always written on every name deactivation (expiration, update or revoke)

## Why

Missing record when revoked or updated.

## Validation steps

Sync until 80114 on testnet

### DIFF
--- a/lib/ae_mdw/db/name.ex
+++ b/lib/ae_mdw/db/name.ex
@@ -119,17 +119,9 @@ defmodule AeMdw.Db.Name do
   @spec expire_name(transaction(), Blocks.height(), Names.plain_name()) :: :ok
   def expire_name(txn, height, plain_name) do
     m_name = cache_through_read!(txn, Model.ActiveName, plain_name)
-    m_exp = Model.expiration(index: {height, plain_name})
 
     if Model.name(m_name, :expire) == height do
-      owner = Model.name(m_name, :owner)
-      cache_through_write(txn, Model.InactiveName, m_name)
-      cache_through_write(txn, Model.InactiveNameOwner, Model.owner(index: {owner, plain_name}))
-      cache_through_write(txn, Model.InactiveNameExpiration, m_exp)
-      cache_through_delete(txn, Model.ActiveName, plain_name)
-      cache_through_delete(txn, Model.ActiveNameOwner, {owner, plain_name})
-      cache_through_delete(txn, Model.ActiveNameExpiration, {height, plain_name})
-
+      deactivate_name(txn, height, m_name)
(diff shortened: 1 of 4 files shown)


lib/ae_mdw/db/name.ex
```diff
@@ -325,17 +317,10 @@ defmodule AeMdw.Db.Name do
...
-    do: expire
-
-  def revoke_or_expire_height({{height, _}, _}, _expire),
-    do: height
-
   @spec revoke_or_expire_height(Model.name()) :: Blocks.height()
-  def revoke_or_expire_height(m_name),
-    do: revoke_or_expire_height(Model.name(m_name, :revoke), Model.name(m_name, :expire))
+  def revoke_or_expire_height(m_name) do
+    do_revoke_or_expire_height(Model.name(m_name, :revoke), Model.name(m_name, :expire))
```
sborrazas, line 322: Why do we use the `do_` prefix here? Perhaps we can pattern match this function and avoid using non-reusable auxiliary functions?

sborrazas (approved):
optional comment

lib/ae_mdw/db/name.ex
jyeshe: when they were together it helped me reading which was the caller and the called but now with one of them private I agree it's not needed anymore. thanks!

jyeshe:
unrelated on testnet (fixed on #594):
```
iex(aeternity@localhost)1> 09:44:24.669 [error] Supervisor 'Elixir.AeMdw.Db.Sync.Supervisor' had child 'Elixir.AeMdw.Db.Sync' started with 'Elixir.AeMdw.Db.Sync':start_link([]) at <0.22907.5> exit with reason #{'__exception__' => true,'__struct__' => 'Elixir.Protocol.UndefinedError',description => <<>>,protocol => 'Elixir.AeMdw.Db.TxnMutation',value => #{'__struct__' => 'Elixir.AeMdw.Db.WriteMutation',record => {int_contract_call,{2932594,0},-1,<<"Chain.spend">>,{aetx,spend_tx,aec_spend_tx,96,{spend_tx,{id,account,<<84,180,196,235,185,254,235,68,37,168,101,128,127,111,97,136,141,11,134,251,228,200,73,71,175,98,22,115,172,159,234,177>>},{id,account,<<75,51,109,3,204,112,60,147,247,74,72,117,144,...>>},...}}},...}} in 'Elixir.AeMdw.Db.TxnMutation':'impl_for!'/1 line 1 in context child_terminated
```

thepiwo (approved):

PR #599, opened 2022-03-22
test: coverage analysis

jyeshe:


jyeshe:
Coveralls failing on `get_env("GITHUB_EVENT_PATH") |> File.read!()`

.github/workflows/test-action.yml
```diff
@@ -24,6 +24,36 @@ jobs:
...
+    name: Test coverage
+    runs-on: ubuntu-latest
+    
+    env:
+      MIX_ENV: test
+      NODEROOT: ./node/local
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    
+    steps:
+      - uses: actions/checkout@v2
```
thepiwo, line 37: can the steps be combined so we use the previous checked out and built version?

thepiwo (commented):
we discussed to only keep the result in the test logs for now

.github/workflows/test-action.yml
jyeshe: The checkout is needed before running `./.github/actions/` actions. The `node-setup` was changed to a `mdw-node-setup` though. Would any additional composition be required? There will be also another PR to make dialyzer job faster that will make more explicit what steps can be grouped to be reused.

jyeshe:
> we discussed to only keep the result in the test logs for now

Done.

thepiwo (approved):
lgtm, good to see a base number, so we can improve there

sborrazas (approved):

PR #601, opened 2022-03-22
fix: handle name ownership and pointers when tx is internal

sborrazas:


lib/ae_mdw/db/name.ex
```diff
@@ -245,6 +245,46 @@ defmodule AeMdw.Db.Name do
...
+          index: plain_name,
+          claims: [{_block_index, last_claim_txi} | _rest_claims],
+          transfers: transfers,
+          owner: owner
+        )
+      ) do
+    Model.tx(id: tx_hash) = read_tx!(last_claim_txi)
+    {:ok, name_hash} = :aens.get_name_hash(plain_name)
+
+    orig_owner =
```
jyeshe, line 260: When there is no transfer, could it use the `Model.name(:owner)`? This way it would call the Node only there is a transfer.

lib/ae_mdw/db/name.ex
```diff
@@ -245,6 +245,46 @@ defmodule AeMdw.Db.Name do
...
+      case AE.Db.get_tx_data(tx_hash) do
+        {_block_hash, :name_claim_tx, _signed_tx, tx_rec} ->
+          :aens_claim_tx.account_id(tx_rec)
+
+        {_block_hash, :contract_call_tx, _signed_tx, _tx_rec} ->
```
jyeshe, line 265: good finding!

lib/ae_mdw/names.ex
```diff
@@ -326,43 +326,29 @@ defmodule AeMdw.Names do
...
-  end
-
-  defp render_pointers([{_bi, txi} | _rest_updates]) do
-    txi
-    |> Txs.fetch!()
+  defp render_pointers(name) do
+    name
+    |> Name.pointers()
+    |> Enum.reverse()
+    |> Enum.at(0)
```
jyeshe, line 338: Good trick. It would be good to have a function to indicate it's getting the biggest key from pointer_kv_raw map with a comment on why the biggest key is the one that matters (from existing code `:aens_pointer.key(ptr)` it is clear if it would be monotonic according to pointer height).

thepiwo (approved):

lib/ae_mdw/db/name.ex
sborrazas: You're right, good call. Done

lib/ae_mdw/names.ex
sborrazas: Added a comment to it, we should investigate the reason a bit further though

jyeshe (approved):

PR #602, opened 2022-03-24
fix: deactivate name on update with ttl 0

jyeshe:
## What

- Fixes the delete of `ActiveNameExpiration` during a name update with ttl 0
- Expires without any validation (assures that previous state transitions were correct)

## Why

Fixes sync on height [162223](https://github.com/aeternity/ae_mdw/pull/594#issuecomment-1075625216)

## Validation steps

Sync from scratch (delete the `mdw.db` dir) until the height 164809 with:
```
$ git checkout rocksdb-contracts
$ git merge --no-commit fix-update-ttl-0
$ make shell # or docker-compose up
```

### DIFF
--- a/lib/ae_mdw/db/name.ex
+++ b/lib/ae_mdw/db/name.ex
@@ -118,14 +118,11 @@ defmodule AeMdw.Db.Name do
 
   @spec expire_name(transaction(), Blocks.height(), Names.plain_name()) :: :ok
   def expire_name(txn, height, plain_name) do
-    m_name = cache_through_read!(txn, Model.ActiveName, plain_name)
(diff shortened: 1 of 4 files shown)


thepiwo (approved):

sborrazas (approved):
LGTM. I wonder if this would fix the name inconsistencies we had before :thinking:

jyeshe:
> LGTM. I wonder if this would fix the name inconsistencies we had before thinking

This in particular was introduced by the previous PR that moved the deactivation to a single point but detecting errors on delete would have fixed some of them (not all because of the intermittent results).

Issue #604, opened 2022-03-26
/names/active fails on getting pointers


jyeshe:
![image](https://user-images.githubusercontent.com/44991200/160236889-a5c93f7c-19b1-43c7-98f3-82bfd2e981db.png)

With this `Stream.take_while` change it passes:
```
  @spec fetch_int_contract_calls(Txs.txi(), Contract.fname()) :: Enumerable.t()
  def fetch_int_contract_calls(txi, fname) do
    @int_contract_call_table
    |> Collection.stream({txi, @min_int})
    |> Stream.take_while(fn {call_txi, _local_txi} -> call_txi == txi end)
    |> Stream.map(&Database.fetch!(@int_contract_call_table, &1))
    |> Stream.filter(&match?(Model.int_contract_call(fname: ^fname), &1))
  end
  ```

But Collection.stream/4 should be fixed.

sborrazas:
@jyeshe not sure what you mean. What's broken about `Collection.stream/4`, it looks good to me? 

What appears to be broken is that an AENS.update internal event transaction doesn't have an associated aens_update transaction belonging to that name

PR #605, opened 2022-03-26
fix: fix usage of Collection.stream for getting int calls

sborrazas:
Problem was the Collection.stream/4 exepcts the scope to always be
{first, last}, regardless of the direction.

Refs #604

### DIFF
--- a/lib/ae_mdw/contracts.ex
+++ b/lib/ae_mdw/contracts.ex
@@ -100,7 +100,8 @@ defmodule AeMdw.Contracts do
   @spec fetch_int_contract_calls(Txs.txi(), Contract.fname()) :: Enumerable.t()
   def fetch_int_contract_calls(txi, fname) do
     @int_contract_call_table
-    |> Collection.stream(:backward, {{txi + 1, @min_int}, {txi, @min_int}}, nil)
+    |> Collection.stream({txi, @min_int})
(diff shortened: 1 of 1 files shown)


thepiwo (approved):

jyeshe (changes requested):
This fix replacing Collection.stream/4 by Collection.stream/2 had been done here: https://github.com/aeternity/ae_mdw/pull/594/commits/3107d95286514fd433a0b8bfa2837f53e526d91d (before reporting the issue, needed to pass 7 integration tests)

We can wait for a fix for the `Collection.stream/4` instead (we could have a validation depending on the direction).

sborrazas:
@jyeshe `Collection.stream/4` wasn't broken, it was incorrectly used because the scope sent was `{last, first}` instead of `{first, last}` which is the correct way of using it

jyeshe:
> @jyeshe `Collection.stream/4` wasn't broken, it was incorrectly used because the scope sent was `{last, first}` instead of `{first, last}` which is the correct way of using it

I see, I think `Collection.stream/4` should have a validation when the direction is `backward` and the expected first < last is not met.

jyeshe:
This will help us until detecting silent errors until we improve the test coverage.

sborrazas:
Fixed on https://github.com/aeternity/ae_mdw/commit/3107d95286514fd433a0b8bfa2837f53e526d91d

PR #610, opened 2022-03-28
refactor: include code to fetch stats inside StatsMutation

sborrazas:
This are the initial changes that will cause mutations to behave in a
transactionl-basis, where all reads are done inside the mutation
itself.


thepiwo (approved):

jyeshe (commented):
lgtm except for the deletion of `test/integration/ae_mdw/sync/stats_test.exs`. Even though we've discussed an idea of keeping integration tests mostly for endpoints this is one that does some validations that are harder to rely on unit tests because of the accumulator behavior of totalstats. It has also helped to check if the deltastats performs the increments in the correct height.

sborrazas:
I wouldn't have labeled that test as integration, it was a unit test that was making use of the database. I think we should stick to what we've discussed, integration = endpoints exclusively. In this case, we can add a test where we check that the values for a given height are correct

jyeshe:
Or if by any other strategy, these unit tests for stats achieve the same validation of these integration tests, we could of course convert them too.

sborrazas:
I've now added integration tests for the *stats endpoints as well

jyeshe:
> I've now added integration tests for the *stats endpoints as well
@sborrazas Could we revert the delete of `test/integration/ae_mdw/sync/stats_test.exs` (for the validations not peformed by `test/ae_mdw/db/mutations/stats_mutation_test.exs`)?

sborrazas:
> > I've now added integration tests for the *stats endpoints as well
> > @sborrazas Could we revert the delete of `test/integration/ae_mdw/sync/stats_test.exs` (for the validations not peformed by `test/ae_mdw/db/mutations/stats_mutation_test.exs`)?

fetching the data is internal to the mutation (as it should for transactions to work), and there's no new_mutation or any of that sort, so perhaps you can provide a code example of a unit test of that you would like?

jyeshe:
> > > I've now added integration tests for the *stats endpoints as well
> > > @sborrazas Could we revert the delete of `test/integration/ae_mdw/sync/stats_test.exs` (for the validations not peformed by `test/ae_mdw/db/mutations/stats_mutation_test.exs`)?
> 
> fetching the data is internal to the mutation (as it should for transactions to work), and there's no new_mutation or any of that sort, so perhaps you can provide a code example of a unit test of that you would like?

Things that test/integration/ae_mdw/sync/stats_test.exs validate:
1. If neither delta stats nor total stats never become negative
2. If there is a block reward with correct value for @first_block_reward_height
3. If there is a delta stats for each of these transitions at the correct height:
```
@first_name_claim_height 194
@first_oracle_register_height 4165
@first_oracle_expired_height 5851
@first_contract_create_height 4187
@first_name_revoked_height 34_923
```
4. If total stats matches the accumulator for the heights above.
5. If total stats accumulates correct values for block and dev reward after multiple increments.

Sorry for being demanding on this but as we've had some issues with stats before so it's better to be more cautious. The last issue fixed by #609 was found thanks to these integration tests.

If we could have these validations on unit tests it would be great.

jyeshe (approved):

PR #616, opened 2022-03-30
fix: initialize aex9 balance when not exists

### LINKED ISSUE #615: aex9 balance not available
```
error 400 http://localhost:4000/aex9/balances/ct_RDRJC5EySx4TcLtGRWYrXfNgyWzEDzssThJYPd9kdLeS5ECaA
{"error":"balance is not yet available: contract ct_RDRJC5EySx4TcLtGRWYrXfNgyWzEDzssThJYPd9kdLeS5ECaA"}
```

jyeshe:
## What

* Convert method_args to proper encoding when saving balance
* Create zeroed aex9 balance for transfer destination when does not exist
* Initialize aex9 presence and balance on contract creation for both regular and child contracts

## Why

Fix for 
```
error 400 http://localhost:4000/aex9/balances/ct_RDRJC5EySx4TcLtGRWYrXfNgyWzEDzssThJYPd9kdLeS5ECaA
{"error":"balance is not yet available: contract ct_RDRJC5EySx4TcLtGRWYrXfNgyWzEDzssThJYPd9kdLeS5ECaA"}
```

Closes #615 

## Notes
* aex9 balance block_index was not used so it was removed


lib/ae_mdw/db/model.ex
```diff
@@ -223,12 +223,10 @@ defmodule AeMdw.Db.Model do
 
   # AEX9 balance:
   #     index = {contract_pk, account_pk}
-  #     block_index = {kbi, mbi}
   #     amounts: float
   @type aex9_balance ::
           record(:aex9_balance,
             index: {Db.pubkey(), Db.pubkey()},
-            block_index: Blocks.block_index(),
```
sborrazas, line 231: :+1: nice workaround to avoid having to redefine the structure, but at some point we should probably change it, perhaps we can create a GH issue?

sborrazas (approved):

lib/ae_mdw/db/model.ex
jyeshe: It was removed because it's not being used on fetch (e.g the history uses the contract balance instead).

sborrazas: I understand that, but it wasn't removed from the database, every time we store this record we are storing 3 values unnecessarily. This is what I think we should work on in the future

jyeshe: Humm, I see what you mean. This is a change that demands to start a sync from scratch.

sborrazas: If we remove it from the default without a sync from scratch it will break the existing records. We can either:

1. leave it on the default and perform the change in a migration (which we can add later on)
2. remove the default and require a sync from scratch

I'm fine either way, but perhaps we can do 1. for now to avoid syncing from scratch?

jyeshe: I will provide a migration, thanks. There is a another fix needed for contract creation, drafting ...

lib/ae_mdw/sync/async_tasks/derive_aex9_presence.ex
```diff
@@ -37,101 +35,21 @@ defmodule AeMdw.Sync.AsyncTasks.DeriveAex9Presence do
 
     Log.info("[:derive_aex9_presence] #{inspect(contract_pk)} after #{time_delta / @microsecs}s")
 
-    recipients_set = contract_pk |> get_aex9_recipients(kbi, mbi) |> MapSet.new()
```
jyeshe, line 40: on the context of contract creation, no need to track recipients anymore

sborrazas (approved):
LGTM

PR #621, opened 2022-04-04
refactor: use declarative state for executing mutations

sborrazas:
The goal of these changes is to be able to use a interchangeable state
that can perform these executions. The functional way of dealing with
any sort of state mutations allows this, and which will also allow
performing in-memory updates to the state.


lib/ae_mdw/db/sync/block.ex
```diff
@@ -54,91 +57,76 @@ defmodule AeMdw.Db.Sync.Block do
...
+    |> Enum.reduce({[], from_txi}, fn {{height, kb_hash}, {_next_height, next_kb_hash}},
+                                      {blocks_mutations, txi} ->
+      {key_block, micro_blocks} = AE.Db.get_blocks(kb_hash, next_kb_hash)
+      kb_header = :aec_blocks.to_key_header(key_block)
+      kb_model = Model.block(index: {height, -1}, hash: kb_hash, tx_index: txi)
+
+      if rem(height, @log_freq) == 0, do: Log.info("creating mutations for block at #{height}")
+
+      pending_micro_blocks =
+        if from_height == height do
```
jyeshe, line 71: `from_mbi != 0` maybe? (for semantics)

lib/ae_mdw/sync/server.ex
```diff
@@ -125,4 +152,17 @@ defmodule AeMdw.Sync.Server do
...
+
+    Producer.commit_enqueued()
+
+    if mbi != -1 do
+      Broadcaster.broadcast_micro_block(block, :mdw)
+      Broadcaster.broadcast_txs(block, :mdw)
+    end
+
+    new_state
+  end
```
jyeshe, line 167: After key block commit it would be a time also for `Broadcaster.broadcast_key_block(key_block, :mdw)`

lib/ae_mdw/sync/server.ex
```diff
@@ -108,8 +120,23 @@ defmodule AeMdw.Sync.Server do
...
+        :none -> -1
+      end
+
+    block_index = {from_height, from_mbi}
+    db_state = State.new()
+    from_txi = Block.next_txi()
+
     spawn_link(fn ->
-      Block.sync(from_height, to_height)
+      {blocks_mutations, _next_txi} = Block.blocks_mutations(block_index, from_txi, to_height)
```
jyeshe, line 134: A nitpick as there is `from_txi` and `to_height` maybe instead of `block_index` could we separate it as `from_height` and `from_mbi`?

lib/ae_mdw/sync/server.ex
```diff
@@ -108,8 +120,23 @@ defmodule AeMdw.Sync.Server do
...
+
+      _new_state =
+        Enum.reduce(blocks_mutations, db_state, fn {block_index, block, mutations}, state ->
+          commit_mutations(state, block_index, block, mutations)
+        end)
```
jyeshe, line 139: There are some blocks that take a long time to load them from node DB and dry-run the events. What you think of having commits for every height?

jyeshe (commented):
Very neat PR! :100: Only one topic to analyse an alternative. If it would be possible to have commits for every height.

lib/ae_mdw/db/origin.ex
```diff
@@ -54,6 +57,14 @@ defmodule AeMdw.Db.Origin do
     end
   end
 
+  @spec tx_index!(State.t(), creation_txi_locator()) :: Txs.txi()
+  def tx_index!(state, {:contract, pk} = creation_txi_locator) do
+    case State.cache_get(state, :ct_create_sync_cache, pk) do
+      {:ok, txi} -> txi
+      :not_found -> tx_index!(creation_txi_locator)
```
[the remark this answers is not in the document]
sborrazas: Good idea, done, added it to the cache

lib/ae_mdw/db/sync/block.ex
sborrazas: If we don't check for `from_height  == height` then we would be skipping `from_mbi` blocks for every height

lib/ae_mdw/sync/server.ex
sborrazas: Added it, thanks!

lib/ae_mdw/sync/server.ex
sborrazas: Done

lib/ae_mdw/sync/server.ex
sborrazas: That's a good idea, I'd like to add that on a following PR so we can test it thoroughly from scratch

lib/ae_mdw/db/sync/block.ex
jyeshe: At least for readability, looks like something can be improved, maybe putting both conditions:
`from_height == height and from_mbi != 0` but even after it the `from_height == height` part (from my perspective of not writing the code) would continue to be not easily readable.

sborrazas: Done

sborrazas:
These final version is now passing the integration tests + diff script. There some differences in the stats which I think is now the correct version. Ready for review new

thepiwo (approved):

thepiwo:
whats the state of this, I think we discussed not to have PRs hanging for so long generally

jyeshe (commented):
After unit tests passing, it will be approved provided that as discussed there will be a following PR that makes commits for each height.

jyeshe (approved):

Issue #396, opened 2021-12-13
name expiration consistency


jyeshe:
```
iex(aeternity@localhost)3> :mnesia.dirty_all_keys(Model.ActiveNameExpiration) |> Enum.group_by(fn {_height, name} -> name end) |> Enum.filter(fn {_name, expirations} -> length(expirations) > 1 end)
[
  {"payment.chain", [{538528, "payment.chain"}, {578734, "payment.chain"}]},
  {"9988776655.chain",
   [{632725, "9988776655.chain"}, {635550, "9988776655.chain"}]},
  {"xu.chain", [{645129, "xu.chain"}, {645130, "xu.chain"}]}
]
```
However, old expirations are always deleted on:
- update: https://github.com/aeternity/ae_mdw/blob/045ee3518e38b4cdff92f7163da6f697c905b7d2/lib/ae_mdw/db/sync/name.ex#L127
- revoke: https://github.com/aeternity/ae_mdw/blob/045ee3518e38b4cdff92f7163da6f697c905b7d2/lib/ae_mdw/db/sync/name.ex#L176
- expire: https://github.com/aeternity/ae_mdw/blob/045ee3518e38b4cdff92f7163da6f697c905b7d2/lib/ae_mdw/db/name.ex#L74 

and created single time on:
- claim (double claim not allowed): https://github.com/aeternity/ae_mdw/blob/045ee3518e38b4cdff92f7163da6f697c905b7d2/lib/ae_mdw/db/sync/name.ex#L63
- auction end: https://github.com/aeternity/ae_mdw/blob/045ee3518e38b4cdff92f7163da6f697c905b7d2/lib/ae_mdw/db/name.ex#L109

PS: transfer changes only the owner and a name cannot be revoked during an auction (SDK doesn't continue with the operation).

jyeshe:
@thepiwo @sborrazas here is there analysis that shows why we shouldn't have any name expiration issues. Summarizing, whenever a new expiration is saved and old is deleted (happen together in the same function without condition). The height of the old expiration deleted is also the very same height that was saved before (without any computation, read directly from the record).

The only possibility would be if the chain allowed in the past multiple claims (that would create duplicate expiration) or a revoke before auction end. Both are very less likely but I will add logs to confirm it.

jyeshe:
[...]

PR #622, opened 2022-04-04
test: fix integration inactive names cases by expiration/deactivation

jyeshe:
## What

* Fixes test cases that considers inactive names expiration/deactivation only by expiration height.

## Why

On v1 endpoint, [at least since April 30 2021 (v1.0.8)](https://github.com/aeternity/ae_mdw/issues/396#issuecomment-1087850775), the `InactiveNameExpiration` table/record is also used for revoked height. This is needed to know which names were deactivated lately (or firstly depending on `backward` or `forward` direction) whether the name had expired or had being revoked.

## Notes

On v2 endpoint the purpose of the parameter can be improved/clarified renaming it from `by=expiration` to ~~`by=height`~~ `by=deactivation`

### DIFF
--- a/test/integration/ae_mdw_web/controllers/name_controller_test.exs
+++ b/test/integration/ae_mdw_web/controllers/name_controller_test.exs
@@ -181,7 +181,10 @@ defmodule Integration.AeMdwWeb.NameControllerTest do
                |> json_response(200)
 
       expirations =
-        Enum.map(names, fn %{"info" => %{"expire_height" => expire_height}} -> expire_height end)
+        Enum.map(names, fn %{"info" => %{"expire_height" => expire_height, "revoke" => revoke}} ->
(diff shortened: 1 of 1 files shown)


sborrazas:
We should probably rename the parameter to be `by=deactivation` and restructure the endpoint accordingly. I would assume the user would want to see the latest deactivated names (either expired or revoked), as opposed to the latest expired names. But this is more of a feature discussion.

sborrazas (approved):

jyeshe:
> We should probably rename the parameter to be `by=deactivation` and restructure the endpoint accordingly. I would assume the user would want to see the latest deactivated names (either expired or revoked), as opposed to the latest expired names. But this is more of a feature discussion.

Exactly, it has always worked with expired and revoked. It will be a naming fix.

thepiwo (approved):

PR #623, opened 2022-04-05
refactor: rename /v2/names/* by=expiration to by=deactivation

jyeshe:
## What

* Accept by=deactivation on /v2/names/* (by=expiration will be deprecated on /v1 deprecation)
* Change documentation to clarify the use of `by` parameter explaining that the sorting considers also the revoke height.

## Why

The v1 `/names/inactive?by=expiration` had always sorted not only by expiration but also included the revoked names. This  is needed to know which names were deactivated lately (or firstly depending on backward or forward direction) whether the name had expired or had being revoked.

### DIFF
--- a/lib/ae_mdw/names.ex
+++ b/lib/ae_mdw/names.ex
@@ -50,7 +50,8 @@ defmodule AeMdw.Names do
 
   @spec fetch_names(pagination(), range(), order_by(), query(), cursor() | nil, boolean()) ::
           {:ok, cursor() | nil, [name()], cursor() | nil} | {:error, reason()}
-  def fetch_names(pagination, range, :expiration, query, cursor, expand?) do
+  def fetch_names(pagination, range, order_by, query, cursor, expand?)
+      when order_by in [:expiration, :deactivation] do
     cursor = deserialize_expiration_cursor(cursor)
     scope = deserialize_scope(range)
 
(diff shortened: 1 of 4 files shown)


thepiwo (approved):

lib/ae_mdw/names.ex
```diff
@@ -50,16 +50,16 @@ defmodule AeMdw.Names do
 
   @spec fetch_names(pagination(), range(), order_by(), query(), cursor() | nil, boolean()) ::
           {:ok, cursor() | nil, [name()], cursor() | nil} | {:error, reason()}
-  def fetch_names(pagination, range, :expiration, query, cursor, expand?) do
+  def fetch_names(pagination, range, order_by, query, cursor, expand?)
+      when order_by in [:expiration, :deactivation] do
```
sborrazas, line 54: Maybe I'm not getting this right, but right now these names are being sorted by whatever is on the `ActiveNameExpiration` or `InactiveNameExpiration` table, which just sorts them by expiration regardless of when they were revoked (deactivated). Shouldn't we use a different index for when `by=deactivation`?

jyeshe: It's the table name that is not so clear but `InactiveNameExpiration` has either an expiration or a revoke. `ActiveNameExpiration` will always have the expiration.

sborrazas (approved):
Like we've discussed with @jyeshe we should deprecate the `by=expiration` option since it's not really sorting by expiration, but by deactivation

PR #628, opened 2022-04-08
chore: add diffing script to compare two different environments

sborrazas:
Usage: `./scripts/endpoint-diff.sh URL1 URL2`

By default it will compare `https://mainnet.aeternity.io` and `http://localhost:4000`.

### DIFF
--- a/scripts/endpoint-diff.sh
+++ b/scripts/endpoint-diff.sh
@@ -6,6 +6,8 @@ URL_2="${2:-http://localhost:4000/}"
 ERROR=0
 SUCCESS=0
 
+mkdir -p /tmp
+
 echo "Diffing $URL_1 and $URL_2 ..."
 
 while read LINE; do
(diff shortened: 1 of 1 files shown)


scripts/endpoint-diff.sh
```diff
@@ -0,0 +1,31 @@
...
+while read LINE; do
+  if [[ $LINE =~ mdw\/(.*?)\" ]]; then
+    ROUTE="${BASH_REMATCH[1]}"
+
+    echo "Diffing '$ROUTE'"
+
+    curl -so '.endpoint-diff.json1.tmp' "$URL_1$ROUTE"
+    curl -so '.endpoint-diff.json2.tmp' "$URL_2$ROUTE"
+
+    if diff .endpoint-diff.json1.tmp .endpoint-diff.json2.tmp; then
```
jyeshe, line 20: Please redirect the output of the diff to some file under /tmp. I will start using it too. Thanks

thepiwo (approved):

scripts/endpoint-diff.sh
thepiwo: then we should also adjust it in the check endpoints script

sborrazas: I've redirected the output of this script to a file. I think it's better to display it on the stdout though so you can see what's failing while the script is running, though

jyeshe: Thanks

jyeshe (approved):

PR #630, opened 2022-04-11
fix: invalidate aex9 balance and update on any call

jyeshe:


lib/ae_mdw/db/model.ex
```diff
@@ -220,14 +220,16 @@ defmodule AeMdw.Db.Model do
...
+  #     block_index: {kbi, mbi},
+  #     amount: float
   @type aex9_balance ::
           record(:aex9_balance,
             index: {Db.pubkey(), Db.pubkey()},
+            block_index: {Blocks.height(), Blocks.mbi()},
             amount: float()
           )
-  @aex9_balance_defaults [index: {<<>>, <<>>}, amount: nil]
+  @aex9_balance_defaults [index: {<<>>, <<>>}, block_index: {-1, -1}, amount: nil]
```
jyeshe, line 232: needed for invalidation (mbi reserved for future use)

lib/ae_mdw/sync/async_tasks/update_aex9_presence.ex
```diff
@@ -1,6 +1,6 @@
-defmodule AeMdw.Sync.AsyncTasks.UpdateAex9Presence do
+defmodule AeMdw.Sync.AsyncTasks.UpdateAex9State do
```
jyeshe, line 1: filename not renamed with `git mv` as by the change percentage it's creating a new one.

lib/ae_mdw/db/model.ex
jyeshe: migration included for this

sborrazas (approved):

thepiwo (approved):

Issue #131, opened 2021-01-21
endpoint to receive outbid auctions of an account


marc0olo:
In our application we want to display active auctions for account where they are participating but got outbid by some other account.

Currently there is `GET  /names/owned_by/:id` which shows the current top bids and the current names owned per account.

I propose to add a separate endpoint `GET /names/outbid/:id` which returns this information.

thepiwo:
would be nice to have, but not high priority.

thepiwo:
@jyeshe @sborrazas  do you have any input on this?

sborrazas:
@thepiwo we can add a new endpoint `/v2/names/:name/outbids` for this, IIRC we already index the previous bids

marc0olo:
I think we only need that for the current auction-cycle (simple array) and not historical (array of arrays). an additional filter by account would be cool :)

thepiwo:
lets index this per auction, not per name, thus historical data is no additional work

PR #652, opened 2022-04-22
refactor: restructure AuctionBid table for better indexing

sborrazas:


thepiwo (approved):

lib/ae_mdw/db/model.ex
```diff
@@ -102,15 +102,22 @@ defmodule AeMdw.Db.Model do
...
   defrecord :auction_bid, @auction_bid_defaults
 
-  @type auction_bid_key ::
-          {String.t(), {Blocks.block_index(), Txs.txi()}, Blocks.height(), Db.pubkey(), list()}
   @type auction_bid ::
           record(:auction_bid,
-            index: auction_bid_key(),
-            unused: nil
+            index: Names.plain_name(),
+            block_index_txi: Blocks.block_index_txi(),
```
jyeshe, line 117: I think multiple bids should be allowed to a name.

jyeshe (commented):
It looks the index needs some change.

lib/ae_mdw/db/model.ex
sborrazas: The history of bids is kept on the `bids: [Blocks.block_index_txi()]` but only one AuctionBid can exist per name

sborrazas: Source: https://github.com/aeternity/ae_mdw/blob/master/lib/ae_mdw/db/mutations/name_claim_mutation.ex#L153

jyeshe: Maybe it could keep the multiple bidders. It might be useful for https://github.com/aeternity/ae_mdw/issues/131

sborrazas: That could be a useful feature, I would leave it out of the scope of this PR since it will require a bigger migration to obtain that information, though. Good thing is that the PR is already referenced by your previous comment

jyeshe (approved):
LGTM

PR #655, opened 2022-04-26
test: add aex9 tests iterating throughout all contracts

jyeshe:
## What

Increase test coverage for AEX9 calling some endpoints with all existing aex9 contracts.

## Why

Most bugs reported were from specific contracts.

## Notes

- Another PR will cover data consistency across multiple endpoints. Something like https://github.com/aeternity/ae_mdw/blob/647062bf35a55e0b23074c40f4706f020ee57b27/test/integration/ae_mdw_web/controllers/aex9_controller_test.exs#L458 that is skipped for now due to long time needed to run.
- Some endpoints does not use saved aex9 balance records yet. There will be a following PR for that.

### DIFF
--- a/test/integration/ae_mdw_web/controllers/aex9_controller_test.exs
+++ b/test/integration/ae_mdw_web/controllers/aex9_controller_test.exs
@@ -1,10 +1,13 @@
 defmodule Integration.AeMdwWeb.Aex9ControllerTest do
   use AeMdwWeb.ConnCase, async: false
 
+  alias AeMdw.Database
   alias AeMdw.Db.Origin
   alias AeMdw.Db.Model
   alias AeMdw.Db.Util
   alias AeMdw.Validate
+
+  import AeMdwWeb.Helpers.Aex9Helper
 
   require Model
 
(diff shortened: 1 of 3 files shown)


test/integration/ae_mdw_web/controllers/aex9_controller_test.exs
```diff
@@ -317,6 +452,22 @@ defmodule Integration.AeMdwWeb.Aex9ControllerTest do
...
+    test "gets balances for each account with aex9 presence", %{conn: conn} do
+      Model.Aex9AccountPresence
+      |> Database.all_keys()
+      |> Enum.map(fn {account_pk, _txi, _contract_pk} ->
+        :aeser_api_encoder.encode(:account_pubkey, account_pk)
+      end)
+      |> Enum.uniq()
+      |> Enum.each(fn account_id ->
+        conn = get(conn, "/aex9/balances/account/#{account_id}")
+        assert balances_response = json_response(conn, 200)
```
sborrazas, line 468: Can we pattern-match the response structure here to avoid an unused variable? Perhaps something like `%{} = ...` if we don't want to map specific keys

sborrazas (commented):
nice to have these for additional checks, thanks!

test/integration/ae_mdw_web/controllers/aex9_controller_test.exs
jyeshe: done

sborrazas (approved):

thepiwo (approved):

PR #656, opened 2022-04-26
chore: remove unused node and db stream code

sborrazas:


jyeshe (approved):
Invalidation of aex9 account presence does not use anymore `DBS.map`. Double checked clear path for invalidation without DBS. 

Thanks for the `chore`!

thepiwo (approved):

PR #657, opened 2022-04-28
fix: write block_index to aex9 balance

jyeshe:
Rebasing is sometimes cumbersome ...

- Fixes regression on aex9 balance after in-memory state.
- Adds test case to prevent similar case and to validate that both aex9 balance and presence were written.

PS: This requires a sync from scratch and a following PR will add the txi.

### DIFF
--- a/lib/ae_mdw/db/mutations/derive_aex9_presence_mutation.ex
+++ b/lib/ae_mdw/db/mutations/derive_aex9_presence_mutation.ex
@@ -4,6 +4,7 @@ defmodule AeMdw.Db.DeriveAex9PresenceMutation do
   """
 
   alias AeMdw.Aex9
+  alias AeMdw.Blocks
   alias AeMdw.Db.Contract
   alias AeMdw.Db.Model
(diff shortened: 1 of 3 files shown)


jyeshe:
rebased ...

thepiwo (approved):

lib/ae_mdw/db/mutations/derive_aex9_presence_mutation.ex
```diff
@@ -45,6 +49,7 @@ defmodule AeMdw.Db.DeriveAex9PresenceMutation do
       m_balance =
         Model.aex9_balance(
           index: {contract_pk, account_pk},
+          block_index: block_index,
```
sborrazas, line 52: How does this affect already existing aex9_balances? Shouldn't we have a migration for them as well?

jyeshe: It requires a full sync.

PR #658, opened 2022-04-28
refactor: use aex9 balance records on account endpoints

jyeshe:
## What

- Adds `txi` to aex9 balance
- Fixes the condition to call an `UpdateAex9State`
- Writes aex9 presence only within the async task (same database txn as the aex9 balance)
- Uses saved aex9 balance on: 
  - `/aex9/balances/account/:account_id`
  - `/aex9/balance/:contract_id/:account_id`
  - `/v2/aex9/:contract_id/balances/:account_id`

## Why

- `txi` is needed by aex9 account endpoints to show tx info;
- Avoid multiple unnecessary dry-runs while the value is already stored;
- Fixes inconsistency when there was an aex9 presence but not a balance by writing everything in the same database txn.


lib/ae_mdw/sync/async_tasks/update_aex9_presence.ex
```diff
@@ -15,27 +17,40 @@ defmodule AeMdw.Sync.AsyncTasks.UpdateAex9State do
...
   @microsecs 1_000_000
-  @min_height 500_000
 
   @spec process(args :: list()) :: :ok
   def process([contract_pk]) do
     Log.info("[update_aex9_state] #{inspect(contract_pk)} ...")
-    {_type, height, _hash} = height_hash = DBN.top_height_hash(false)
+    {{kbi, mbi} = block_index, call_txi} = get_call_bi_and_txi(contract_pk)
 
-    if height > @min_height do
```
jyeshe, line 25: removed this pre-condition once it's difficult to discriminate the reason of missing records.

thepiwo (approved):

sborrazas (approved):

Issue #638, opened 2022-04-17
How does Websocket stay long?


Lfh2msn:
WSS does not have a heartbeat mechanism, can't send ping, and receive pong. How to confirm disconnect? How to keep a long connection?

thepiwo:
@sborrazas or @jyeshe can you check?

thepiwo:
@sborrazas can we close this?

sborrazas:
Yes indeed, closing

PR #664, opened 2022-05-04
feat: add Ping operation to websocket

sborrazas:
this allows clients to check their subscriptions, plus that the
connection is still live

refs #638


thepiwo (approved):
I'm ok to include this also with 1.9.0

Issue #629, opened 2022-04-08
Plan support AEX 141 (NFT)


thepiwo:
In Review Standard proposal:
https://github.com/aeternity/AEXs/issues/141

First plan support, how do we do the integration, what endpoints will be needed? First use-case would be the explorer I'd imagine.

@marc0olo can you provide some input from your experience?
@jyeshe do you wish to support that from MDW side?

jyeshe:
Definitely yes @thepiwo !

marc0olo:
let me think about that in detail and come back to this. cannot wait to see that coming! =)

maybe @zkvonsnarkenstein also wants to provide some requirements addressing the API

I also invite @the-icarus to provide feedback here :)

marc0olo:
what I think we should have from functional perspective is:
- search for an NFT contract by name included in meta_info
- get all holders of a specific NFT contract (+ providing the ID's the holder owns)
- get all NFT contracts
- all NFTs owned by a specific account
   - for all NFT contracts
   - for a specific NFT contract
- transfer history
   - all transfers of a specific NFT contract
   - all transfers of a specific account (across all kinds of NFT contracts)
       - ideally also filterable by a specific NFT contract

the NFT contract info should contain meaningful data:
- meta_info
- metadata_type
- extensions (mintable, ...)
- total amount of NFTs
- amount of unique holders

will think about this again. but I think with these functionalities I'd be very happy already =)

ideally we wait until @arjanvaneersel comes up with the examples (reference implementation) because we faced some issues in the AEX-141 proposal. there will be some small changes.

jyeshe:
This gives a north to start implementing the basic structure on what needs to be stored and retrieved. Thanks a lot @marc0olo

thepiwo:
As discussed in our last call, it will be good not to make a proprietary support for AEX 141, as we did for AEX 9, but as the inner workings are similar on how we can index that data find a shared approach to reuse our AEX 9 infrastructure.

jyeshe:
Contract successfully deployed: NFT (missing *approve* methods)
owner : [ak_2WAgvMCvx3dxD5oz2y7UNdBDaJdpRGer3BLvrLxDNzDcG8XN3n](https://explorer.testnet.aeternity.io/account/transactions/ak_2WAgvMCvx3dxD5oz2y7UNdBDaJdpRGer3BLvrLxDNzDcG8XN3n)
[...]

PR #667, opened 2022-05-05
refactor: generalize aex9 meta info with aexn contract

jyeshe:
## What

- Moves metadata from `RevAex9Contract` to `AexnContract`
- Replace usage of `RevAex9Contract` for rendering
- Shared fetch and rendering for Aex9 and Aex141

## Why

Refs #629

## Notes

- RevAex9Contract will be completely removed after handling forks/invalidations in a generalized way for Aex9 and Aex141


lib/ae_mdw_web/websocket/socket_handler.ex
```diff
@@ -1,4 +1,6 @@
 defmodule AeWebsocket.Websocket.SocketHandler do
+  @moduledoc false
+  # credo:disable-for-this-file
```
sborrazas, line 3: If this branch is rebased, this isn't really needed anymore

test/ae_mdw/aexn_tokens_test.exs
```diff
@@ -0,0 +1,89 @@
...
+                contract_txi: ^txi,
+                contract_id: ^contract_id
+              }} = AexnTokens.fetch_token({:aex9, contract_pk})
+    end
+
+    test "returns a AEX-141 contract meta info" do
+      contract_pk = <<141_123_456::256>>
+
+      aex141_meta_info =
+        {name, symbol, base_url, type} = {"AE Boots", "Boot", Faker.Internet.url(), :url}
```
sborrazas, line 47: We're using the `Faker` library just to make up an internet URL? I would try to avoid bloating the app with libraries, even if it's just for testing purposes

sborrazas (changes requested):

test/ae_mdw/aexn_tokens_test.exs
thepiwo: :D

jyeshe: bloating with one library? (((: It is the first use but no problem, I will implement one module for that.

lib/ae_mdw_web/websocket/socket_handler.ex
jyeshe: the base branch wasn't updated, done

thepiwo (approved):

PR #670, opened 2022-05-09
refactor: replace aex9 sync cache by non deduped params

jyeshe:


thepiwo (approved):

lib/ae_mdw/sync/async_tasks/store.ex
```diff
@@ -30,21 +30,21 @@ defmodule AeMdw.Sync.AsyncTasks.Store do
...
     |> Enum.take(max_amount)
-    |> Enum.filter(fn Model.async_tasks(index: index) ->
+    |> Enum.filter(fn Model.async_task(index: index) ->
       not :ets.member(@processing_tab, index)
     end)
   end
 
-  @spec save_new(atom(), list()) :: :ok
-  def save_new(task_type, args) do
+  @spec save_new(atom(), list(), list()) :: :ok
```
sborrazas, line 42: Perhaps we can start naming these `(task_type(), args(), args())` to better understand what's going on in all of these modules? We can also include it in the model:

```elixir
record(:async_task, index: {id??(), AsyncTasksStore.task_type()}, args: AsyncTasksStore.task_args(), extra_args: ...task_args())
```

Not sure what that initial number is as well, is it an ID?

jyeshe: The param names are self-explanatory but added strict types specially to make more explicit the index.

This is the index: `index = {System.system_time(), task_type}`

PR #676, opened 2022-05-12
feat: sync latest gens in-memory for instant invalidation

sborrazas:
This is done using the new `MemStore` which updates the in-memory
database changes using the GbTree structure.


jyeshe (commented):
Looks good and the amount of unit tests helping to understand the kind of state transactions are expected.  

Looking from a different angle there were some refactoring opportunities and would add one PoC test as I have never used `persistent_term` to write big chunks of data, so depending on the number and size of blocks that the `MemStore` will keep, it would be needed to investigate if there is any limit for the `persistent_term.put` size/frequency.

With the iEx tried these two amounts of data:
1)
```
iex(1)> list = for i <- 1..100_000_000, do: i; nil
nil
iex(2)> :timer.tc(fn -> :persistent_term.put(State, %{list: list}) end)
literal_alloc: Cannot allocate 1600000088 bytes of memory (of type "literal").
```
2) 
```
iex(1)> list = for i <- 1..50_000_000, do: i; nil                      
nil
iex(2)> :timer.tc(fn -> :persistent_term.put(State, %{list: list}) end)
{411536, :ok}
```

(1) maybe fixable with vm args if needed, (2) took half of a second, doing only this write, so if the data per write/put is bigger than that or the processing in parallel affects the performance of `persistent_term` it would become a bottleneck with the time of writing taking more time than the time of producing data.

lib/ae_mdw/db/mem_store.ex
```diff
@@ -0,0 +1,116 @@
...
+  alias AeMdw.Util.GbTree
+
+  @derive AeMdw.Db.Store
+  defstruct [:fallback_store, :tables]
+
+  @typep key() :: Database.key()
+  @typep record() :: Database.record()
+  @typep table() :: Database.table()
+  @opaque t() :: %__MODULE__{
+            fallback_store: Store.t(),
```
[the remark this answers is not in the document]
jyeshe: This module is very neat and the code clear, maybe we should only analyse if time allows whether it would make more sense for the `State` to do it accessing different stores.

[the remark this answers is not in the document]
jyeshe: From one conceptual point of view a `State` would know where to look for the info (when it doesn't exist in memory) but the way it is is elegant because keeps the `MemStore` fully working as the behaviour it implements.

sborrazas:
@jyeshe I've now added the state machine implementation using GenStateMachine on https://github.com/aeternity/ae_mdw/pull/676/commits/44115da03b4e47d5007582ecfed1c4282600f7c0

lib/ae_mdw/sync/watcher.ex
```diff
@@ -1,64 +1,75 @@
...
+  def handle_cast(:start_sync, state) do
+    :aec_events.subscribe(:chain)
+    :aec_events.subscribe(:top_changed)
 
-  def handle_info({:DOWN, ref, :process, _, :normal}, %__MODULE__{sync_mon_ref: ref} = s),
-    do: %{s | sync_mon_ref: nil}
+    Server.new_height(chain_height())
 
-  def handle_info({:DOWN, ref, :process, _, reason}, %__MODULE__{sync_mon_ref: ref} = s) do
-    notify_operators(s.operators, reason)
```
jyeshe, line 27: I wonder if we could improve the notification feature rather than remove it. It might become helpful to be notified of unexpected behaviors with debugging info.

lib/ae_mdw/sync/server.ex
```diff
@@ -1,181 +1,353 @@
...
-        {:noreply, %__MODULE__{state | syncing?: false}}
-    end
+  def handle_event(
+        :info,
+        {:DOWN, _ref, :process, pid, reason},
+        {:syncing_mem, pid},
+        %__MODULE__{restarts: restarts} = state_data
+      )
+      when restarts < @max_restarts do
+    Log.info("Mem Sync.Server error: #{inspect(reason)}")
```
jyeshe, line 274: After monitoring it's better use symmetrical calls including the `demonitor`.

lib/ae_mdw/sync/server.ex
```diff
@@ -1,181 +1,353 @@
...
+  @spec handle_event(:cast, cast_event(), state(), state_data()) ::
+          :gen_statem.event_handler_result(state())
+  @spec handle_event(:call, call_event(), state(), state_data()) ::
+          :gen_statem.event_handler_result(state())
+  @spec handle_event(:info, info_event(), state(), state_data()) ::
+          :gen_statem.event_handler_result(state())
+  @spec handle_event(:internal, internal_event(), state(), state_data()) ::
+          :gen_statem.event_handler_result(state())
+  def handle_event(:cast, {:new_height, chain_height}, :initialized, state_data) do
+    actions = [{:next_event, :internal, :check_sync}]
```
jyeshe, line 125: A constant (module attribute) would be helpful to know what actions are passed in `:next_state`.

lib/ae_mdw/sync/server.ex
```diff
@@ -1,181 +1,353 @@
...
-    db_state = State.new(db_store)
+    db_state = State.new()
     mem_state = State.new(mem_store)
 
-    event_handler = EventHandler.init(chain_height(), db_state, mem_state, &spawn_with_monitor/1)
+    state_data = %__MODULE__{
+      chain_height: nil,
+      fork_height: nil,
+      db_state: db_state,
+      mem_state: mem_state,
```
jyeshe, line 107: Maybe there is a way to avoid exposing the whole database and in memory state to the State machine, having in the `state_data` only the state needed for the transitions.

jyeshe (commented):
Code readability and maintainability was improved. Some refactoring or fixes suggested on encapsulation and process monitoring.

lib/ae_mdw/sync/watcher.ex
sborrazas: I agree, but right now this functionality is of no use at all, writing files and sending emails with the `mail` command. I think we should work on an actual useful feature with a module called `ErrorNotifier` or some sort of more descriptive name

lib/ae_mdw/sync/server.ex
```diff
@@ -1,181 +1,353 @@
...
+    |> Enum.reduce(state, fn {{_height, mbi}, block, block_mutations}, state ->
+      new_state = State.commit_mem(state, block_mutations)
+
+      broadcast_block(block, mbi == -1)
+
+      new_state
+    end)
+  end
 
   defp spawn_with_monitor(fun) do
```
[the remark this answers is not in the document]
sborrazas: Good idea! Done

lib/ae_mdw/sync/server.ex
sborrazas: Done, implemented using Task.Supervisor as suggested

lib/ae_mdw/sync/server.ex
sborrazas: Not sure what you mean by this, there's a typespec for all possible states and events. Can you elaborate please?

lib/ae_mdw/sync/server.ex
sborrazas: Same here

lib/ae_mdw/sync/watcher.ex
jyeshe: okay

lib/ae_mdw/sync/server.ex
jyeshe: it's a small detail and on a personal taste for bigger values that are repeated I usually do something like:
`@check_sync_next_event [{:next_event, :internal, :check_sync}]`

lib/ae_mdw/sync/server.ex
jyeshe: The goal would be separate what is state needed to control the sync (the state machine `state_data`) from the state needed to execute the sync. But with the last refactor the use of `db_state` and `mem_state` are better encapsulated :+1: 

I would add only a comment explaining on the state machine notes that it needs only to compare the height of `db_state` and `mem_state` with the `chain_height` in order to decide what is the next transition (to the state `syncing_db` or `syncing_mem`).

jyeshe (changes requested):
Some thoughts:
- Similarly to the controllers that were updated to read from in-memory state using the recent `StatePlug`, the async tasks would also need some way to be enqueued (written and read from) in-memory state. 
- Fork handling is missing on `Watcher` (subscriber) and `Server`
- Since `Server.gens_per_min()` is called after each status controller request it would be better to move `gens_per_min` out of the syncing server state in order to avoid `:gens_per_min` messages competing with `:new_height` on `Server` mailbox.

sborrazas:
> Some thoughts:
> 
> * Similarly to the controllers that were updated to read from in-memory state using the recent `StatePlug`, the async tasks would also need some way to be enqueued (written and read from) in-memory state.

Async-tasks are not persisted in-memory because the time it takes to process them is unknown, and by the time they are done, the memory contents might have completely changed. I think this is why we made them async tasks in the first place.

> * Fork handling is missing on `Watcher` (subscriber) and `Server`

For every new key block that appears in the chain (either by a new kb added or by a fork) we re-process all of the latest generations, regardless of whether it was caused by a fork or not. This is why there isn't really a need for determining if it's a fork or not, we just handle both the same way

> * Since `Server.gens_per_min()` is called after each status controller request it would be better to move `gens_per_min` out of the syncing server state in order to avoid `:gens_per_min` messages competing with `:new_height` on `Server` mailbox.

As per OTP guidelines it's not recommended to have messages that take long to process, which is why all messages that arrive on Sync.Server are instantly processed, including `:new_height` and `:gens_per_min`.

jyeshe:
> > Some thoughts:
> > 
> > * Similarly to the controllers that were updated to read from in-memory state using the recent `StatePlug`, the async tasks would also need some way to be enqueued (written and read from) in-memory state.
> 
> Async-tasks are not persisted in-memory because the time it takes to process them is unknown, and by the time they are done, the memory contents might have completely changed. I think this is why we made them async tasks in the first place.

The purpose of the async tasks is to run them in parallel with the sync. After 32c2575 for 1.7.3 the async task are enqueued only after 8 heights. I was waiting the in memory handling to fix this delay in order to restore the previous behaviour of 1.7.2. Opening an issue though. 

> > * Fork handling is missing on `Watcher` (subscriber) and `Server`
> 
> For every new key block that appears in the chain (either by a new kb added or by a fork) we re-process all of the latest generations, regardless of whether it was caused by a fork or not. This is why there isn't really a need for determining if it's a fork or not, we just handle both the same way

What if the fork happens during the processing of the in-memory first height? Shouldn't the data the was just committed be really invalidated? (: I am giving the example of one height of difference but the fork might be even detected with higher delay (after more heights). 

> > * Since `Server.gens_per_min()` is called after each status controller request it would be better to move `gens_per_min` out of the syncing server state in order to avoid `:gens_per_min` messages competing with `:new_height` on `Server` mailbox.
> 
> As per OTP guidelines it's not recommended to have messages that take long to process, which is why all messages that arrive on Sync.Server are instantly processed, including `:new_height` and `:gens_per_min`.

It would be to avoid concurrency between multiple `/status` requests and any syncing process message as the GenServers can process only one message at a time while ETS allows multiple reads. The idea would would be to reduce a low risk of bottleneck (considering a change or a condition that makes any message slower) to none risk.

jyeshe (commented):
Note on required invalidation and optional `:gens_per_minute` improvement

jyeshe:
> Note on required invalidation and optional `:gens_per_minute` improvement

An example/history of sync and fork that can happen:
1. in memory sync 80, 81, 82, 83, 84, 85, 86, 87
2. 80 to 87 heights are committed and during the processing of 88 the chain detects that 87 doesn't belong to the main/rightful chain.

sborrazas:
@jyeshe I've now updated the gens_per_min to be stored into a global persistent_term readable from anywhere as you requested.

Regarding the fork handling, like I mentioned previously the entire in-memory storage is re-processed for every change in the chain, including forks and new blocks, so that's handled already

jyeshe (approved):
Thanks for the refactor and for clarifying that a fork cannot happen on Aeternity blockchain after 10 heights from the top. I was still thinking of:
https://cointelegraph.com.br/news/ethereum-beacon-chain-experiences-7-block-reorg-what-s-going-on

thepiwo (approved):
looks really good from what I can see

jyeshe (approved):
Discussed pending points here: "run in-memory aex9 tasks asynchronously" #749

PR #681, opened 2022-05-17
fix: revert change on prev key iteration

jyeshe:
## What

Revert 996b686e (1.9.1 change) on previous key iteration (revert the circular behavior)

## Why

Fix the intermittent error caught by:
```
1) test prev_key/2 returns :not_found when there is no previous key (AeMdw.Db.RocksDbCFTest)
Error:      test/ae_mdw/db/rocksdbcf_test.exs:149
     match (=) failed
     code:  assert :not_found = RocksDbCF.prev_key(Model.Tx, 0)
     left:  :not_found
     right: {:ok, 100000000001}
     stacktrace:
       test/ae_mdw/db/rocksdbcf_test.exs:150: (test)
```

## Notes

Adds a fix for unrelated error (due to rebase on `test/ae_mdw/db/contract_call_mutation_test.exs`)


lib/ae_mdw/db/rocksdb_cf.ex
```diff
@@ -240,7 +240,7 @@ defmodule AeMdw.Db.RocksDbCF do
           end
 
         :not_found ->
-          last_key(table)
+          :not_found
```
sborrazas, line 243: Why does `iterator_prev_key` defaults to `:not_found`, but `iterator_next_key` defaults to `first_key(tab)`? I don't think we should have any asymmetric behavior in that sense

jyeshe: an error that we haven't before

thepiwo (approved):

PR #682, opened 2022-05-17
chore: include first version of aexn into migration

jyeshe:
## What

- Includes unreleased first version of AEX-N contracts (`Model.AexNContractPubkey`) in the migration

## Why

- Continue validation without resync


lib/ae_mdw/db/model.ex
```diff
@@ -274,6 +274,15 @@ defmodule AeMdw.Db.Model do
...
   defrecord :aex9_contract_pubkey, @aex9_contract_pubkey_defaults
 
+  # AEX-N contract pubkey:
+  #     index: {type, pubkey} where type = :aex9, :aex141, ...
+  #     txi: txi
+  @aexn_contract_pubkey_defaults [
+    index: nil,
+    txi: nil
+  ]
+  defrecord :aexn_contract_pubkey, @aexn_contract_pubkey_defaults
```
sborrazas, line 284: Perhaps every time we add a new record we can include the type as well?

sborrazas (changes requested):

sborrazas (approved):

Issue #308, opened 2021-10-27
Index Contract Remote Call Transactions


thepiwo:
Currently remote contract calls are not exposed by middleware at all, we should show those from the scope of the child contract, as well as from the scope of the remote calling transaction of the parent contract.

calling remote: 
https://testnet.aeternity.io/mdw/txs/forward?contract=ct_mT1pE3evLqyYSMCMjeqPPQsFobpVvKCEZxxh1tarWbyPu49yY

remote:
https://testnet.aeternity.io/mdw/txs/forward?contract=ct_okiZvxvjrQvBuYcbB1EHkLi3bCtKthFHQVPk3GmKvei7fVrHo

remote calling transaction:
https://testnet.aeternity.io/mdw/tx/th_GDA4M16zfBqU6DFPwCakYVYtpsiuXFyotcYZCxTfBNvRvVUpp

jyeshe:
Not obtained through dry-running:
```
iex(aeternity@localhost)1> tx_rec = {:aetx, :contract_call_tx, :aect_call_tx, 172,
 {:contract_call_tx,
  {:id, :account,
   <<96, 140, 121, 106, 41, 137, 85, 113, 246, 224, 119, 235, 221, 21, 63, 40,
     31, 56, 92, 129, 110, 169, 24, 82, 164, 11, 249, 77, 249, 58, 4, 113>>},
  67,
  {:id, :contract,
   <<209, 57, 35, 175, 53, 15, 176, 243, 22, 6, 247, 156, 155, 82, 192, 167, 24,
     160, 33, 221, 104, 76, 172, 118, 9, 60, 108, 67, 140, 72, 120, 2>>}, 3,
  183440000000000, 0, 0, 25000, 1000000000,
  <<43, 17, 132, 161, 93, 161, 59, 159, 2, 160, 5, 30, 233, 11, 26, 145, 179,
    208, 144, 148, 221, 186, 208, 249, 46, 114, 46, 217, 163, 236, 102, 74, 127,
    83, 219, 136, 128, 193, 100, 93, 221, 38, 159, 0, 160, 231, 65, 213, 29, 14,
    127, 254, 227, 197, 222, 177, 15, 4, 17, 239, 43, 110, 157, 67, 123, 231,
    105, 156, 175, 103, 149, 108, 68, 139, 171, 46, 41, 5, 49>>, [],
[...]

PR #683, opened 2022-05-17
fix: remove aex9 presence for remote calls

jyeshe:
## What

- Skips writing `Aex9AccountPresence` for remote calls

## Why

- Aex9 balances are not yet computed for remote calls (shall be part of the scope of #308)

## Notes

- Additionally this PR fixes test cases used to test the root cause of the error


thepiwo (approved):

jyeshe:
rebased with master

sborrazas (approved):

PR #698, opened 2022-05-31
fix: handle update aex9 state on contract create logs

jyeshe:
## What /Why

Fix contract creation having aex9 logs.

## Validation steps

`elixir --sname aeternity@localhost -S mix test test/ae_mdw/db/contract_create_mutation_test.exs`

### DIFF
--- a/lib/ae_mdw/db/mutations/contract_create_mutation.ex
+++ b/lib/ae_mdw/db/mutations/contract_create_mutation.ex
@@ -3,25 +3,29 @@ defmodule AeMdw.Db.ContractCreateMutation do
   Processes contract_create_tx.
   """
 
+  alias AeMdw.Blocks
   alias AeMdw.Contract
   alias AeMdw.Db.Contract, as: DBContract
   alias AeMdw.Db.State
   alias AeMdw.Txs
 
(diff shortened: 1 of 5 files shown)


lib/ae_mdw/db/mutations/contract_create_mutation.ex
```diff
@@ -40,6 +45,6 @@ defmodule AeMdw.Db.ContractCreateMutation do
     state
     |> State.inc_stat(:contracts_created)
     |> State.cache_put(:ct_create_sync_cache, contract_pk, txi)
-    |> DBContract.logs_write(nil, txi, txi, call_rec)
+    |> DBContract.logs_write(block_index, txi, txi, call_rec)
```
sborrazas, line 48: Can we update the type signature for `Db.Contract.logs_write/4` not to allow `nil` as well? https://github.com/aeternity/ae_mdw/blob/master/lib/ae_mdw/db/contract.ex#L171

jyeshe: done

sborrazas (approved):

PR #700, opened 2022-06-01
fix: retrieve block hash for name ptr_resolve from state

sborrazas:


jyeshe (commented):
Could we have a unit or integration test? This way we can continue increasing the coverage for existing code without taking a long time to implement big test suites.

Issue #674, opened 2022-05-10
Improve Swagger File


thepiwo:
- [ ] Consider Referencing the swagger file of the node for datastructures as proposed by @davidyuk 
- [x] have .info.version in swagger file equal to middleware version https://github.com/aeternity/ae_mdw/issues/525
- [x] Page parameter is documented in Swagger, but not working actually https://github.com/aeternity/ae_mdw/issues/512
- [x] Wrong maximum of limit parameter of get_all_auctions method in Swagger file https://github.com/aeternity/ae_mdw/issues/511
- [ ] Different auction schema depending on if name was registered before https://github.com/aeternity/ae_mdw/issues/509
- [x] Replace names/pointees with name/pointees in Swagger file https://github.com/aeternity/ae_mdw/issues/508
- [x] /mdw/ is not defined in testnet swagger file https://github.com/aeternity/ae_mdw/issues/503
- [x] consider removing the static swagger.json from repo https://github.com/aeternity/ae_mdw/issues/189
- [x] Wrong schemes value of swagger file on production https://github.com/aeternity/ae_mdw/issues/160
- [x] Name swagger file consistent with Aeternity node  https://github.com/aeternity/ae_mdw/issues/108
- [x] add swagger route to http-endpoints section  https://github.com/aeternity/ae_mdw/issues/98

sborrazas:
Additionally, we're going to need to version the swagger file as well. As discussed, this can be done using two separate yaml files for specifying each version API, removing the need for the automated phoenix/swagger elixir library.

jyeshe:
Set the correct limit parameter for all endpoints

davidyuk:
The parts of this issue should fix ability to make requests to middleware through https://validator.swagger.io/. Currently it fails requesting something like http://testnet.aeternity.io/txs?account=ak_g5vQK6beY3vsTJHH7KBusesyzq9WMdEYorF8VyvZURXTjLnxT that misses https, `mdw` part, and direction. I'm providing https://testnet.aeternity.io/mdw/swagger/swagger.json at the top.

thepiwo:
/mdw is not part of the path that middleware provides, it's added by the SRE team in their deployment

davidyuk:
do we have a board to track SRE issues? cc: @venimus @mmpetarpeshev 

Also, as an alternative to combining multiple issues, you can create a special swagger label.

thepiwo:
[...]

PR #709, opened 2022-06-07
docs: update v2 swagger docs with latest paths

sborrazas:
AEx9 paths are still pending for specification

refs #674


jyeshe (approved):
:clap:

thepiwo (approved):

PR #710, opened 2022-06-07
feat: save and display aexn extensions

jyeshe:
refs #629

### DIFF
--- a/lib/ae_mdw/aexn_contracts.ex
+++ b/lib/ae_mdw/aexn_contracts.ex
@@ -104,6 +104,7 @@ defmodule AeMdw.AexnContracts do
   defp decode_meta_info({name, symbol, variant_url, variant_type}) do
     url =
       case variant_url do
+        {:variant, [0, 1], 1, {url}} -> url
         {:variant, [0, 1], 1, url} -> url
         _other -> nil
       end
@@ -174,4 +175,6 @@ defmodule AeMdw.AexnContracts do
       match?({_code, {[:address], :integer}, _body}, functions[@check_swap_hash]) and
       match?({_code, {[], {:map, :address, :string}}, _body}, functions[@swapped_hash])
   end
+
+  defp valid_aex141_extension?(_any, _functions), do: false
 end
--- a/lib/ae_mdw/aexn_contracts.ex
+++ b/lib/ae_mdw/aexn_contracts.ex
@@ -1,4 +1,4 @@
-defmodule AeMdw.Contracts.AexnContract do
+defmodule AeMdw.AexnContracts do
   @moduledoc """
   AEX-N detection and common calls to interact with the contract.
   """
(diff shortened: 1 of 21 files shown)


lib/ae_mdw/db/sync/contract.ex
```diff
@@ -112,21 +112,26 @@ defmodule AeMdw.Db.Sync.Contract do
...
-        contract_pk,
-        aexn_meta_info,
-        block_index,
-        txi
-      )
+         {:ok, aexn_extensions} <- AexnContracts.call_extensions(aexn_type, contract_pk),
+         {:ok, aexn_meta_info} <- AexnContracts.call_meta_info(contract_pk) do
+      if aexn_type == :aex9 or
+           (aexn_type == :aex141 and
+              AexnContracts.has_valid_aex141_extensions?(aexn_extensions, contract_pk)) do
```
sborrazas, line 125: Perfectly clear 👍

sborrazas, line 125: 👍

lib/ae_mdw/db/model.ex
```diff
@@ -308,17 +308,20 @@ defmodule AeMdw.Db.Model do
...
-            meta_info: aexn_meta_info()
+            meta_info: aexn_meta_info(),
+            extensions: aexn_extensions()
           )
   @aexn_contract_defaults [
     index: nil,
     txi: -1,
-    meta_info: nil
+    meta_info: nil,
+    extensions: []
```
sborrazas, line 324: Do we really want to store this potentially infinite amount of extensions? We just need a single contract with a million extensions to kill the entire database/server/app

sborrazas, line 324: Do we really want to store this potentially infinite amount of extensions? We just need a single contract with a million extensions to kill the entire database/server

sborrazas (changes requested):

lib/ae_mdw/aexn_contracts.ex
```diff
@@ -39,15 +39,52 @@ defmodule AeMdw.Contracts.AexnContract do
...
+        false
+    end
+  end
 
-  @spec call_meta_info(pubkey()) :: {:ok, Model.aexn_meta_info()} | :not_found
+  def has_valid_aex141_extensions?(extensions, {:fcode, functions, _hash_names, _code}) do
+    Enum.all?(extensions, &valid_aex141_extension?(&1, functions))
+  end
+
+  def has_valid_aex141_extensions?(_extensions, _no_fcode) do
```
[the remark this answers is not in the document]
jyeshe: It's a handling for AEVM that also exists for AEX9. As it's legacy for the old VM, maybe the Node doesn't return the info in this other format anymore. It should be logged and checked to be safely removed.

One option would be to add `aevm_type :: any` in the spec but it would loosen the type check. 

For now I have just returned `false` as there would be no 141 at that time.

lib/ae_mdw/db/model.ex
jyeshe: On blockchain world this is one the reasons that `gas` and `gas limit` exist. Currently the Mdw doesn't truncate any information coming from the Chain. If that is needed could describe into an issue?

jyeshe: A lower `gas limit` was used for the `aex9_extensions` and `aex141_extensions` methods. Even though it's stateless some limit still applies on dry-running.

sborrazas (approved):
LGTM, thanks for looking into the extensions exploit, we would need to merge both simulaneously

jyeshe:
> LGTM, thanks for looking into the extensions exploit, we would need to merge both simulaneously

NP, the current one already handles it so the PRs are independent. If both are approved great, if not it doesn't block the release.

PR #712, opened 2022-06-07
fix: display unencoded block hash when not found

sborrazas:
Previously the value displayed was the Elixir binary representation
(e.g. `<<210, 20, 143, ...>>`). Instead, the AE encoded hash is used
as displayed on the URL.

### DIFF
--- a/lib/ae_mdw/blocks.ex
+++ b/lib/ae_mdw/blocks.ex
@@ -77,13 +77,19 @@ defmodule AeMdw.Blocks do
 
   @spec fetch(State.t(), block_index() | block_hash()) :: {:ok, block()} | {:error, Error.t()}
   def fetch(state, block_hash) when is_binary(block_hash) do
-    case :aec_chain.get_block(block_hash) do
-      {:ok, _block} ->
-        # note: the `nil` here - for json formatting, we reuse AE node code
-        {:ok, Format.to_map(state, {:block, {nil, nil}, nil, block_hash})}
+    case Validate.id(block_hash) do
+      {:ok, encoded_hash} ->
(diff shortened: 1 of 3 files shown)


thepiwo (approved):

lib/ae_mdw/blocks.ex
```diff
@@ -76,13 +76,19 @@ defmodule AeMdw.Blocks do
...
   @spec fetch(block_index() | block_hash()) :: {:ok, block()} | {:error, Error.t()}
   def fetch(block_hash) when is_binary(block_hash) do
-    case :aec_chain.get_block(block_hash) do
-      {:ok, _block} ->
-        # note: the `nil` here - for json formatting, we reuse AE node code
-        {:ok, Format.to_map({:block, {nil, nil}, nil, block_hash})}
+    case Validate.id(block_hash) do
+      {:ok, encoded_hash} ->
+        case :aec_chain.get_block(encoded_hash) do
+          {:ok, _block} ->
```
jyeshe, line 82: Could we use a `with` instead?

test/integration/ae_mdw_web/controllers/block_controller_test.exs
```diff
@@ -278,6 +278,13 @@ defmodule Integration.AeMdwWeb.BlockControllerTest do
       assert json_response(conn, 400) ==
                %{"error" => "invalid range: #{range}"}
     end
+
+    test "when block is not found, it returns a 404 error message", %{conn: conn} do
```
jyeshe, line 282: :pray:

jyeshe (approved):

PR #724, opened 2022-06-14
feat: truncate aexn name and symbol sorting fields

### LINKED ISSUE #721: define aexn meta info limits
AEX9 and AEX141 contracts might have big name or symbol that are returned by `meta_info` method and are stored by the Mdw.

Questions for the requirements refinement:
1. If the `meta_info` dry-run demands more gas than the limit (150% of node base gas for `contract_call_tx`), should the contract be indexed? Currently without `meta_info`, the contract is not indexed as AEX9 or AEX141.
2. Even with 100% of Node `contract_call_tx` base gas, the name returned might have more than 1000 chars. Should it be truncated to a certain length before saving it? 
3. The same as above applies for the AEX-N meta info symbol. Should it be truncated before saving?

jyeshe:
Resolves #721

### DIFF
--- a/lib/ae_mdw/aexn_contracts.ex
+++ b/lib/ae_mdw/aexn_contracts.ex
@@ -125,9 +125,10 @@ defmodule AeMdw.AexnContracts do
     {name, symbol, url, metadata_type}
   end
 
-  defp error_meta_info(:aex9), do: {:error, :error, nil}
-  defp error_meta_info(:aex141), do: {:error, :error, nil, nil}
-  defp error_meta_info(_aexn_type), do: {:error, :error}
+  defp error_meta_info(:aex9), do: {:out_of_gas_error, :out_of_gas_error, nil}
+
+  defp error_meta_info(:aex141),
+    do: {:out_of_gas_error, :out_of_gas_error, :out_of_gas_error, nil}
 
   defp has_all_signatures?(aexn_signatures, functions) do
     Enum.all?(aexn_signatures, fn {hash, type} ->
--- a/lib/ae_mdw/aexn_contracts.ex
+++ b/lib/ae_mdw/aexn_contracts.ex
@@ -69,10 +69,14 @@ defmodule AeMdw.AexnContracts do
 
   def has_valid_aex141_extensions?(_extensions, _no_fcode), do: false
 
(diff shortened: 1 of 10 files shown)


lib/ae_mdw/db/contract.ex
```diff
@@ -424,4 +434,12 @@ defmodule AeMdw.Db.Contract do
       :not_found -> Model.aex9_balance(index: {contract_pk, account_pk}, amount: 0)
     end
   end
+
+  defp sort_field_truncate(field_value) do
+    if String.length(field_value) < @max_sort_field_length do
+      field_value
+    else
+      String.slice(field_value, 0, @max_sort_field_length) <> "..."
```
jyeshe, line 442: allows filtering by prefix of any length up to 100

thepiwo (commented):
will there be conflicts eventually when long similar names are truncated to the same short version?

jyeshe:
> 

No problem as both will be indexed by name and it's possible to have a cursor that can go directly to the second contract with the same name because the contract pubkey is part of the cursor.

lib/ae_mdw/aexn_contracts.ex
```diff
@@ -69,10 +69,14 @@ defmodule AeMdw.AexnContracts do
...
-    with {:ok, {:tuple, meta_info_tuple}} <- call_contract(contract_pk, "meta_info", []) do
-      {:ok, decode_meta_info(meta_info_tuple)}
+  @spec call_meta_info(Model.aexn_type(), pubkey()) :: {:ok, Model.aexn_meta_info()}
+  def call_meta_info(aexn_type, contract_pk) do
+    case call_contract(contract_pk, "meta_info", []) do
+      {:ok, {:tuple, meta_info_tuple}} ->
+        {:ok, decode_meta_info(meta_info_tuple)}
+
+      :error ->
+        {:ok, error_meta_info(aexn_type)}
```
sborrazas, line 79: I'm having a hard time understanding the return value of this function. Perhaps we can return `{:ok, aexn_meta_info()} | :error` instead of a tuple with bogus data? The tuple `{:error, :error, nil}` has no use

sborrazas (changes requested):

sborrazas (approved):

thepiwo (approved):

PR #726, opened 2022-06-16
feat: sync up to latest micro-block

sborrazas:


thepiwo (approved):

lib/ae_mdw/node/db.ex
```diff
@@ -14,12 +14,61 @@ defmodule AeMdw.Node.Db do
...
+  @spec get_blocks_per_height(Blocks.height(), Blocks.block_hash() | Blocks.height()) :: [
+          {Blocks.height(), [micro_block()], Blocks.block_hash()}
+        ]
+  def get_blocks_per_height(from_height, block_hash) when is_binary(block_hash),
+    do: get_blocks_per_height(from_height, block_hash, nil)
+
+  def get_blocks_per_height(from_height, to_height) when is_integer(to_height) do
+    {:ok, header} = :aec_chain.get_key_header_by_height(to_height + 1)
+
+    last_mb_hash = :aec_headers.prev_hash(header)
```
jyeshe, line 34: There are some "empty" generations (without microblocks).

lib/ae_mdw/sync/server.ex
```diff
@@ -163,7 +166,7 @@ defmodule AeMdw.Sync.Server do
 
     new_state_data = %__MODULE__{
       state_data
-      | mem_height: State.height(new_db_state),
+      | mem_hash: State.height(new_db_state),
```
jyeshe, line 169: Implement something like `State.height_hash` ?

lib/ae_mdw/node/db.ex
sborrazas: this is contemplated by the `get_blocks_per_height`, mbs will be an empty array

jyeshe (approved):
nice! optionally would rename or add comment for someone eventually debugging to indicate that ` it is already handling the case when `last_mb_hash` = `prev_kb_hash

PR #743, opened 2022-06-27
feat: add nft holder endpoints

jyeshe:
refs #629

### DIFF
--- a/lib/ae_mdw/aex141.ex
+++ b/lib/ae_mdw/aex141.ex
@@ -1,0 +1,110 @@
+defmodule AeMdw.Aex141 do
+  @moduledoc """
+  Returns NFT info interacting with AEX-141 contracts or from transfer history.
+  """
+
+  alias AeMdw.AexnContracts
+  alias AeMdw.Collection
+  alias AeMdw.Db.Model
+  alias AeMdw.Db.State
+  alias AeMdw.Error
+  alias AeMdw.Error.Input, as: ErrInput
+  alias AeMdw.Util
+
+  import AeMdwWeb.Helpers.AexnHelper, only: [enc_ct: 1]
+
+  require Model
+
+  @typep pagination :: Collection.direction_limit()
+  @typep cursor() :: binary()
+
+  @typep pubkey :: AeMdw.Node.Db.pubkey()
+  @type token_id :: integer()
+
+  @ownership_table Model.NftOwnership
+
+  @spec fetch_nft_owner(pubkey(), token_id()) :: {:ok, pubkey()} | {:error, Error.t()}
+  def fetch_nft_owner(contract_pk, token_id) do
+    with :ok <- validate_aex141(contract_pk),
+         {:ok, {:address, account_pk}} <-
+           AexnContracts.call_contract(contract_pk, "owner", [token_id]) do
+      {:ok, account_pk}
+    else
+      {:error, exception} ->
+        {:error, exception}
+
+      _invalid_call_return ->
+        {:error, ErrInput.ContractReturn.exception(value: enc_ct(contract_pk))}
+    end
+  end
+
+  @spec fetch_owned_nfts(State.t(), pubkey(), cursor(), pagination()) ::
+          {:ok, {cursor() | nil, [token_id()], cursor() | nil}} | {:error, Error.t()}
+  def fetch_owned_nfts(state, account_pk, cursor, pagination) do
+    case deserialize_cursor(cursor) do
+      {:ok, cursor_key} ->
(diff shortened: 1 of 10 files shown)


lib/ae_mdw/aex141.ex
```diff
@@ -0,0 +1,110 @@
...
+        {:error, ErrInput.ContractReturn.exception(value: enc_ct(contract_pk))}
+    end
+  end
+
+  @spec fetch_owned_nfts(State.t(), pubkey(), cursor(), pagination()) ::
+          {:ok, {cursor() | nil, [token_id()], cursor() | nil}} | {:error, Error.t()}
+  def fetch_owned_nfts(state, account_pk, cursor, pagination) do
+    case deserialize_cursor(cursor) do
+      {:ok, cursor_key} ->
+        {prev_cursor_key, nfts, next_cursor_key_tuple} =
```
sborrazas, line 46: Why is one called `_key` and the other one `_tuple`? I'd rather describe values by what they represent (key) instead of what they are (tuple)

lib/ae_mdw/error.ex
```diff
@@ -68,6 +72,8 @@ defmodule AeMdw.Error do
     defexception!(NotFound)
     defexception!(Expired)
     defexception!(NotAex9)
+    defexception!(NotAex141)
+    defexception!(ContractReturn)
```
sborrazas, line 76: 🙏

lib/ae_mdw_web/controllers/aex141_controller.ex
```diff
@@ -0,0 +1,45 @@
...
+
+  import AeMdwWeb.Helpers.AexnHelper
+
+  plug(PaginatedPlug)
+  action_fallback(FallbackController)
+
+  @spec nft_owner(Conn.t(), map()) :: Conn.t()
+  def nft_owner(conn, %{"contract_id" => contract_id, "token_id" => token_id}) do
+    with {:ok, contract_pk} <- Validate.id(contract_id, [:contract_pubkey]),
+         {token_id, ""} <- Integer.parse(token_id),
```
sborrazas, line 25: This is going to return `:error` which is not handled by the `FallbackController`.

I'd perhaps send the contract and token ID directly to `Aex141.fetch_nft_owner/2` function and let it return an error if any of these things are invalid

lib/ae_mdw_web/router.ex
```diff
@@ -10,7 +10,11 @@ defmodule AeMdwWeb.Router do
     {"/contracts/logs", AeMdwWeb.ContractController, :logs},
     {"/contracts/calls", AeMdwWeb.ContractController, :calls},
     {"/totalstats/", AeMdwWeb.StatsController, :total_stats},
-    {"/status", AeMdwWeb.UtilController, :status}
+    {"/status", AeMdwWeb.UtilController, :status},
+    {"/aex141", AeMdwWeb.AexnTokenController, :aex141_tokens},
+    {"/aex141/:contract_id", AeMdwWeb.AexnTokenController, :aex141_token},
+    {"/aex141/:contract_id/owner/:token_id", AeMdwWeb.Aex141Controller, :nft_owner},
+    {"/aex141/owned_nfts/:account_id", AeMdwWeb.Aex141Controller, :owned_nfts}
```
sborrazas, line 17: Like with `account-balances` down below, I think we should make all of our routes use `-` instead of `_` which is the friendly URL way of formatting it

sborrazas (changes requested):

lib/ae_mdw/aex141.ex
jyeshe: the tuple one is the cursor used by `PaginatedPlug` with the `is_reversed?` check

lib/ae_mdw_web/controllers/aex141_controller.ex
jyeshe: Done here in the controller because `call_contract` also returns `:error`

lib/ae_mdw/aex141.ex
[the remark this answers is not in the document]
jyeshe: It's true, it was a misleading by this:
https://github.com/aeternity/ae_mdw/blob/ec2095d8c6108aba075fa0dd28f3bcf2c8f92c1f/lib/ae_mdw/collection.ex#L44

lib/ae_mdw_web/controllers/aex141_controller.ex
jyeshe: This way endpoint params validation can be done also closer to border (entry door)

sborrazas (approved):

PR #759, opened 2022-06-30
chore: present aexn tokens using state from the StatePlug

sborrazas:


jyeshe (changes requested):
Glad unit tests coverage was helpful (:

jyeshe:
> @jyeshe I've now fixed the tests that broke with that last commit.
> 
> Out of curiosity, what is it that you found useful about coverage for this PR?

Two benefits when increased or at least kept at the same percentage implementing unit tests for new features:

1. If you receive a review request after one hour of the last commit and there are unit tests that have failed you don't need to waste time reviewing code that needs to be changed.
2. Unit test might have detected errors that a code review by reading or partial manual tests wouldn't.

thepiwo (approved):

PR #779, opened 2022-07-08
chore: add aex9 validation to v1 hash endpoints

### LINKED ISSUE #780: Exception on v1 aex9 blockhash endpoints when contract is not present yet
Internal server error on https://mainnet.aeternity.io/mdw/aex9/balances/hash/kh_NM2cxdzg6mf4KMFMXw1kAzBJGwFoqiGHQtaKx3DvaAGM5CAkn/ct_RDRJC5EySx4TcLtGRWYrXfNgyWzEDzssThJYPd9kdLeS5ECaA

https://mainnet.aeternity.io/mdw/aex9/balance/hash/kh_NM2cxdzg6mf4KMFMXw1kAzBJGwFoqiGHQtaKx3DvaAGM5CAkn/ct_RDRJC5EySx4TcLtGRWYrXfNgyWzEDzssThJYPd9kdLeS5ECaA/ak_Yc8Lr64xGiBJfm2Jo8RQpR1gwTY8KMqqXk8oWiVC9esG8ce48

Same `:not_present` error when looking for on state tree as #771

jyeshe:
Resolves #780

### DIFF
--- a/lib/ae_mdw_web/controllers/aex9_controller.ex
+++ b/lib/ae_mdw_web/controllers/aex9_controller.ex
@@ -79,23 +79,26 @@ defmodule AeMdwWeb.Aex9Controller do
         )
 
   @spec balance_for_hash(Plug.Conn.t(), map()) :: Plug.Conn.t()
-  def balance_for_hash(conn, %{
+  def balance_for_hash(%Conn{assigns: %{state: state}} = conn, %{
         "blockhash" => block_hash_enc,
         "contract_id" => contract_id,
         "account_id" => account_id
-      }),
-      do:
-        handle_input(
-          conn,
-          fn ->
-            balance_for_hash_reply(
-              conn,
-              ensure_aex9_contract_pk!(contract_id),
-              Validate.id!(account_id, [:account_pubkey]),
-              ensure_block_hash_and_height!(block_hash_enc)
-            )
-          end
-        )
+      }) do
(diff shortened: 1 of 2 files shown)


thepiwo (approved):

lib/ae_mdw_web/controllers/aex9_controller.ex
```diff
@@ -79,23 +79,26 @@ defmodule AeMdwWeb.Aex9Controller do
...
+      }) do
+    with {:ok, {type, height, hash}} <- ensure_block_hash(block_hash_enc),
+         {:ok, contract_pk} <- ensure_aex9_contract_at_block(state, contract_id, hash) do
+      handle_input(
+        conn,
+        fn ->
+          balance_for_hash_reply(
+            conn,
+            contract_pk,
+            Validate.id!(account_id, [:account_pubkey]),
```
sborrazas, line 95: Perhaps we can start using `{:ok, account_pk} <- Validate.id(....) do` at the top to stop dealing with both error return values and exception errors?

It would be nice if we only dealt with just one of them, in particular, I think exceptions are very much error prone, so I would like to continue with the functional approach of returning an error tuple. Thoughts?

jyeshe: I think the exceptions help with deeper propagation but for these validations close to the border I like the idea of refactoring incrementally to use tagged errors.

sborrazas (approved):

Issue #764, opened 2022-07-01
Expose transaction hash everywhere where txi is currently exposed


thepiwo:

sborrazas:
On #789 we added the parameter `tx_hash=true` to transform tx indexes into tx hashes, but for the v3 version this should start being the default

thepiwo:
lets keep this open as a reminder for v3

subhod-i:
this is important for debuggings as txi is generally not exposed to the public. it would be better to know the transaction id and see what is wrong.

PR #789, opened 2022-07-15
feat: display tx hash instead of txi when tx_hash=true

sborrazas:
refs #764

### DIFF
--- a/lib/ae_mdw/names.ex
+++ b/lib/ae_mdw/names.ex
@@ -47,7 +47,7 @@ defmodule AeMdw.Names do
   @table_inactive_expiration Model.InactiveNameExpiration
   @table_inactive_owner Model.InactiveNameOwner
 
-  @pagination_params ~w(limit cursor rev direction scope)
+  @pagination_params ~w(limit cursor rev direction scope tx_hash expand)
   @states ~w(active inactive)
   @all_lifecycles ~w(active inactive auction)a
 
@@ -342,10 +342,10 @@ defmodule AeMdw.Names do
     %{
       active_from: active,
       expire_height: expire,
-      claims: Enum.map(claims, &expand_txi(state, Format.bi_txi_txi(&1), opts)),
-      updates: Enum.map(updates, &expand_txi(state, Format.bi_txi_txi(&1), opts)),
-      transfers: Enum.map(transfers, &expand_txi(state, Format.bi_txi_txi(&1), opts)),
-      revoke: (revoke && expand_txi(state, Format.bi_txi_txi(revoke), opts)) || nil,
+      claims: Enum.map(claims, &expand_txi(state, &1, opts)),
+      updates: Enum.map(updates, &expand_txi(state, &1, opts)),
+      transfers: Enum.map(transfers, &expand_txi(state, &1, opts)),
+      revoke: (revoke && expand_txi(state, revoke, opts)) || nil,
       auction_timeout: auction_timeout,
       pointers: render_pointers(state, name),
       ownership: render_ownership(state, name)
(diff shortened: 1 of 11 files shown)


lib/ae_mdw/names.ex
```diff
@@ -408,10 +408,17 @@ defmodule AeMdw.Names do
...
-    else
-      bi_txi
+    txi = Format.bi_txi_txi(bi_txi)
+
+    cond do
+      Keyword.get(opts, :expand?, false) ->
+        Txs.fetch!(state, txi)
+
+      Keyword.get(opts, :tx_hash?, false) ->
+        Enc.encode(:tx_hash, Txs.txi_to_hash(state, txi))
```
jyeshe, line 418: It would be good for the user to get some kind of exception like [this](https://github.com/aeternity/ae_mdw/blob/98e180459cbfcbd2a69a1b9fc17367e9f4f350af/lib/ae_mdw/names.ex#L98) otherwise the param gets ignored silently. This would indicate that it should be used one param or the other.

lib/ae_mdw/oracles.ex
```diff
@@ -210,10 +211,15 @@ defmodule AeMdw.Oracles do
   defp expand_txi(state, bi_txi, opts) do
     txi = Format.bi_txi_txi(bi_txi)
 
-    if Keyword.get(opts, :expand?, false) do
-      Txs.fetch!(state, txi)
-    else
-      txi
+    cond do
+      Keyword.get(opts, :expand?, false) ->
```
jyeshe, line 215: same as on `names.ex`

test/integration/ae_mdw_web/controllers/name_controller_test.exs
```diff
@@ -849,6 +849,27 @@ defmodule Integration.AeMdwWeb.NameControllerTest do
...
+                                   "info" => %{
+                                     "claims" => claims,
+                                     "updates" => updates,
+                                     "transfers" => transfers
+                                   }
+                                 } ->
+               Enum.all?(claims ++ updates ++ transfers, fn tx_hash ->
+                 match?(
+                   {:ok, _encoded_tx_hash},
+                   :aeser_api_encoder.safe_decode(:tx_hash, tx_hash)
```
jyeshe, line 868: It could be checked that besides present the `tx_hash` is correct, here on integration or on unit tests. 

An integration test for oracle would be missing or if the `tx_hash` becomes validated on oracle unit test it would be enough.

jyeshe (changes requested):

thepiwo (approved):

lib/ae_mdw/names.ex
sborrazas: Done, added this validation on the `PaginationPlug` so we didn't have to on every top module

test/integration/ae_mdw_web/controllers/name_controller_test.exs
sborrazas: Done, added oracle test which validates this now

jyeshe (commented):
Could it test if the hash is correct for each oracle (having it's own `tx_hash`)?

thepiwo (approved):

jyeshe:
It would be to test if the hash belongs to the correct oracle.

jyeshe:
> > Could it test if the hash is correct for each oracle (having it's own `tx_hash`)?
> 
> I think it's enough to test that at least one tx hash was converted, testing that the conversion txi->hash is not always the same seems to be too far fetch for values that were chosen randomly (8989-> th_5aUc7Xqk3TkG38XVjCzZPAWc1geGFfjGcgEytcZAFe3FJyn4g)

It would be to test if the hash belongs to the correct oracle.

jyeshe (approved):
For a complete test, it would be to test if the hash belongs to the correct oracle.

PR #792, opened 2022-07-18
chore: raise detailed message when deleting txn missing key

jyeshe:
## What

Check if key exists before deleting on db transaction 

## Why

Raise detailed message when key is missing.

## Notes

The missing key being deleted was from a record not indexed by 1.11

### DIFF
--- a/lib/ae_mdw/database.ex
+++ b/lib/ae_mdw/database.ex
@@ -155,6 +155,10 @@ defmodule AeMdw.Database do
 
   @spec delete(transaction(), table(), key()) :: :ok
   def delete(txn, table, key) do
+    if not RocksDbCF.exists?(txn, table, key) do
+      raise "Txn delete on missing key: #{table}, #{inspect(key)}"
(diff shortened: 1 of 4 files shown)


sborrazas:
@jyeshe how does this error get triggered otherwise? What's being deleted twice in the same transaction?

jyeshe:
> @jyeshe how does this error get triggered otherwise? What's being deleted twice in the same transaction?

Some operation on testnet around 600k

thepiwo (approved):

sborrazas (approved):
This will add a database read for every deletion being performed, which will impact syncing performance.

We should remove this check once we find out what the cause is

jyeshe:
> This will add a database read for every deletion being performed, which will impact syncing performance.
> 
> We should remove this check once we find out what the cause is

Most of the transactions don't delete data (: Even if it would for each transaction, RocksDB reads like a bullet for the amount of txs.

Issue #749, opened 2022-06-29
Run in-memory aex9 tasks asynchronously


jyeshe:
The #730 enables the in-memory aex9 balance update. 

It's performed synchronously (with parallel tasks) and with a timeout of 1 second per task (that involves a contract dry-run).

Pending cases to be handled having the instantaneous AEX9 sync requirement in mind:
- Process update for contracts with bigger balances (currently discarded and as a reference `balances` contract dry-run with thousand entries takes more than 1 minute)
- Process multiple calls to same contract only once if already enqueued (dedup tasks)
- Handle many aex9 balance updates avoiding the result of other transactions (like `:spend_tx` for AE transfer) to be available with a delay. An example of delay lower limit would be 10 seconds for 80+ balance updates taking 1 second each on a machine with 8 threads.

jyeshe:
* Differentiate on balance endpoints when it is not yet available (pending dry-run) from not existing ones.

jyeshe:
From #818 notes:
- Keep the sorting while enqueuing with State
- Allow enqueuing tasks to rerun if it's in processing state
- Avoid rerunning the tasks during in-memory sync by detecting invalidations outside of the task
- After 3, cache next hashes in a way to avoid walking through the blocks repeatedly for every task call.

thepiwo:
I feel this get way less important when we will be moving more towards tracking events, any thoughts @jyeshe

jyeshe:
> I feel this get way less important when we will be moving more towards tracking events, any thoughts @jyeshe 

Yes, except for the contract init that is still needed to be async to avoid blocking the txs sync.

thepiwo:
@jyeshe agree, but usually after init there is not much load on dry-running (except for discussed attack vectors) as init is gas limited, so not high priority

also referencing #969 here

jyeshe:
> @jyeshe agree, but usually after init there is not much load on dry-running (except for discussed attack vectors) as init is gas limited, so not high priority
> 
> also referencing #969 here

On AEX-9 the need for the feature continues the same by any mdw solution to balances. I would like to close this as the event based solution will coexist on at least on one release.

thepiwo:
done except for the events adaptation and the total_pending counter that are covered by #969 and #869

(copy pasted from @jyeshe chat)

PR #793, opened 2022-07-18
feat: add store to serve async tasks

jyeshe:
## What

* Add in-memory `Store` that can be used by (multiple) async tasks
* Change AEX9 tasks to use this `Store`
* Handles :chain_too_short condition when updating aex9 state for the last microblocks

## Why

Shared in-memory among multiple processes (tasks) in order to allow writing to the same `State`.
Refs #749

### DIFF
--- a/lib/ae_mdw/db/async_store.ex
+++ b/lib/ae_mdw/db/async_store.ex
@@ -1,0 +1,89 @@
+defmodule AeMdw.Db.AsyncStore do
+  @moduledoc """
+  Implementation of Store protocol to be used by asynchronous tasks.
+
+  Its operations have immediate in-memory effect (not persisted) and
+  is backed by a cache with TTL of 1 day.
+  """
+
+  alias AeMdw.Database
+  alias AeMdw.EtsCache
+
+  @derive AeMdw.Db.Store
+  defstruct tid: nil
+
+  @table_id :single_async_store
+  @ttl_minutes 24 * 60
+
+  @typep key() :: Database.key()
+  @typep record() :: Database.record()
+  @typep table() :: Database.table()
+  @opaque t() :: %__MODULE__{tid: :single_async_store}
+
+  @spec init() :: :ok
+  def init do
+    EtsCache.new(@table_id, @ttl_minutes, :ordered_set)
+    :ok
+  end
+
+  @spec instance() :: t()
+  def instance do
+    %__MODULE__{tid: @table_id}
+  end
+
+  @spec put(t(), table(), record()) :: t()
+  def put(%__MODULE__{tid: tid} = store, table, record) do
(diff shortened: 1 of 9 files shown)


jyeshe:
rebasing to master (separating from resource-busy protection)...

lib/ae_mdw/db/async_store.ex
```diff
@@ -0,0 +1,89 @@
...
+  @derive AeMdw.Db.Store
+  defstruct tid: nil
+
+  @table_id :single_async_store
+  @ttl_minutes 24 * 60
+
+  @typep key() :: Database.key()
+  @typep record() :: Database.record()
+  @typep table() :: Database.table()
+  @opaque t() :: %__MODULE__{tid: :single_async_store}
```
sborrazas, line 21: Can we use the same constant we defined before here (`@table_id`)? to avoid repeating constants that are supposed to match with each other

lib/ae_mdw/db/async_store.ex
```diff
@@ -0,0 +1,89 @@
...
+
+  Its operations have immediate in-memory effect (not persisted) and
+  is backed by a cache with TTL of 1 day.
+  """
+
+  alias AeMdw.Database
+  alias AeMdw.EtsCache
+
+  @derive AeMdw.Db.Store
+  defstruct tid: nil
```
sborrazas, line 13: `:tid` is never nil, or a possible valid value defined by the spec, so we should use `defstruct [:tid]` here instead

lib/ae_mdw/db/async_store.ex
```diff
@@ -0,0 +1,89 @@
...
+  @spec count_keys(t(), table()) :: non_neg_integer()
+  def count_keys(%__MODULE__{tid: tid}, table) do
+    do_count_keys(tid, table, nil, 0)
+  end
+
+  defp do_count_keys(tid, table, key, count) do
+    case EtsCache.next(tid, {table, key}) do
+      {^table, next_key} -> do_count_keys(tid, table, next_key, count + 1)
+      _other_table_or_nil -> count
+    end
```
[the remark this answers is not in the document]
jyeshe: yes, equivalent

lib/ae_mdw/db/async_store.ex
jyeshe: when a value is not specified, it assigns nil (same as `defstruct [:tid]`)

lib/ae_mdw/db/async_store.ex
jyeshe: it's due the compilation phase (it processes types before).

thepiwo (approved):

sborrazas (approved):

Issue #752, opened 2022-06-29
Expose transaction count by type and time period (generations)


sborrazas:
refs #124 

We could re-use the existing `/v2/txs/count` endpoint, but include different filters for it:
* `/v2/txs/count?type=spend_tx`
* `/v2/txs/count?id=ak_foobar`
* `/v2/txs/count?scope=gen:100-200`

This requires a new index to be created for types to be indexed.

sborrazas:
Done indeed

PR #798, opened 2022-07-20
feat: allow filtering transactions count by scope

sborrazas:
refs #752

### DIFF
--- a/lib/ae_mdw/txs.ex
+++ b/lib/ae_mdw/txs.ex
@@ -47,6 +47,22 @@ defmodule AeMdw.Txs do
 
   @type_spend_tx "SpendTx"
 
+  @spec count(state(), range(), map()) :: {:ok, non_neg_integer()} | {:error, Error.t()}
+  def count(_state, nil, %{"type" => _tx_type}),
+    do: {:error, ErrInput.Query.exception(value: "counting by type not yet supported")}
+
+  def count(state, nil, params) when map_size(params) == 0,
+    do: {:ok, DbUtil.last_txi!(state)}
+
+  def count(_state, {:txi, first_txi..last_txi}, params) when map_size(params) == 0,
+    do: {:ok, last_txi - first_txi + 1}
+
+  def count(state, {:gen, first_gen..last_gen}, params) when map_size(params) == 0,
+    do: {:ok, DbUtil.gen_to_txi(state, last_gen + 1) - DbUtil.gen_to_txi(state, first_gen)}
+
+  def count(_state, _range, _params),
+    do: {:error, ErrInput.Query.exception(value: "can't query by multiple values")}
+
   @spec fetch_txs(
           State.t(),
           pagination(),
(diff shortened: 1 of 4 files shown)


lib/ae_mdw/txs.ex
```diff
@@ -47,6 +47,22 @@ defmodule AeMdw.Txs do
...
+    do: {:ok, DbUtil.last_txi!(state)}
+
+  def count(_state, {:txi, first_txi..last_txi}, params) when map_size(params) == 0,
+    do: {:ok, last_txi - first_txi + 1}
+
+  def count(state, {:gen, first_gen..last_gen}, params) when map_size(params) == 0,
+    do: {:ok, DbUtil.gen_to_txi(state, last_gen + 1) - DbUtil.gen_to_txi(state, first_gen)}
+
+  def count(_state, _range, _params),
+    do: {:error, ErrInput.Query.exception(value: "can't query by multiple values")}
```
jyeshe, line 64: It matched with unknown params like ?gen=1-100

test/ae_mdw_web/controllers/tx_controller_test.exs
```diff
@@ -25,4 +30,36 @@ defmodule AeMdwWeb.TxControllerTest do
...
+               |> json_response(200)
+    end
+
+    test "it returns the difference between first and last txi", %{conn: conn} do
+      first_txi = 600
+      last_txi = 500
+
+      assert 101 =
+               conn
+               |> get("/txs/count", scope: "txi:#{first_txi}-#{last_txi}")
```
jyeshe, line 52: We could start using a shorthand for scopes like:
`http://localhost:4000/txs/count?gen=5:40` instead of (`scope=gen:5-40`) 

What you think?

jyeshe (commented):
Nice the pagination already handles decreasing range. Only one usability suggestion.

lib/ae_mdw/txs.ex
sborrazas: You're right, changing the error message

test/ae_mdw_web/controllers/tx_controller_test.exs
sborrazas: I would rather not introduce two different ways of doing the same thing, this adds complexity to the API. I also think scope filtering (either txi or gen) should be done with a single parameter, this makes it clear to the user what scoping means and makes it impossible to filter by two different things (e.g. what happens when `?gen=0-10&txi=20-30`)

We should also revisit this matter once we get rid of txis from the API entirely

thepiwo (approved):

test/ae_mdw_web/controllers/tx_controller_test.exs
jyeshe: I think it would be good for the user to allow also the shorthand but it would be an improvement. When both params are used there would be an error handling similar to the one that there was on line 64:
     `do: {:error, ErrInput.Query.exception(value: "can't query by multiple values")}`

jyeshe (approved):
Approving with optional shorthand improvement.

PR #805, opened 2022-07-26
feat: imports hardforks preset accounts

### LINKED ISSUE #347: Expose internal transfers from migration phases
in migration phases 0, 1, 2 and 3 balance was added in hard forks for token migration from erc 20 on ethereum to aeternity native tokens, the token balance changes are listed in json files, maybe also exposed by the node in some way.

https://github.com/aeternity/aeternity/tree/master/data/aecore then for each hard fork the respective accounts(_uat).json

My best guess would be to index them as internal transfers

jyeshe:
Refs #347


jyeshe:
README notes to be added...

lib/ae_mdw/db/hardfork_presets.ex
```diff
@@ -0,0 +1,54 @@
...
+  alias AeMdw.Db.IntTransfersMutation
+  alias AeMdw.Db.Model
+  alias AeMdw.Db.State
+
+  @doc """
+  Imports hardfork migrated accounts.
+  """
+  @spec import_account_presets() :: :ok
+  def import_account_presets do
+    case Database.next_key(Model.KindIntTransferTx, {"accounts_lima", {-1, -1}, <<>>, -1}) do
```
sborrazas, line 15: Can we turn this entirely into a mutation? That way we continue encapsulating Database calls through the State

sborrazas (changes requested):

lib/ae_mdw/db/hardfork_presets.ex
jyeshe: The `Database` is used because the goal is to check if the data is already persisted.

sborrazas: I understand that's **what** it does. I was suggesting changing **how** it does it. It's optional anyway, feel free to merge it

sborrazas (approved):

lib/ae_mdw/db/hardfork_presets.ex
jyeshe: I am talking about `how` not to encapsulate things to make the purpose clearer.

thepiwo (approved):

PR #807, opened 2022-07-27
refactor: process in-memory async tasks asynchronously

sborrazas:
This will keep the in-memory state from loading the AEx9 data.


sborrazas:
closing since it's going to be done differently

Issue #808, opened 2022-07-27
Expose in-memory data instantly in websockets


thepiwo:
type "mdw" should be sent as soon as the data would be available to query using the http api

user flow is
1. receive "mdw" type websocket event
2. optionally query http api for more data regarding event

thepiwo:
@sborrazas this should be done, right?

PR #809, opened 2022-07-27
fix: broadcast in-memory blocks

sborrazas:
This state check will avoid sending the same mb/kb multiple times, which happens when re-syncing in-memory gens.

refs #808

### DIFF
--- a/lib/ae_mdw_web/websocket/broadcaster.ex
+++ b/lib/ae_mdw_web/websocket/broadcaster.ex
@@ -4,11 +4,13 @@ defmodule AeMdwWeb.Websocket.Broadcaster do
   """
   use GenServer
 
+  defstruct [:dispatched_hashes]
+
   require Ex2ms
 
   @dialyzer {:no_return, broadcast: 2}
 
-  @no_state %{}
+  @max_hashes_size 10_000
   @subs_target_channels :subs_target_channels
 
   @spec start_link(any()) :: GenServer.on_start()
@@ -16,7 +18,7 @@ defmodule AeMdwWeb.Websocket.Broadcaster do
 
   @impl GenServer
   def init(:ok) do
-    {:ok, @no_state}
+    {:ok, %__MODULE__{dispatched_hashes: :queue.new()}}
   end
 
   @spec broadcast_key_block(tuple(), :node | :mdw) :: :ok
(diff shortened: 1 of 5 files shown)


thepiwo (approved):

lib/ae_mdw_web/websocket/broadcaster.ex
```diff
@@ -232,4 +249,18 @@ defmodule AeMdwWeb.Websocket.Broadcaster do
...
   end
+
+  defp already_processed?(%__MODULE__{dispatched_hashes: dispatched_hashes}, type_hash) do
+    :queue.member(type_hash, dispatched_hashes)
+  end
+
+  defp push_hash(%__MODULE__{dispatched_hashes: dispatched_hashes}, type_hash) do
+    queue = :queue.in(type_hash, dispatched_hashes)
+
+    if length(:queue.to_list(queue)) > @max_hashes_size do
```
jyeshe, line 260: Could you increase this number please? I counting I am doing is 10 generations (max of sync `Server`), 20 microblocks and more than 50 transactions each.

jyeshe: Or maybe use some kind of TTL to remove the older ones.

sborrazas: Moved to ETS for the TTL expiring. thanks!

lib/ae_mdw_web/websocket/broadcaster.ex
```diff
@@ -4,19 +4,23 @@ defmodule AeMdwWeb.Websocket.Broadcaster do
...
+  @expiration_minutes 120
   @subs_target_channels :subs_target_channels
 
   @spec start_link(any()) :: GenServer.on_start()
   def start_link(_arg), do: GenServer.start_link(__MODULE__, :ok, name: __MODULE__)
 
   @impl GenServer
   def init(:ok) do
-    {:ok, @no_state}
+    EtsCache.new(@hashes_table, @expiration_minutes)
```
jyeshe, line 22: One last thing, I would move the creation of this named ETS to outside the GenServer init in case it restarts.

sborrazas: Done

jyeshe (approved):

PR #818, opened 2022-08-02
feat: synchronize async tasks write

jyeshe:
## What

- Runs aex9 tasks asynchronously
- Purges aex9 in-memory and persisted records belonging to invalidated blocks
- Synchronizes the writing of async task result with the main sync database transaction so that if the task finishes later it writes directly to the database.
- Avoid rerunning the tasks for on-db sync (use previous in-memory task result to write to database)

## Why

Refs #749 

## Validation steps

- test/ae_mdw/sync/async_tasks/update_aex9_state_test.exs 
  - validates writing to async (in-memory) store
- test/ae_mdw/sync/async_tasks/store_test.exs
  - validates not to insert a task when it had been added before (if added to on-memory doesn't do for on-db sync)
- test/ae_mdw/db/state_test.exs 
  - validates the synchronization to write to memory or to disk
  - validates that if task was enqueued by on-memory sync it does not run again during on-db sync 
  - validates that commit to database uses previously run task results 
  - validates that pending tasks are only persisted during commit to database
- test/ae_mdw/db/sync/async_tasks/producer_consumer_test.exs
  - validates enqueuing and dequeuing tasks
- test/ae_mdw/db/sync/async_tasks/producer_test.exs
  - validates that pending tasks are persisted after saving in case of server restart

## Notes

Next tasks:
1. Keep the sorting while enqueuing with State
2. Allow enqueuing tasks to rerun if it's in processing state
3. Avoid rerunning the tasks during in-memory sync by detecting invalidations outside of the task
4. After 3, cache next hashes in a way to avoid walking through the blocks repeatedly for every task call.

### DIFF
--- a/.iex.exs
+++ b/.iex.exs
@@ -1,6 +1,7 @@
 alias AeMdw.Db.Model
 alias AeMdw.Db.Name
 alias AeMdw.Db.Util
+alias AeMdw.Db.State
 
 alias AeMdw.Contract
 alias AeMdw.Database
--- a/lib/ae_mdw/db/contract.ex
+++ b/lib/ae_mdw/db/contract.ex
@@ -66,11 +66,17 @@ defmodule AeMdw.Db.Contract do
   @spec aex9_write_presence(state(), pubkey(), integer(), pubkey()) :: state()
   def aex9_write_presence(state, contract_pk, txi, account_pk) do
     m_acc_presence = Model.aex9_account_presence(index: {account_pk, contract_pk}, txi: txi)
-    m_idx_presence = Model.idx_aex9_account_presence(index: {txi, account_pk, contract_pk})
 
-    state
-    |> State.put(Model.Aex9AccountPresence, m_acc_presence)
-    |> State.put(Model.IdxAex9AccountPresence, m_idx_presence)
+    State.put(state, Model.Aex9AccountPresence, m_acc_presence)
+  end
+
+  @spec aex9_delete_presence(state(), pubkey(), pubkey()) :: state()
+  def aex9_delete_presence(state, account_pk, contract_pk) do
+    if State.exists?(state, Model.Aex9AccountPresence, {account_pk, contract_pk}) do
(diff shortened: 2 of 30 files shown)


lib/ae_mdw/db/contract.ex
```diff
@@ -66,11 +66,17 @@ defmodule AeMdw.Db.Contract do
   @spec aex9_write_presence(state(), pubkey(), integer(), pubkey()) :: state()
   def aex9_write_presence(state, contract_pk, txi, account_pk) do
     m_acc_presence = Model.aex9_account_presence(index: {account_pk, contract_pk}, txi: txi)
-    m_idx_presence = Model.idx_aex9_account_presence(index: {txi, account_pk, contract_pk})
 
-    state
-    |> State.put(Model.Aex9AccountPresence, m_acc_presence)
-    |> State.put(Model.IdxAex9AccountPresence, m_idx_presence)
```
jyeshe, line 73: IdxAex9AccountPresence to be removed from database later to allow reusing previous snapshots without faking the table.

.iex.exs
```diff
@@ -1,6 +1,7 @@
 alias AeMdw.Db.Model
 alias AeMdw.Db.Name
 alias AeMdw.Db.Util
+alias AeMdw.Db.State
```
sborrazas, line 4: 🙏

lib/ae_mdw/sync/async_store_server.ex
```diff
@@ -0,0 +1,94 @@
...
+  end
+
+  @impl GenServer
+  def handle_call(
+        {:write_mutations, {kbi, _mbi} = block_index, mutations},
+        _from,
+        %{last_db_kbi: last_db_kbi} = state
+      ) do
+    # IO.inspect {kbi, Database.exists?(Model.Block, block_index)}, label: :write_mutations
+    # IO.inspect {kbi, last_db_kbi}
```
sborrazas, line 48: Debug comments

sborrazas (approved):

thepiwo (approved):

PR #833, opened 2022-08-09
refactor: decrease consumer async server wait and sleep

jyeshe:
Decreases the number of soft real time preemption on `AsyncStoreServer` by delegating the signaling of async task done.

PR #835, opened 2022-08-10
fix: handle dry-run error when contract is not present

jyeshe:
## What

- Replaces unnecessary use of banging `:aect_state_tree.get_contract(pubkey, contracts)`
- Discards the async task when dry-run fails due to `:contract_does_not_exist` or other error like insufficient funds. 

## Why

A contract may not exist anymore on a microblock referenced by block_index, which might happen syncing the last microblocks.


thepiwo (approved):

jyeshe:
rebased with master after #832 and #833 merge

sborrazas (approved):

PR #836, opened 2022-08-11
fix: increase long tasks throughput

jyeshe:
## What

Put all tasks into the same category/queue, not separating into long or regular anymore.

## Why

Mainnet has plenty of long aex9 contract calls.

## Note

There might be a PR to refactor long tasks processing in order to avoid significant delays for short tasks (with fast contract call) when all consumers are busy with long tasks.


sborrazas (approved):
🙏

thepiwo (approved):

PR #840, opened 2022-08-14
feat: log open/closed channels together with their locked AE

sborrazas:


thepiwo (commented):
lgtm, does it need a resync as I don't see a migration. Let's wait until release with the merge

lib/ae_mdw/channels.ex
```diff
@@ -0,0 +1,52 @@
...
+  def close_mutation(tx_type, tx), do: ChannelCloseMutation.new(tx_type, tx)
+
+  @spec open_mutation(Node.aetx()) :: ChannelOpenMutation.t()
+  def open_mutation(tx), do: ChannelOpenMutation.new(tx)
+
+  @spec deposit_mutation(Node.aetx()) :: ChannelSpendMutation.t()
+  def deposit_mutation(tx), do: ChannelSpendMutation.new(:aesc_deposit_tx.amount(tx))
+
+  @spec withdraw_mutation(Node.aetx()) :: ChannelSpendMutation.t()
+  def withdraw_mutation(tx), do: ChannelSpendMutation.new(-:aesc_withdraw_tx.amount(tx))
```
jyeshe, line 33: It's neat but personally I would call the `new` directly avoiding the indirection as they are used only once.

jyeshe (approved):
Looks good. Would only add unit tests for the new mutations and refactor one minor point.

sborrazas:
@thepiwo I've now added the migration which to me took ~30min to run. This migration will avoid having to perform a resync

lib/ae_mdw/channels.ex
sborrazas: The purpose of this was to encapsulate the :aesc module inside the `Channels` module so that the `Transaction` module didn't have to deal with them directly

thepiwo (approved):

PR #857, opened 2022-08-22
test: skip creating a store on integration tests

sborrazas:


thepiwo (approved):
why have the numbers changed?

sborrazas:
These two PRs #826 and #819 fix the fact that before them these numbers were calculated the wrong way once the server was restarted. It then propagated these wrong calculations in the future generations and made these numbers non-deterministic.

PR #861, opened 2022-08-22
fix: filter contracts after account balance dry-run on blockhash

jyeshe:
Contracts filtering needs to be performed after the dry-run since the presence is updated with last call.

Balances are consistent with expected 10 integration entries.

### DIFF
--- a/lib/ae_mdw_web/controllers/aex9_controller.ex
+++ b/lib/ae_mdw_web/controllers/aex9_controller.ex
@@ -315,8 +315,8 @@ defmodule AeMdwWeb.Aex9Controller do
           {:ok, account_balance} ->
             account_balance
 
-          {:error, unavailable_error} ->
-            raise unavailable_error
+          {:error, reason} ->
+            {:error, ErrInput.Aex9BalanceNotAvailable.exception(value: reason)}
         end
       else
         case Aex9.fetch_amount_and_keyblock(state, contract_pk, account_pk) do
@@ -384,7 +384,7 @@ defmodule AeMdwWeb.Aex9Controller do
          account_pk,
          {kbi, mbi} = block_index
        ) do
-    {:ok, Model.block(hash: block_hash)} = State.get(state, Model.Block, block_index)
+    Model.block(hash: block_hash) = State.fetch!(state, Model.Block, block_index)
 
     type = if mbi == -1, do: :key, else: :micro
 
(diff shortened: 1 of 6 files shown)


lib/ae_mdw_web/controllers/aex9_controller.ex
```diff
@@ -310,7 +311,13 @@ defmodule AeMdwWeb.Aex9Controller do
   defp balance_reply(%Conn{assigns: %{state: state, opts: opts}} = conn, contract_pk, account_pk) do
     {amount, {type, height, hash}} =
       if top?(opts) do
-        DBN.aex9_balance(contract_pk, account_pk, top?(opts))
+        case DBN.aex9_balance(contract_pk, account_pk, top?(opts)) do
+          {:ok, account_balance} ->
+            account_balance
+
+          {:error, unavailable_error} ->
```
sborrazas, line 318: Can we handle all errors in a friendly manner to users? Following the `Runner.call_error()` type we could do something like
```elixir
{:error, :contract_does_not_exist} -> {:error, ErrInput.Aex9BalanceNotAvailable.exception(value: contract_pk)}
{:error, :dry_run_error} -> {:error, ErrInput.Aex9BalanceNotAvailable.exception(value: "dry run error")}
{:error, reason} -> {:error, ErrInput.Aex9BalanceNotAvailable.exception(value: reason)}
```

lib/ae_mdw_web/controllers/aex9_controller.ex
```diff
@@ -377,17 +384,20 @@ defmodule AeMdwWeb.Aex9Controller do
          account_pk,
          {kbi, mbi} = block_index
        ) do
-    {:ok, Model.block(tx_index: block_txi, hash: block_hash)} =
-      State.get(state, Model.Block, block_index)
+    {:ok, Model.block(hash: block_hash)} = State.get(state, Model.Block, block_index)
```
sborrazas, line 387: Can we use `State.fetch!/3` here?

thepiwo (approved):

lib/ae_mdw_web/controllers/aex9_controller.ex
jyeshe: I think exceptions shouldn't be used without raising them but this is sth that needs to be chored throughout the code. Added for a prefixed user message.

lib/ae_mdw_web/controllers/aex9_controller.ex
jyeshe: The `block_index` is validated before but changed.

sborrazas (approved):

thepiwo (approved):

PR #865, opened 2022-08-24
fix: remove rocksdb wrapping code that created DB inconsistencies

sborrazas:
This will allow deleting records created inside the transaction and
non-existent values.


jyeshe (approved):
Looks good to me at this moment to get rid of the write-through and read-through Mnesia-like behaviours.

PR #876, opened 2022-08-29
chore: add micro_blocks to /v2/blocks/{height}

### LINKED ISSUE #873: Microblock ordering
https://mainnet.aeternity.io/mdw/blocks/28437

exposes the microblocks as object, thus the ordering coming from https://mainnet.aeternity.io/v3/generations/height/28437 is lost
can we either add some index attribute or think of another way to enable a consumer to restore the correct order implicitly (without iterating prev)

jyeshe:
## What

Add `micro_blocks` to response when `/v2/blocks` is requested not for a block but for a generation.

## Why

Closes #873

### DIFF
--- a/lib/ae_mdw_web/controllers/block_controller.ex
+++ b/lib/ae_mdw_web/controllers/block_controller.ex
@@ -4,6 +4,7 @@ defmodule AeMdwWeb.BlockController do
   alias AeMdw.Blocks
   alias AeMdw.Validate
   alias AeMdw.Util
+  alias AeMdw.Error.Input, as: ErrInput
   alias AeMdwWeb.FallbackController
   alias AeMdwWeb.Plugs.PaginatedPlug
   alias AeMdwWeb.Util, as: WebUtil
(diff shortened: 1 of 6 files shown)


sborrazas (commented):
Why are we returning different structures for `/v2/blocks/:kbi` and `/v2/blocks/:kb_hash`? We should probably return always the same structure in all 3 endpoints

jyeshe:
> Why are we returning different structures for `/v2/blocks/:kbi` and `/v2/blocks/:kb_hash`? We should probably return always the same structure in all 3 endpoints

Thank you for asking. I imagine when you wrote these endpoints the idea was to replace `v1` `/blocki/:kbi` and `/block/:hash` with a RESTful approach. 

There are some nuances that might justify the current response format:
1. `/v2/blocks/:kb_hash` or `/v2/blocks/:mb_hash` would always return a single block as requested (not a list and the microblock doesn't have nested microblock);
2. `/v2/blocks/:kbi` differently from `/v2/blocks` never returns multiple entries, only a single generation (with its `micro_blocks` added by this PR);
3. `/v2/blocks` is the one with the responsibility to return a list of generations.

For the goal of a REST api I think they are fine. Maybe would only differentiate on v3 `/v2/blocks/:hash` from `/v2/blocks/:kbi`with a separate endpoints because the former is for querying a single block while the latter is for generations (like `/v2/blocks`).

lib/ae_mdw_web/controllers/block_controller.ex
```diff
@@ -18,7 +19,27 @@ defmodule AeMdwWeb.BlockController do
   Endpoint for block info by hash or kbi.
   """
   @spec block(Conn.t(), map()) :: Conn.t()
-  def block(%Conn{assigns: %{state: state}} = conn, %{"hash_or_kbi" => hash_or_kbi} = params) do
+  def block(%Conn{assigns: %{state: state}} = conn, %{"hash_or_kbi" => hash_or_kbi}) do
+    with {:ok, kbi} <- WebUtil.parse_gen(state, hash_or_kbi),
+         {_prev_cursor, [block | _], _next_cursor} <-
```
sborrazas, line 24: Perhaps we can avoid using State functions in the controller and leave them up to the contexts

That way we can also avoid checking for it twice

```elixir
case Validate.non_neg(hash_or_kbi) do
  {:ok, kbi} -> 
    case Blocks.fetch_blocks(kbi..kbi, 1,..) do
      {nil, [block], nil} ->
        json(conn, block)
      {nil, [], nil} ->
        {:error, ErrInput.NotFound.exception(value: hash_or_kbi)}
    end
  {:error, _reason} -> 
    with {:ok, block} <- Blocks.fetch(state, hash_or_kbi) do
      json(conn, block)
    end
 end
```

sborrazas (commented):
I believe we should also change this behavior on v3, having the same endpoint return different structures is un-intuitive and dangerous. The current node API does this quite well by using different endpoints for key blocks and micro blocks https://api-docs.aeternity.io/

lib/ae_mdw_web/controllers/block_controller.ex
jyeshe: I think with `State.exists?(state, Model.Block, {gen, -1})` it makes more explicit the validation of the input but I don't mind changing to this similar code.

sborrazas (approved):
Thank you 🙏

thepiwo (approved):

PR #878, opened 2022-08-29
fix: convert transfer event token_id to integer

jyeshe:


priv/migrations/20220829223500_aex141_int_token_id.ex
```diff
@@ -0,0 +1,38 @@
...
+    updated_count =
+      Model.NftOwnership
+      |> Database.all_keys()
+      |> Enum.map(fn
+        {owner_pk, contract_pk, <<token_id::256>>} = key ->
+          Database.dirty_delete(Model.NftOwnership, key)
+
+          Database.dirty_write(
+            Model.NftOwnership,
+            Model.nft_ownership(index: {owner_pk, contract_pk, token_id})
```
sborrazas, line 24: If this migration is stopped half-way it will run into an error when running it the next time. We should make use of the `WriteMutation` and `DeleteKeysMutation` and execute them all at the end atomically

Rest looks good 👍

sborrazas (changes requested):

priv/migrations/20220829223500_aex141_int_token_id.ex
jyeshe: I was not needed. Inverting the operations makes it idempotent even with lowest probability of a delete without a write.

sborrazas (approved):

thepiwo (approved):

PR #882, opened 2022-08-31
feat: generalize transfer history for aex141

jyeshe:
## What

- Adds transfer history endpoints for nft contracts (AEX-141)
- Indexes and reads transfers from generic Aexn* records

## Why

Refs #629 

## Notes

- Includes a migration to move aex9 transfers to generic ones.
- The aex9 transfers tables and its contents will be dropped on a following release.

### DIFF
--- a/lib/ae_mdw/aex9.ex
+++ b/lib/ae_mdw/aex9.ex
@@ -78,38 +78,6 @@ defmodule AeMdw.Aex9 do
     end
   end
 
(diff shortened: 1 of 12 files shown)


thepiwo (approved):

jyeshe:
additional note: The second pubkey and the value (amount or token_id) might be moved out of the index of the record (to a second field on the record). It will be done after indexing the transfers by contracts.

jyeshe:
rebased with new Dockerfile

sborrazas (approved):

PR #897, opened 2022-09-08
feat: create nft ownership based on Mint event

jyeshe:
Refs #629


marc0olo:
@jyeshe you're already preparing for upcoming changes, right? just wondering :)

jyeshe:
> @jyeshe you're already preparing for upcoming changes, right? just wondering :)

yes Marco, it's related to previpus PR so better get it done. Transfers continue to be independent from minting (not required before transfer). I've left the validations up to the contracts for both mint and transfer operations.

jyeshe:
Later we could have a feature related to contract verification that flags it if it doesn't comply to the standard.

thepiwo (approved):

PR #899, opened 2022-09-12
feat: add nft collection stats

jyeshe:
Refs #629

### DIFF
--- a/lib/ae_mdw/db/sync/stats.ex
+++ b/lib/ae_mdw/db/sync/stats.ex
@@ -1,0 +1,62 @@
+defmodule AeMdw.Db.Sync.Stats do
+  @moduledoc """
+  Update general and per contract stats during the syncing process.
+  """
+
+  alias AeMdw.Db.Model
+  alias AeMdw.Db.State
+  alias AeMdw.Stats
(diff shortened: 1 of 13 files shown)


jyeshe:
Ready with migration.

lib/ae_mdw/db/sync/stats.ex
```diff
@@ -0,0 +1,62 @@
...
+        update_stat_counter(state, Stats.nft_owners_count_key(contract_pk), fn count ->
+          max(count - 1, 0)
+        end)
+    end
+  end
+
+  defp increment_collection_owners(state, contract_pk, to_pk) do
+    case State.next(state, Model.NftOwnerToken, {contract_pk, to_pk, -Util.max_256bit_int()}) do
+      {:ok, {^contract_pk, ^to_pk, _token}} -> state
+      _new_owner -> update_stat_counter(state, Stats.nft_owners_count_key(contract_pk))
```
sborrazas, line 49: Since the `index` for `Model.Stat` is no longer `atom()`, can we update the type in the Model module?

jyeshe: Done

Issue #888, opened 2022-09-01
Improve unpaginated micro-blocks performance in /v2/blocks


thepiwo:

sborrazas:
Taking into account the info the explorer displays we should return the following info in these new endpoints (pseudo-JSON):

### GET /v3/blocks (paginated)
```
{
  "data": [
    {
      height: 1,
      hash: ..
      beneficiary: ...
      pow: ...
      microblocks_count: 23,
      transactions_count: 1234
  ]
}
```

### GET /v3/blocks/1
```
{
  hash: ..
  beneficiary: ...
  pow: ...
  microblocks_count: 23,
  transactions_count: 1234
}
```

### GET /v3/blocks/1/microblocks (paginated)
```
{
  "data": [
    {
      height: 1,
      mbi: 0,
      hash: ..
      beneficiary: ...
      pow: ...
      transactions_count: 57
  ]
}
```

### GET /v3/blocks/1/microblocks/0
```
{
  height: 1,
  mbi: 0,
  hash: ..
  beneficiary: ...
  pow: ...
  transactions_count: 57
}
```

Thoughts?

thepiwo:
generally sounds good, lets check the overlap with core provided apis to not duplicate them to much

sborrazas:
> 

It would basically end up being what the node returns + the counts. These counts would be the added value provided by the mdw

sborrazas:
Following the [Node API Docs](https://api-docs.aeternity.io/) naming we should implement the following endpoints:

* `GET /generations`
* `GET /generations/:kbi_or_hash`
* `GET /generations/:kbi_or_hash/micro-blocks`
* `GET /generations/:kbi_or_hash/micro-blocks/:mbi_or_hash`
* `GET /generations/:kbi_or_hash/micro-blocks/:mbi_or_hash/txs`

Optionally, we can also include
* `GET /micro-blocks/:hash`
* `GET /micro-blocks/:hash/transactions`

When it exists, each of these resources should contain the counts for the nested resources (e.g. each key-block should contain the counts for micro-blocks and transactions).

thepiwo:
@sborrazas why not just `/key-blocks` and `/micro-blocks`

sborrazas:
@thepiwo not following, what would these two return? all of the key blocks and micro blocks paginated? How do you get a specific key block micro blocks?

thepiwo:
[...]

PR #900, opened 2022-09-12
feat: add /v2/micro-blocks/:hash/txs endpoint

sborrazas:
refs #888

### DIFF
--- a/README.md
+++ b/README.md
@@ -1509,7 +1509,7 @@ $ curl -s "https://mainnet.aeternity.io/mdw/v2/key-blocks?limit=1" | jq '.'
 ### `/v2/key-blocks/:hash/micro-blocks`
 
 ```
-$ curl https://mainnet.aeternity.io/key-blocks/kh_2HvzkfTvRjfwbim8YZ2q2ETKLhuYK125JGpisr1Cc9m2VSa5iC/micro-blocks
+$ curl https://mainnet.aeternity.io/key-blocks/kh_2HvzkfTvRjfwbim8YZ2q2ETKLhuYK125JGpisr1Cc9m2VSa5iC/micro-blocks?limit=1
 {
   "data": [
     {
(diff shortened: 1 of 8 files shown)


lib/ae_mdw/db/util.ex
```diff
@@ -107,6 +112,55 @@ defmodule AeMdw.Db.Util do
...
+  end
+
+  @spec micro_block_height_index(state(), binary()) ::
+          {:ok, Blocks.height(), Blocks.mbi()} | {:error, Error.t()}
+  def micro_block_height_index(state, mb_hash) do
+    with {:ok, height, decoded_hash} <- extract_height_hash(state, :micro, mb_hash) do
+      mbi =
+        decoded_hash
+        |> Db.get_micro_blocks()
+        |> Enum.count()
```
jyeshe, line 144: Same reasoning as https://github.com/aeternity/ae_mdw/pull/898#discussion_r969051347

The proposed change is to add this to `AeMdw.Node.Db`:
``` 
  def micro_block_index(hash), do: micro_block_counter(hash, -1)

  defp micro_block_counter(hash, acc) do
    with block <- :aec_db.get_block(hash),
         :micro <- :aec_blocks.type(block) do
      micro_block_counter(:aec_blocks.prev_hash(block), acc + 1)
    else
      :key -> acc
    end
  end
  ```
and use `micro_block_index(hash)` instead of calling `Db.get_micro_blocks() |> Enum.count()`

Similar to:
```
defp micro_block_walker(hash) do
    with block <- :aec_db.get_block(hash),
         :micro <- :aec_blocks.type(block) do
      {block, :aec_blocks.prev_hash(block)}
    else
      :key -> nil
    end
  end
  ```
  but without accumulating all previous micro blocks and transactions.

README.md
```diff
@@ -1552,6 +1552,38 @@ $ curl https://mainnet.aeternity.io/mdw/v2/micro-blocks/mh_HqJKqWdJ1vaPcr82zYNue
...
+  "data": [
+    {
+      "block_hash": "mh_3TzzPsMhgnJBYAtSJ6c4SdbQppZi64mxP61b1u1E8g3stDQwk",
+      "block_height": 14085,
+      "hash": "th_2Eo84A8gYkaNnRXkkEe9gPg5jcbKGdPVkZvK9XUSEQhDD6kmqm",
+      "micro_index": 59,
+      "micro_time": 1545877257605,
+      "signatures": [
+        "sg_E6tbrssPGL4a1mXyN5EW9d3UwRfYN9pSsBDtDEQVyQqTQjhQVBPKNJV6qyc43M5zY2tLE8VQa8Jb3q1XGYKJYaHM5Q3T4"
+      ],
```
jyeshe, line 1569: The fields related to the block could be moved to an upper level to avoid repeating them for all transactions.

test/ae_mdw_web/controllers/tx_controller_test.exs
```diff
@@ -102,4 +105,67 @@ defmodule AeMdwWeb.TxControllerTest do
...
+         ]}
+      ] do
+        assert %{"data" => [tx2, tx1]} =
+                 conn
+                 |> with_store(store)
+                 |> get("/v2/micro-blocks/#{encoded_mb_hash}/txs")
+                 |> json_response(200)
+
+        assert %{"a" => 1} = tx1
+        assert %{"b" => 2} = tx2
```
jyeshe, line 149: Could we have some real fields to validate the rendering at least for the format and values that are not serialized with Node's `serialize_for_client`?

jyeshe (changes requested):
Looking neat for me. Some proposals on performance, response structure and testing.

README.md
thepiwo: if this is how the core node exposes the data, then I suggest to keep it as it is

thepiwo (approved):

README.md
jyeshe: okay, similar to here: https://api-docs.aeternity.io/#/external/GetMicroBlockTransactionsByHash

could you share the reason behind replicating a similar node's structure? it's not 100% the same.

lib/ae_mdw/db/util.ex
sborrazas: Discussed in previous PR

test/ae_mdw_web/controllers/tx_controller_test.exs
sborrazas: Ended up adding integration tests because I had to much so many node functions that the test became pointless

jyeshe (approved):

Issue #902, opened 2022-09-13
use `contract` param on all endpoints that filter by contract 


jyeshe:
Some endpoints use `contract_id` others `contract`

jyeshe:
> @jyeshe can this be closed?

For v3 also Phillip. Added the tags, thanks!

PR #903, opened 2022-09-13
chore: accept contract param besides contract_id

jyeshe:
## What / Why
Accept `contract` param besides `contract_id` on nft transfer filter because some endpoints work or are documented with `contract` (while contract logs and calls expect `contract_id`).

Refs #629 

## Notes
- Changes also `/contract/calls` and `/contract/logs` to accept `contract`
- #902 was opened to discuss unifying these params on `/v3`

### DIFF
--- a/lib/ae_mdw/aexn_transfers.ex
+++ b/lib/ae_mdw/aexn_transfers.ex
@@ -60,25 +60,6 @@ defmodule AeMdw.AexnTransfers do
     )
   end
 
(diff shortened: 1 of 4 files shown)


thepiwo (approved):

sborrazas (approved):
Like we discussed today, it'd be nice if for the next version we got rid of the `contract_id` parameter entirely

PR #905, opened 2022-09-15
refactor: allocate smaller tuples for query streams

jyeshe:
## What

- Unify scope deserialization by separating key boundary from the gen or txi scope
- Remove use of `max_blob` for all `/v2/contract/calls` and `/v2/contract/logs` except for data log filtered one
- Remove use of `max_blob` for name and auction prefixing 
- Leverage Erlang term comparison to allocate less memory on query key boundaries
- Adds unit test coverage for `/v2/contract/calls` and `/v2/contract/logs`

## Why

- Allocate less memory by replacing `max_256bit_int` with `nil`
- max_blob is a 256KB

## Notes

- Gladly the BEAM doesn't reallocate every time the `max_blob` 
- Analysis of NFT token id with negative values will be done in a separte PR

### DIFF
--- a/lib/ae_mdw/util.ex
+++ b/lib/ae_mdw/util.ex
@@ -6,6 +6,8 @@ defmodule AeMdw.Util do
   alias AeMdw.Collection
   alias AeMdw.Db.State
 
+  @max_name_bin String.duplicate("z", 128)
+
   @type opt() :: {:expand?, boolean()} | {:top?, boolean()}
   @type opts() :: [opt()]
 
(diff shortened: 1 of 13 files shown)


jyeshe:
Adding a test case for contract calls ...

lib/ae_mdw/util.ex
```diff
@@ -188,22 +190,26 @@ defmodule AeMdw.Util do
...
 
   @spec min_bin() :: binary()
   def min_bin(), do: <<>>
 
+  # minimum small integer from https://www.erlang.org/doc/efficiency_guide/advanced.html
+  @spec min_int() :: integer()
+  def min_int(), do: -576_460_752_303_423_488
+
+  @spec min_256bit_int() :: integer()
+  def min_256bit_int(), do: -0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
```
sborrazas, line 211: Not sure why we're using either `min_int()` or `min_256bit_int()`, but we should probably always use just one, I can't find a reason why we wouldn't

jyeshe: `min_256bit_int` was left for one case that needs to be investigated when NFT token id is negative.

thepiwo (approved):

Issue #885, opened 2022-08-31
Update node version to 6.6+


sborrazas:
This will require updating erlang/OTP together with elixir and the necessary dependencies

thepiwo:
we also have #200 maybe we can try to update to latest elixir?

thepiwo:
I don't see any otp23 elixir images, how would we handle that? https://hub.docker.com/_/elixir

sborrazas:
Docker image [`elixir:1.11`](https://github.com/erlef/docker-elixir/blob/master/1.11/Dockerfile) uses [`erlang:23`](https://github.com/erlang/docker-erlang-otp/blob/master/23/Dockerfile) so we should probably go with that one.

feel free to close either GH issue, since it's a duplicate

thepiwo:
lets keep it as two separate ones, just to keep in mind its distinct things but probably closed by the same PR

thepiwo:
can be closed, right?

sborrazas:
Yes

Issue #200, opened 2021-06-29
update dependencies and elixir version


thepiwo:

thepiwo:
@sborrazas @jyeshe any thoughts on this, discussed with @jyeshe he maybe wants to take this in the future

jyeshe:
Closing after #918 (demanding Elixir 1.12 for Phoenix 1.6). The docker image improvement will be handled by #195

PR #909, opened 2022-09-21
chore: bump erlang to OTP 23 and elixir to 1.11

jyeshe:
## What

- Use `elixir:1.11.4` image that is built on top of `erlang:23`
- Bump riverside to work with OTP 23
- Cleanup github actions

## Why 

- Start updating elixir deps
- Prepare Mdw to work with Node v6.6+
- Refs #200 #885


thepiwo (approved):
let's not merge it, maybe make 2 releases one with current changes from master and then one specific with only this

sborrazas (approved):

PR #912, opened 2022-09-23
chore: remove phoenix_swagger

jyeshe:
## What

- Removes phoenix_swagger lib

## Why

- Latest riverside requires poison 5.0 and phoenix_swagger up to 3.0.
- phoenix_swagger is not used anymore to generate swagger files, only to render the Swagger UI


jyeshe:
@thepiwo @sborrazas a PR for the phoenix_swagger was submitted. If preferred for some reason to avoid the static assets there is also this option:
https://github.com/aeternity/ae_mdw/blob/e81a2b622028b4311537400b722576237dba4a64/mix.exs#L143 (to be removed if  the static assets are merged)

jyeshe:
Drafting to solve existing issue of: https://mainnet.aeternity.io/mdw/swagger

jyeshe:
> Drafting to solve existing issue of: https://mainnet.aeternity.io/mdw/swagger

Prefixed the `/mdw` when not present on the forwarding path (currently applies to dev, test and prod and after reorganizing the prod we might remove from dev and test).

thepiwo (commented):
do we really need to serve the swagger UI? as long as we expose the definitions any hosted UI can be used right? Maybe we can include an example for that in the docs

jyeshe:
> do we really need to serve the swagger UI? as long as we expose the definitions any hosted UI can be used right? Maybe we can include an example for that in the docs

Done (cc @sborrazas)

thepiwo (approved):

sborrazas (approved):

PR #923, opened 2022-09-28
chore: improve dialyzer warnings to catch unmatched results

sborrazas:
Also fixes some minor bug regarding validating aex9 ranges on the aex9 `balance_range` endpoint which was found when enabling these warnings

### DIFF
--- a/test/ae_mdw_web/controllers/aex9_controller_test.exs
+++ b/test/ae_mdw_web/controllers/aex9_controller_test.exs
@@ -104,6 +104,20 @@ defmodule AeMdwWeb.Aex9ControllerTest do
     end
   end
 
+  describe "balance_range" do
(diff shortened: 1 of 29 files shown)


jyeshe (approved):

test/ae_mdw_web/controllers/aex9_controller_test.exs
```diff
@@ -104,6 +104,20 @@ defmodule AeMdwWeb.Aex9ControllerTest do
     end
   end
 
+  describe "balance_range" do
+    test "it validates the range", %{conn: conn} do
```
jyeshe, line 108: Thanks for the test case.

jyeshe (approved):

thepiwo (approved):

PR #926, opened 2022-09-29
fix: return proper error when aex141 token is a partial int

sborrazas:
Since `Integer.parse("123abc")` returns `{123, "abc"}` it was causing owner path to return 500 error.


marc0olo:
how can that happen? according to the standard only `int` should be allowed. maybe I don't understand the problem 🤔

sborrazas:
> how can that happen? according to the standard only `int` should be allowed. maybe I don't understand the problem 🤔

it simply handles an invalid token gracefully, instead of returning a 500 it returns a 400+ 

What it does right now:
https://mainnet.aeternity.io/mdw/v2/aex141/ct_azbNZ1XrPjXfqBqbAh1ffLNTQ1sbnuUDFvJrXjYz7JQA1saQ3/owner/123abc

marc0olo:
it's interesting. that shouldn't be a valid token at all and ideally not even indexed then 🤔

marc0olo:
question is if we can detect that on contract level already to not include invalid contracts in the list of aex-141

marc0olo:
got it, thanks for clarification :)

thepiwo (approved):

jyeshe (approved):

PR #927, opened 2022-09-30
fix: add missing origin to oracle created by internal call

jyeshe:
## What/Why

- Add `Oracle.register_mutations` to register an oracle after internal call the same way after a `:oracle_register_tx`
- Refactors Oracle to separate methods that are used only for specific syncing purposes
- Add unit tests

## Notes

- Existing oracle response handling (and test case) would need an extra review to analyse if the query exception should be raised/propagated
- This origin was saved from the beginning for oracles with `:oracle_register_tx` and can be used for listing created oracles at a certain height
- Oracle endpoint serialization fails on 1.20 (before `:jason` upgrade)

### DIFF
--- a/lib/ae_mdw/db/mutations/oracle_extend_mutation.ex
+++ b/lib/ae_mdw/db/mutations/oracle_extend_mutation.ex
@@ -6,14 +6,20 @@ defmodule AeMdw.Db.OracleExtendMutation do
   alias :aeser_api_encoder, as: Enc
   alias AeMdw.Blocks
   alias AeMdw.Db.Model
-  alias AeMdw.Db.Oracle
   alias AeMdw.Db.State
   alias AeMdw.Log
   alias AeMdw.Node.Db
   alias AeMdw.Txs
 
+  import AeMdw.Db.Oracle, only: [cache_through_read: 3]
+
+  import AeMdw.Db.Sync.Oracle,
+    only: [
+      cache_through_delete: 3,
+      cache_through_write: 3
+    ]
+
   require Model
-  require Logger
 
   @derive AeMdw.Db.Mutation
   defstruct [:block_index, :txi, :oracle_pk, :delta_ttl]
@@ -45,7 +51,7 @@ defmodule AeMdw.Db.OracleExtendMutation do
         },
         state
       ) do
-    case Oracle.cache_through_read(state, Model.ActiveOracle, oracle_pk) do
+    case cache_through_read(state, Model.ActiveOracle, oracle_pk) do
       {:ok, Model.oracle(expire: old_expire, extends: extends) = m_oracle} ->
         new_expire = old_expire + delta_ttl
         extends = [{block_index, txi} | extends]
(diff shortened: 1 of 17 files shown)


lib/ae_mdw/db/mutations/oracle_extend_mutation.ex
```diff
@@ -6,14 +6,20 @@ defmodule AeMdw.Db.OracleExtendMutation do
...
   alias AeMdw.Txs
 
+  import AeMdw.Db.Oracle, only: [cache_through_read: 3]
+
+  import AeMdw.Db.Sync.Oracle,
+    only: [
+      cache_through_delete: 3,
+      cache_through_write: 3
+    ]
+
```
sborrazas, line 21: Why use imports instead of regular aliases? I think we should stick to regular aliases since it makes it easier to find functions by grepping them

lib/ae_mdw/db/sync/oracle.ex
```diff
@@ -0,0 +1,110 @@
+defmodule AeMdw.Db.Sync.Oracle do
```
sborrazas, line 1: 🙏

thepiwo (approved):

test/support/ae_mdw/db/mutation_case.ex
```diff
@@ -0,0 +1,24 @@
...
+    quote do
+      import AeMdw.TestUtil
+    end
+  end
+
+  setup tags do
+    alias AeMdw.Db.MemStore
+    alias AeMdw.Db.NullStore
+
+    if Map.get(tags, :integration, false) or Map.get(tags, :skip_store, false) do
```
sborrazas, line 18: I would remove adding any conditionals to this, or adding new configuration values.

We're simply creating a new store object which won't affect the way records are retrieved by default

Without this tests that have this pattern match `test "test", %{store: store} do` are simply going to fail

lib/ae_mdw/db/oracle.ex
```diff
@@ -39,6 +25,13 @@ defmodule AeMdw.Db.Oracle do
       )
   end
 
+  @spec oracle_tree!(Blocks.block_hash()) :: tuple()
+  def oracle_tree!(block_hash) do
+    block_hash
+    |> :aec_db.get_block_state()
+    |> :aec_trees.oracles()
+  end
+
```
sborrazas, line 34: Can we move this function back to where it was since it wasn't changed?

sborrazas (changes requested):

lib/ae_mdw/db/mutations/oracle_extend_mutation.ex
jyeshe: Because the definition of the functions is well located from Oracle to Oracle module. 

But we can adopt always the alias expecting to find them with Module.<function> or Sync<Module>.<function>

test/support/ae_mdw/db/mutation_case.ex
jyeshe: This is needed to allow a single setup for all test cases by `setup_all` like here:
https://github.com/aeternity/ae_mdw/blob/8fb32efa103cdf135d9c0c4df1aee976fee4786e/test/ae_mdw_web/controllers/aexn_transfer_controller_test.exs#L32

lib/ae_mdw/db/oracle.ex
jyeshe: you're right, during the refactoring it was only moved.

sborrazas (approved):

PR #931, opened 2022-09-30
refactor: print migrations total/duration using returned values

sborrazas:


thepiwo (approved):

jyeshe (commented):
Nice that the timings were moved to outside migration.

Wouldn't pass a non-transactional state (pointing to database) to every migration though as it has no prior state set and suggests replacing the `Database` use with a `State` that has the same effect with an extra abstraction that is not used. 

On the other hand it might be a good moment to rethink the migrations contract/interface. Following the idea of a `State` would use a transactional one outside the migration. The migrations would only return mutations in order to help achieving idempotent behaviour in a safer way (even though some past migrations could be simplified with `Database` and still be idempotent).

jyeshe:
Are you going to write test cases for the migrations? (:

I think testing directly with snapshots is more reliable (and needed). It's something that we already do for any migration right?

Just wanted to avoid an abstraction that wouldn't be used.

jyeshe (approved):

Issue #924, opened 2022-09-29
Update oracle expiration afer Oracle.extend internal call


jyeshe:

jyeshe:
solved by #933

PR #933, opened 2022-10-03
feat: index oracle extend internal calls

jyeshe:
Solve #924


jyeshe:
> doesn't this need a migration?

It's fine because every `Oracle.extend` was for a short delta and old oracles (the `extend` wouldn't make them active now).

Issue #943, opened 2022-10-10
display logs in the same order as they are emitted 


jyeshe:
Contract logs are displayed and sorted by contract calls but not by index (the order they were emitted)

Affected versions: since 1.0.0

PR #944, opened 2022-10-10
fix: sort event logs by index

jyeshe:
## What

Displays the event logs by the order they happened (sorting by index on the `Model.idx_contract_log`).

## Why

Solves #943

### DIFF
--- a/lib/ae_mdw/contracts.ex
+++ b/lib/ae_mdw/contracts.ex
@@ -3,7 +3,6 @@ defmodule AeMdw.Contracts do
   Context module for dealing with Contracts.
   """
 
-  alias :aeser_api_encoder, as: Enc
   alias AeMdw.Collection
   alias AeMdw.Contract
   alias AeMdw.Db.Format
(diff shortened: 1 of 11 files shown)


lib/ae_mdw/contracts.ex
```diff
@@ -17,6 +17,8 @@ defmodule AeMdw.Contracts do
   alias AeMdw.Util
   alias AeMdw.Validate
 
+  import AeMdwWeb.Helpers.ViewHelper
```
jyeshe, line 20: Single place to find encoding functions.

thepiwo (approved):

lib/ae_mdw/contracts.ex
sborrazas: Do we want for the `AeMdw` app to depend on `AeMdwWeb` app? Shouldn't it be the other way round? Perhaps we can move these functions to a module inside the `AeMdw` app. `Format` maybe?

jyeshe: It's a fair point even though all context `AeMdw` modules are already dealing with `AeMdwWeb` concerns like pagination and rendering the view. Moved to `AeMdw.Util.Encoding` until the view is rendered by specific modules for its owner controller replacing the `Format` module that renders for all controllers.

sborrazas (approved):

PR #952, opened 2022-10-12
chore: update aex141 signatures

jyeshe:
## What

Updates the detection of AEX-141 contracts respecting the Hackaton ones when there is a resync (by using the previous spec signatures).

## Why

Expects `total_supply` and new signatures for `transfer` and `meta_info` (without IPFS).


jyeshe:
Please find test cases at: test/ae_mdw/aexn_contracts_test.exs

thepiwo (approved):

Issue #725, opened 2022-06-16
all transactions for account endpoint


thepiwo:
make an endpoiint that returns all tranactions for one account
 - aex 9
 - aex 141
 - internal transfers (name fee, ...)
 - token migration tokens
 - function calls (Chain.spend)
 - native transactions
 
 with both plain tx that has that information, as well as

thepiwo:
(cc @CedrikNikita )

thepiwo:
name proposal from @sborrazas activity with different payload fields for different activity types. Paginated, never nested.

thepiwo:
## Goal

Expose all "activity" of one account, the consumer of the api wants to know everything that happened to the account, so
we should expose all data we know of.

## Basic Design

- always expose three field
    - enum for what type the "activity" is
    - original transaction it has been included (no nested aggregated data, maybe null if not exposed via transaction)
    - additional information depending on the "activity" type as we have it stored
- filters available
    - by account (eventually mandatory)
    - by enum "activity" (multiple if possible)
    - by height, txi and other scopes we usually provide

```json
[
  {
    "transaction": {
      ...
    },
    "type": "native_transaction",
    "additional_data": null
  },
  {
    "transaction": {
      ...
    },
    "type": "aex9_transfer",
    "additional_data": {
      "amount": 10,
      "contract_id": "ct_UAKGhMFJThpxgyFQi4fXztvT2fZRcMgZxjqrT9ehwsZT1yUe1",
      "recipient": "ak_y87WkN4C4QevzjTuEYHg6XLqiWx3rjfYDFLBmZiqiro5mkRag",
      "sender": "ak_45tM79u5EgCmsorADwuhf4PXkzEEBSHzXRZwjfBe5ww48XX63",
      ...
    }
  }
]
```

## Examples

### SpendTx (any native)

```json
{
  "transaction": {
    "block_hash": "mh_EgvpfzAWAxes5xoFut3ADQBk9G9FpwdSFGnfXiV3anErouSC5",
    "block_height": 621066,
    "hash": "th_zDTyHWyVxMvmZmPhQZgnpdtHsbRMR6aLj7aADBKJqekkstmzG",
    "micro_index": 0,
    "micro_time": 1656616116824,
    "signatures": [
[...]

PR #957, opened 2022-10-13
feat: add tx internal transfers to activities

sborrazas:
refs #725

### DIFF
--- a/test/ae_mdw_web/controllers/activities_controller_test.exs
+++ b/test/ae_mdw_web/controllers/activities_controller_test.exs
@@ -411,6 +411,122 @@ defmodule AeMdwWeb.ActivitiesControllerTest do
       assert %{"data" => ^data} = conn |> with_store(store) |> get(prev_url) |> json_response(200)
     end
 
+    test "when it has both gen-based and tx-based int transfers", %{conn: conn} do
+      account_pk = TS.address(0)
+      account = Enc.encode(:account_pubkey, account_pk)
+      height = 398
+      [txi1, txi2, txi3] = [123, 456, 789]
+
+      store =
+        empty_store()
+        |> Store.put(
+          Model.TargetKindIntTransferTx,
+          Model.target_kind_int_transfer_tx(
+            index: {account_pk, "reward_oracle", {height, txi1}, txi1}
+          )
+        )
+        |> Store.put(
+          Model.IntTransferTx,
+          Model.int_transfer_tx(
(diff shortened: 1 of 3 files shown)


test/ae_mdw_web/controllers/activities_controller_test.exs
```diff
@@ -324,6 +324,87 @@ defmodule AeMdwWeb.ActivitiesControllerTest do
...
+               "type" => "InternalTransferEvent",
+               "payload" => %{
+                 "kind" => "fee_lock_name",
+                 "amount" => 20
+               }
+             } = activity2
+
+      assert %URI{query: query} = URI.parse(next_url)
+
+      assert %{"cursor" => _cursor, "direction" => "forward", "limit" => "2"} =
```
jyeshe, line 404: May it make a request to the `next_url` to test the cursor deserialization?

test/integration/ae_mdw_web/controllers/activity_controller_test.exs
```diff
@@ -183,4 +183,26 @@ defmodule Integration.AeMdwWeb.ActivityControllerTest do
              ] = events
     end
   end
+
+  test "it gets tx-based internal transfer events", %{conn: conn} do
```
jyeshe, line 187: Could we have an extra one that would have both generational and transactional internal transfer events and test the pagination if it skips both internal transfers on the next page?

jyeshe (commented):
Code looks good for me. Only some test automation requests.

thepiwo (approved):

test/ae_mdw_web/controllers/activities_controller_test.exs
sborrazas: Done

test/integration/ae_mdw_web/controllers/activity_controller_test.exs
sborrazas: I searched the entire transfers database and wasn't able to find a generation that contains both internal and external transfers from the same account

jyeshe: Thanks, could you add in the CI then? It's possible for an account that receives a dev reward to also create an oracle, right?

sborrazas: Added a unit tests with both cases combined now, thanks

jyeshe (approved):

PR #958, opened 2022-10-14
fix: render bid for names in auction

jyeshe:
## What/Why

Fix for https://mainnet.aeternity.io/mdw/v2/names/:name when name is in auction. 

## Notes

Adds `:name_claim_tx` to `BlockchainSim`

### DIFF
--- a/lib/ae_mdw/db/format.ex
+++ b/lib/ae_mdw/db/format.ex
@@ -31,9 +31,16 @@ defmodule AeMdw.Db.Format do
   def to_raw_map(
         state,
         {:tx, index, hash, {kb_index, mb_index}, mb_time},
-        {block_hash, type, signed_tx, tx_rec}
+        {block_hash, tx_type, signed_tx, tx_rec}
       ) do
-    tx_map = to_raw_map(state, tx_rec, type) |> put_in([:type], type)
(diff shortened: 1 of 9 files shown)


thepiwo (approved):

lib/ae_mdw/db/format.ex
```diff
@@ -65,8 +72,11 @@ defmodule AeMdw.Db.Format do
 
     {status, auction} =
       case Name.locate_bid(state, plain_name) do
-        nil -> {:name, nil}
-        key -> {:auction, to_raw_map(state, key, Model.AuctionBid)}
+        nil ->
+          {:name, nil}
+
+        m_auction_bid ->
```
sborrazas, line 78: Perhaps we can leave `locate_bid/2` to return `{:ok, auction} | nil` since it's better to pattern match on it and only update this line to handle it without any other changes?

jyeshe: There are some other uses that would make it not straightforward and a non-tagged tuple is more symmetrical with `nil` that might also be returned.

I propose merging it as is and the legacy code can be refactored to work with ok/error tuples.

sborrazas (approved):

Issue #942, opened 2022-10-10
Allow names filtered by owner to be ordered by expiration


sborrazas:
Right now you get an error back: `invalid query: can't order by expiration when filtering by owner`

marc0olo:
would be
- https://mainnet.aeternity.io/mdw/v2/names?owned_by=ak_25BWMx4An9mmQJNPSwJisiENek3bAGadze31Eetj4K4JJC8VQN&by=deactivation

I guess, right?

marc0olo:
wondering a bit about the name change:
- https://github.com/aeternity/ae_mdw/pull/623

sborrazas:
@marc0olo that's right! It would be deactivation, not expiration

marc0olo:
ok what's the exact difference between?
- expiration
- deactivation

sborrazas:
They are both the same, but since we're using the word `activation` for when a name becomes active, we're also using the word `deactivation` when it becomes inactive

marc0olo:
just for clarification: `activation` comes from the fact that a `NameClaimTx` doesn't immediately activate the name in case of auctions.

personally I still prefer `expiration` over `deactivation` 🤔

marc0olo:
makes absolute sense, thanks!

sborrazas:
@thepiwo not yet

thepiwo:
@sborrazas but now it is

janmichek:
Seems like this is still not avialable. THe error message is the same. Is it released?
https://mainnet.aeternity.io/mdw/v2/names?owned_by=ak_25BWMx4An9mmQJNPSwJisiENek3bAGadze31Eetj4K4JJC8VQN&by=deactivation

thepiwo:
@janmichek the fix is not released yet, will be ready in the next days

PR #965, opened 2022-10-18
feat: allow filtering names by owner/state ordered by deactivation

sborrazas:
refs #942

### DIFF
--- a/lib/ae_mdw/names.ex
+++ b/lib/ae_mdw/names.ex
@@ -45,8 +45,8 @@ defmodule AeMdw.Names do
   @table_activation Model.ActiveNameActivation
   @table_active_expiration Model.ActiveNameExpiration
   @table_active_owner Model.ActiveNameOwner
-  @table_active_owner_expiration Model.ActiveNameOwnerExpiration
-  @table_inactive_owner_expiration Model.InactiveNameOwnerExpiration
+  @table_active_owner_deactivation Model.ActiveNameOwnerDeactivation
+  @table_inactive_owner_deactivation Model.InactiveNameOwnerDeactivation
   @table_inactive Model.InactiveName
   @table_inactive_expiration Model.InactiveNameExpiration
   @table_inactive_owner Model.InactiveNameOwner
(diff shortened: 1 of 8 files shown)


thepiwo (approved):

lib/ae_mdw/names.ex
```diff
@@ -175,8 +182,58 @@ defmodule AeMdw.Names do
...
+      state
+      |> Collection.stream(@table_active_owner_expiration, direction, scope, cursor)
+      |> Stream.map(fn {^owner_pk, expiration_height, plain_name} ->
+        {{expiration_height, plain_name}, :active}
+      end)
+    end
+  end
+
+  defp build_height_streamer(%{owned_by: owner_pk, state: "inactive"}, state, scope, cursor) do
+    scope = serialize_owner_expiration_scope(owner_pk, scope)
```
jyeshe, line 199: nit pick for no sending through the wires so instead of `serialize` could it be a `get`? (also because all over the places we are using serialize when encoding the cursor so serialize cursor on the line below gets mixed with the same naming for other purpose)

jyeshe: As the scope concept is already been applied to `gen:first_gen-last_gen` and `txi:first_txi-last_txi` may we continue calling it `key_boundary` (or other better name if you want)? I understand you're avoiding big names (: but I think it's needed because it's not a scope.

jyeshe (changes requested):
Needs to update the index also on name revoke because by effect the `order_by` parameter is on name deactivation not only on expiration. One point for the naming importance (: while it suggests a renaming from `expiration` to `deactivation` for the created functions and model.

After that could you also update these test cases?
test/ae_mdw/db/name_claim_mutation_test.exs
test/ae_mdw/db/name_revoke_mutation_test.exs

lib/ae_mdw/names.ex
sborrazas: Done, renamed the variables/functions

sborrazas:
> Needs to update the index also on name revoke because by effect the `order_by` parameter is on name deactivation not only on expiration. One point for the naming importance (: while it suggests a renaming from `expiration` to `deactivation` for the created functions and model.
> 
> After that could you also update these test cases? test/ae_mdw/db/name_claim_mutation_test.exs test/ae_mdw/db/name_revoke_mutation_test.exs

The inactive record is inserted on the private function `Sync.Name.deactivate_name` which applies both for expiring names and name revokes. Added these two cases on the revoke/claim test cases and they indeed insert it

jyeshe:
> > Needs to update the index also on name revoke because by effect the `order_by` parameter is on name deactivation not only on expiration. One point for the naming importance (: while it suggests a renaming from `expiration` to `deactivation` for the created functions and model.
> > After that could you also update these test cases? test/ae_mdw/db/name_claim_mutation_test.exs test/ae_mdw/db/name_revoke_mutation_test.exs
> 
> The inactive record is inserted on the private function `Sync.Name.deactivate_name` which applies both for expiring names and name revokes. Added these two cases on the revoke/claim test cases and they indeed insert it

It would be missing a rename all (for the PR) from `expiration` to `deactivation`.

jyeshe (commented):
Renaming to complete.

jyeshe (approved):

Issue #751, opened 2022-06-29
Expose active accounts count and active generalized accounts count


sborrazas:
refs #124 

@marc0olo what would be considered an "active" account?

marc0olo:
> @marc0olo what would be considered an "active" account?

IMO accounts that have been seen on the network, meaning not returning a 404 from node 😅 but it's kind of debatable in that sense because you can send AE or AEX-9 tokens to accounts which will then never turn into "active accounts".

I think that's sth. we have to agree upon. what do you think @thepiwo ?

maybe we could even distinguish between those two types:
- passive in existence (never performed a tx, but seen on the network)
- active in existence (performed at least one transaction, including outgoing tx via `PayingForTx`)

thepiwo:
@marc0olo sounds good

marc0olo:
cool :)

I think for generalized accounts count we more or less have that already to some extent by calling https://mainnet.aeternity.io/mdw/v2/txs/count?type=ga_attach

but I am actually not sure if this is a reliable number because the `GaAttachTx` could also revert, right? 🤔 

anyway, I would consider every valid and successful `GaAttachTx` as active GA. I see no reason to declare them as "passive".

just thinking about this topic since we aim to display the amount of GA contracts currently on explorer dashboard in the contracts section.

thepiwo:
@sborrazas do you have any input on that?

I agree, `GaAttachTx` could revert, but would probably still be included in the count, maybe we can make a `count_ok` or similar.

sborrazas:
Counting GA accounts that way would be an easy feature to incorporate. Counting total node accounts wouldn't be that easy, though

thepiwo:
For GA this is already fixed, for normal accounts its expensive to compute similar to #755

Maybe we can include fixing count per transaction type fix for successful within this. @marc0olo would that be good?

marc0olo:
I think it needs to be a separate attribute in the stats endpoint if it's about the GA contract count.

for counting `ga_attach` I would still expect to have a count for all txs, including the failed/reverted txs.

opinions?

sborrazas:
@marc0olo this information is already exposed in the txs count endpoint filtered by tx_type: https://mainnet.aeternity.io/mdw/v2/txs/count?type=ga_attach

marc0olo:
yes, I just want to make sure we don't replace that count somehow.

> Maybe we can include fixing count per transaction type fix for successful within this. @marc0olo would that be good?

[...]

PR #971, opened 2022-10-21
fix: increment ga contract stat only on success

jyeshe:


thepiwo (commented):
for attach its fine, won't be a GA then, do we have any of those?

thepiwo:
refs #751 right?

jyeshe:
> refs #751 right?

correct and taken as related to return_type recently implemented

jyeshe:
> for attach its fine, won't be a GA then, do we have any of those?

Do you mean the contract will exist any way even if return is no `:ok`?

thepiwo:
the transaction will exist, but no action from it, I was wondering if there are any reverted ga attach on chain, so if this change makes a difference at all (with current state of course)

jyeshe:
> the transaction will exist, but no action from it, I was wondering if there are any reverted ga attach on chain, so if this change makes a difference at all (with current state of course)

From Node api it is like a contract create where the tx is added to the block but the contract is not created because the call failed.

sborrazas (approved):

thepiwo (approved):

marc0olo:
for contracts this was already correct, right?

PR #976, opened 2022-10-26
fix: render list of keyword lists args

jyeshe:
Fix for https://testnet.aeternity.io/mdw/txi/32440285 involving `:contract_create_tx` args rendering

### DIFF
--- a/lib/ae_mdw/contract.ex
+++ b/lib/ae_mdw/contract.ex
@@ -473,6 +473,14 @@ defmodule AeMdw.Contract do
     }
   end
 
+  defp element_value({type, [head | _tail] = keyword_list})
+       when is_atom(type) and is_tuple(head) and tuple_size(head) == 2 do
+    %{
+      "type" => to_string(type),
+      "value" => Enum.into(keyword_list, %{})
+    }
+  end
+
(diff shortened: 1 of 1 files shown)


thepiwo (approved):

lib/ae_mdw/contract.ex
```diff
@@ -473,6 +473,14 @@ defmodule AeMdw.Contract do
     }
   end
 
+  defp element_value({type, [head | _tail] = keyword_list})
+       when is_atom(type) and is_tuple(head) and tuple_size(head) == 2 do
```
sborrazas, line 477: Is is guaranteed that if the first element is a tuple of 2 values then the rest will also be?

jyeshe: For `type = :tuple` I would expect that the SDK encoding is uniform. The code is more inclusive because if `type != :tuple` and `keyword_list` is any list of pairs it's able to render it. Any other encoding would require it's own investigation.

sborrazas: Makes sense. Thanks

sborrazas (approved):

PR #981, opened 2022-10-28
fix: events from the node are obtained in reverse order

sborrazas:
Events are added to the tx_env in a reverse order: https://github.com/aeternity/aeternity/blob/e66a588a41d8af5f4ce8467f28258d3f2c74cd8e/apps/aetx/src/aetx_env.erl#L324

### DIFF
--- a/lib/ae_mdw/contract.ex
+++ b/lib/ae_mdw/contract.ex
@@ -407,7 +407,7 @@ defmodule AeMdw.Contract do
       {:ok, _, _, events} =
         :aec_block_micro_candidate.apply_block_txs_strict(txs_taken, trees_in, env)
 
-      events
+      Enum.reverse(events)
     else
       []
     end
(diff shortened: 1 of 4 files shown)


lib/ae_mdw/contract.ex
```diff
@@ -407,7 +407,7 @@ defmodule AeMdw.Contract do
       {:ok, _, _, events} =
         :aec_block_micro_candidate.apply_block_txs_strict(txs_taken, trees_in, env)
 
-      events
+      Enum.reverse(events)
```
jyeshe, line 410: This needs to be adapted for the reversal:
https://github.com/aeternity/ae_mdw/blob/c19e09b18e2cba71adde92c4963da25ea2e6b81f/lib/ae_mdw/db/sync/contract.ex#L60-L62

Looks like we should have a test case for this dependency on the sequence related to `Chain.create` and `Chain.clone`

sborrazas: Thanks for realizing! I've updated events mutations creation and tests accordingly

jyeshe (approved):

thepiwo (approved):

Issue #969, opened 2022-10-20
Index AEX-N primarly using events


thepiwo:
As discussed we want to continuously index by tracking events, rather than dry-runs.

For AEX-9 we would need specifically one dry-run after contract creation as `init` won't give any events in the current implementation. Later account presence changes can be observed using Transfer and Mint Events. For balance updates I think we currently don't actively track them, single account balances could be dry-runned in constant time.

For AEX-141 I think we can have similar approach, but eventually skip dry-running at all (cc @marc0olo)

@jyeshe is there anything you would like to add. I think for planning on what specifically to do it would be nice to have a concise but full summary of the changes we want to do.

jyeshe:
As contract samples after `init` state (on testnet):
`ct_ZwpiXbWme5EC72FCDoAvJMqkuwgGaJkKEyN5VCwewrYoWaTEW`, 3000 account balances, 1.2 second
`ct_HQyn8crD6APzS7VPguNKEdVMYqdupPG95rfQU2a9ztosFLtmh`, 4000 account balances 1.6 second

thepiwo:
@jyeshe interesting so they created thousands of accounts in contract create?

jyeshe:
> @jyeshe interesting so they created thousands of accounts in contract create?

Exactly

jyeshe:
As discussed we gonna approach the contract `init` state with synchronous calls limiting the dry-run time with the `gas`.

jyeshe:
Aex9 contracts with no logs returned by https://testnet.aeternity.art/mdw/v2/contracts/logs?contract=:contract_id

```
["ct_2cYkXZjXcd6qjnoZPPmH4NVpo6FR4PUYmA14mmckg7oUuqhDtj",
 "ct_NfTemzEwdWvLQ2wwjmpJuRjTHdTdqqtcqeVGmY9MhsbJHkk42",
 "ct_2mj7kpXoR36Y2iMrHFcZxww7fyzfmgZHNjqmmGUFoRHseVDgWF",
 "ct_HCQLprHjoTRZMb1kGDvStSa3bhK59b167SAzGBgDxuDCyD2pD",
 "ct_2Fj6Km8886DdRea4Tp92ggBaoAyknkMB1kG7o3CGux2CB1ynGC",
 "ct_2DbY6e5useaDjbd3bWX38e4tZYw7gS9ZgP79mbDFcgCmmYtaPq",
[...]

PR #983, opened 2022-10-31
feat: index aex9 balance based on events

jyeshe:
## What

- Indexes aex9 balances based on `Burn`, `Mint`, `Swap` and `Transfer` events. As `Mint` wasn't mandatory on contract creation it also dry-runs after contract init.
- Exposes these balances on new endpoint: `/v2/aex9/:contract_id/event-balances` (until it's validated and chosen to replace the dry-run only approach).

## Why

Refs #969


sborrazas (approved):

thepiwo (approved):
awesome work! Have you compared the results?

jyeshe:
> awesome work! Have you compared the results?

For a full scan it would be better on staging. A script comparing the results for all contracts will be used helping to identify invalid contracts (non-compliant ones). 

In terms of speed some results were clearly in favor of the event-balances.

Issue #984, opened 2022-10-31
Index template NFTs


jyeshe:
This is a class of NFTs that might also be considered semi-fungible but due to their unique token ids and for the sake of communication they are also called NFTs.

Besides tracking the minting of NFTs derived from a template (having the same metadata and therefore the same media), the Mdw will display info about the templates themselves (list of template ids along with the edition limit) according to https://github.com/aeternity/AEXs/pull/149

PR #991, opened 2022-11-02
feat: index nft templates

jyeshe:
## What

- Indexes nft templates and the templates ids
- Expose templates at `/v2/aex141/:contract/templates/

## Why
Refs #984 

## Notes
Signatures will be validated after https://github.com/aeternity/AEXs/pull/149 review and merge.


thepiwo (approved):

lib/ae_mdw/aex141.ex
```diff
@@ -126,6 +149,16 @@ defmodule AeMdw.Aex141 do
     end
   end
 
+  defp deserialize_cursor(Model.NftTemplate, cursor_bin64) do
+    with {:ok, cursor_bin} <- Base.decode64(cursor_bin64),
+         {<<_pk::256>>, template_id} = cursor when is_integer(template_id) <-
+           :erlang.binary_to_term(cursor_bin) do
```
sborrazas, line 155: We should add another check for the template_id value type and length, it could be any erlang term that the use can build, including functions and large structures

sborrazas (changes requested):

lib/ae_mdw/aex141.ex
jyeshe: There is a `when is_integer(template_id)`

sborrazas: Thought of something else, my bad!

sborrazas (approved):

PR #1003, opened 2022-11-08
fix: get recipient account pointer by id type

jyeshe:
Gets recipient account pointer by value (id type) not by key. Some account pointers are not `account_pubkey` but have account ids: https://testnet.aeternity.io/mdw/txi/30950462

### DIFF
--- a/lib/ae_mdw/db/name.ex
+++ b/lib/ae_mdw/db/name.ex
@@ -264,21 +264,11 @@ defmodule AeMdw.Db.Name do
   end
 
   defp pointee_at(state, Model.name(index: name, updates: updates), ref_txi) do
-    updates
-    |> find_update_txi_before(ref_txi)
-    |> case do
-      nil ->
-        {:error, {:pointee_not_found, name, ref_txi}}
+    update_txi = find_update_txi_before(updates, ref_txi)
 
-      update_txi ->
-        {:id, :account, pointee_pk} =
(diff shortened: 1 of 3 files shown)


thepiwo (approved):

thepiwo:
After our earlier discussion I disagree here, we should only consider the protocol defined `account_pubkey` for recipients

jyeshe:
> After our earlier discussion I disagree here, we should only consider the protocol defined `account_pubkey` for recipients

Even though the Node allows other account pointer keys, I couldn't indeed submit successfully a spendTx with a name id pointing to account with key "0". Using `buildTx` it doesn't submit and currently the SDK resolves with `"account_pubkey"`:
https://github.com/aeternity/aepp-sdk-js/blob/9b4cee8636daf28931896ca6744c03e265fa78fb/src/spend.ts#L73

Previously it was possible to the name `testauction1.chain` but it had a later `name_update_tx` including the `account_pubkey`:
https://testnet.aeternity.art/mdw/v2/txs/th_xmGuLJ6B9CPJBGT8BJKtvyEPBVY4efaNYAizAxLa4SQQmoUVq
https://testnet.aeternity.art/mdw/v2/txs/th_2RDoTT22XbT5RDAGePrPdCRkxhAqPX85uTZtKpxea69tZv7LUD

Closing as not a bug.

PR #1004, opened 2022-11-08
fix: render binary pointer key on name related endpoints

### LINKED ISSUE #986: name pointers with arbitrary binary key is not rendered on names endpoint
Name pointers will only diplay a single value right now. We should look into why this behaves like this https://github.com/aeternity/ae_mdw/blob/f6f69e3bdf90aa2547746d6e85aed0d3724cb49b/lib/ae_mdw/names.ex#L382

This might have been introduced here https://github.com/aeternity/ae_mdw/commit/ce9293bf582a7e799710f90ca4974f033dc45b84#diff-6509d83849ebb5bd9a1db2a6cf655650104e3f1d93a0a09d4f9a54b93f16eaa2R201

jyeshe:
## What

Fixes the rendering of binary pointer keys: https://testnet.aeternity.art/mdw/v2/txs?type=name_update&limit=20&direction=forward  
where the "invalid byte 0xAE" is in the non-string binary pointer: `<<104, 65, 117, 174, 49, 251, 29, 202, 69, 174, 147, 56, 60, 150, 188, 247, 149, 85, 150, 148, 88, 102, 186, 208, 87, 101, 78, 111, 189, 5, 144, 101>>`

Sample urls fixed:
http://localhost:4000/v2/names?cursor=138231-test.test&direction=forward&limit=1
http://localhost:4000/v2/names/test.test
http://localhost:4000/name/test.test
http://localhost:4000/v2/txs?cursor=29614&limit=1
http://localhost:4000/v2/txs/29614

## Why

Closes #986 (it turns out that ce9293bf was a revert).

## Notes

*binary* here in the meaning of not text (every pointer key is a `binary()`)

### DIFF
--- a/lib/ae_mdw/db/format.ex
+++ b/lib/ae_mdw/db/format.ex
@@ -93,13 +93,13 @@ defmodule AeMdw.Db.Format do
 
   @spec encode_pointers(list() | map()) :: %{iodata() => String.t()}
   def encode_pointers(pointers) when is_map(pointers) do
-    Enum.into(pointers, %{}, fn {key, id} -> {maybe_to_list(key), enc_id(id)} end)
+    Enum.into(pointers, %{}, fn {key, id} -> {maybe_encode_base64(key), enc_id(id)} end)
   end
 
   def encode_pointers(pointers) do
     Enum.map(pointers, fn
-      %{"key" => key, "id" => id} -> %{"key" => maybe_to_list(key), "id" => id}
-      %{key: key, id: id} -> %{key: maybe_to_list(key), id: id}
+      %{"key" => key, "id" => id} -> %{"key" => maybe_encode_base64(key), "id" => id}
+      %{key: key, id: id} -> %{key: maybe_encode_base64(key), id: id}
     end)
   end
 
@@ -388,6 +388,14 @@ defmodule AeMdw.Db.Format do
 
   defp custom_encode(_state, _tx_type, tx, _tx_rec, _signed_tx, _txi, _block_hash),
     do: tx
+
+  defp maybe_encode_base64(bin) do
+    if String.valid?(bin) do
+      bin
+    else
+      Base.encode64(bin)
+    end
+  end
 
   defp maybe_base64(bin) do
     try do
--- a/lib/ae_mdw/db/format.ex
+++ b/lib/ae_mdw/db/format.ex
@@ -93,21 +93,19 @@ defmodule AeMdw.Db.Format do
 
   @spec encode_pointers(list() | map()) :: %{iodata() => String.t()}
   def encode_pointers(pointers) when is_map(pointers) do
-    Enum.into(pointers, %{}, fn {key, id} -> {maybe_encode_base64(key), enc_id(id)} end)
+    Enum.into(pointers, %{}, fn {key, id} -> {maybe_base64_pointer_key(key), enc_id(id)} end)
   end
 
   def encode_pointers(pointers) do
     Enum.map(pointers, fn
-      %{"key" => key, "id" => id} -> %{"key" => maybe_encode_base64(key), "id" => id}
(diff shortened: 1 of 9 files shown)


sborrazas (commented):
I think we should use the same representation the node uses for rendering pointers: https://github.com/aeternity/aeternity/blob/master/apps/aens/src/aens_pointer.erl#L41-L44

Perhaps we can use this function itself?

jyeshe:
> I think we should use the same representation the node uses for rendering pointers: https://github.com/aeternity/aeternity/blob/master/apps/aens/src/aens_pointer.erl#L41-L44
> 
> Perhaps we can use this function itself?

The `tx` comes serialized by the Node with `:aetx_sign.serialize_for_client(header, signed_tx)`.

The fix is still needed because for binary arbitrary pointer keys the value is forwarded as is (L43).

jyeshe:
@thepiwo @sborrazas there were other endpoints from v1 and `/v2/names/:name` that had the same issue. The fix was generalized in this PR.

Please notice for
http://localhost:4000/v2/names?cursor=138231-test.test&direction=forward&limit=1
http://localhost:4000/v2/names/test.test
http://localhost:4000/name/test.test
http://localhost:4000/v2/txs?cursor=29614&limit=1
http://localhost:4000/v2/txs/29614

the encoding is the same (if not a valid UTF8 string, it converts to a list and the browser might render this list in different ways)

jyeshe:
> just to clarify, for pointer keys defined by the protocol we want to always return the correct decoded id, only for arbitrary pointer keys we want to always return base64 as we shouldn't guesstimate.

The detection of valid string is very objective and reliable provided by `String.valid?`. 

 

> just to clarify, for pointer keys defined by the protocol we want to always return the correct decoded id, only for arbitrary pointer keys we want to always return base64 as we shouldn't guesstimate.

The current solution already detects when it's not printable in a very objective and reliable way with `String.valid?`. 

Changed the encoding from `:erlang.binary_to_list(bin)` used since legacy v1.0 to `Base.encode64`, after detecting is not printable.

thepiwo:
as discussed with @hanssv all serialization is `binary()` thus we should only differentiate between protocol and arbitrary pointers as summarized here https://github.com/aeternity/ae_mdw/pull/1004#issuecomment-1308555264 not based on them being a valid string potentially.

jyeshe:
> as discussed with @hanssv all serialization is `binary()` thus we should only differentiate between protocol and arbitrary pointers as summarized here [#1004 (comment)](https://github.com/aeternity/ae_mdw/pull/1004#issuecomment-1308555264) not based on them being a valid string potentially.
 
All serialization been `binary()` wouldn't be not enough for JSON but no problem, in the chat we had talked about 3 kinds of pointer keys:
```
a) valid string/printable standard
b) valid string/printable non-standard/custom
c) invalid string
```
it was requested to encode (c) as base64, now (b) is also encoded as per last request.

lib/ae_mdw/db/format.ex
```diff
@@ -365,6 +387,12 @@ defmodule AeMdw.Db.Format do
   defp custom_encode(_state, _tx_type, tx, _tx_rec, _signed_tx, _txi, _block_hash),
     do: tx
 
+  defp maybe_base64_pointer_key(key)
+       when key in ["account_pubkey", "oracle_pubkey", "contract_pubkey", "channel_pubkey"],
```
sborrazas, line 391: Contrary to intuition, `channel_pubkey` should be just `channel`, as per [this implementation](https://github.com/aeternity/aeternity/blob/a5b4cfbf3f2afb30ae6cb06f0b21858a06652bb2/apps/aens/src/aens.erl#L38)

sborrazas (approved):
Just one comment, rest LGTM

thepiwo (approved):

PR #1005, opened 2022-11-08
fix: scope contract calls filtered by function properly

sborrazas:
When filtering by a function name prefix the scoping of the query was being ignored

### DIFF
--- a/lib/ae_mdw/contracts.ex
+++ b/lib/ae_mdw/contracts.ex
@@ -114,15 +114,17 @@ defmodule AeMdw.Contracts do
     |> Stream.filter(&match?(Model.int_contract_call(fname: ^fname), &1))
   end
 
-  defp build_logs_pagination(
-         %{data_prefix: data_prefix},
-         state,
-         {first_call_txi, last_call_txi},
-         cursor
-       ) do
+  defp build_logs_pagination(%{data_prefix: data_prefix}, state, scope, cursor) do
     key_boundary =
-      {{data_prefix, first_call_txi, @min_txi, @min_hash, @min_idx},
-       {data_prefix <> @max_blob, last_call_txi, @max_txi, @max_hash, @max_idx}}
+      case scope do
+        nil ->
+          {{data_prefix, @min_txi, @min_txi, @min_hash, @min_idx},
+           {data_prefix <> @max_blob, @max_txi, @max_txi, @max_hash, @max_idx}}
+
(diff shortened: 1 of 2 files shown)


thepiwo (approved):

lib/ae_mdw/contracts.ex
```diff
@@ -222,14 +222,12 @@ defmodule AeMdw.Contracts do
...
-         [fname: fname_prefix],
+         [fname: fname],
          state,
          {first_call_txi, last_call_txi},
          cursor
        ) do
-    key_boundary =
-      {{fname_prefix, first_call_txi, @min_idx},
-       {fname_prefix <> Util.max_256bit_bin(), last_call_txi, @max_idx}}
+    key_boundary = {{fname, first_call_txi, @min_idx}, {fname, last_call_txi, @max_idx}}
```
jyeshe, line 230: Don't know if we can solve this way because in the docs and historically it works with `fname` prefix. 
Could you share how it would affect or not the client apps and how would be the policy for the general public?

jyeshe: I am not sure, since 1.0 it works as a prefix:
https://github.com/aeternity/ae_mdw/blob/a7175ed73078a378f50647417105acc786c5191b/lib/ae_mdw_web/controllers/contract_controller.ex#L279
Described also in the [README.md](https://github.com/aeternity/ae_mdw/blob/1.0.9/README.md?plain=1#L3151)

jyeshe: Here is an example from 1.0 for all Oracle.* internal calls:
https://github.com/aeternity/ae_mdw/blob/a7175ed73078a378f50647417105acc786c5191b/README.md?plain=1#L3189

sborrazas: @jyeshe you're right 100%, what would you suggest we do? We can't return invalid results

jyeshe: I am sure you can come with an intelligent solution. That would be a very interesting problem. I would start looking at secondary indexes implementation or negotiate with @thepiwo if it would be acceptable for public and internal client apps.

sborrazas:
As per discussed with @thepiwo, we're simply going to disable filtering by both scope and function prefix

jyeshe (approved):

thepiwo (approved):

PR #1008, opened 2022-11-09
chore: allow /txs/count to be filtered by tx_type

sborrazas:
This change allows filtering txs count by both `type` and `tx_type` to avoid the request being blocked by browser tracker/ad blockers.


thepiwo (approved):

thepiwo:
@lukeromanowicz we will let you know when it is released

jyeshe (changes requested):
Could you provide a test case?

sborrazas:
@jyeshe done

PR #1009, opened 2022-11-09
fix: ignore oracle queries that do not have the right calculated nonce

sborrazas:


thepiwo (approved):
seems to work for me :)

jyeshe (approved):
Possibly ignoring could be avoided using the `AeMdw.Txs` to make a query as:
https://testnet.aeternity.art/mdw/v2/micro-blocks/mh_2WKHGWCxmM1rQ8okcjTytFyrKeAEs5Ark9Vasyp53QxJwtnPqM/txs?account=ak_2R3dHQ65d63ZBxcKiT2wzConHnbNd5ewVcpBo2ezkSCVw3zY7w
and then consider the nonce of the transactions that happened before the ones with `Oracle.query`. 

However like the original PR, this query nonce fix is temporary and not complete in favor of Nodes PR.

PR #1016, opened 2022-11-13
feat: display name buyer from inner claim tx

### LINKED ISSUE #1012: When listing names, the buyer should also be present for ga_meta
The buyer of the name is the one that claimed the name last, meaning, the original owner before any transfer

jyeshe:
Closes #1012

### DIFF
--- a/lib/ae_mdw/db/name.ex
+++ b/lib/ae_mdw/db/name.ex
@@ -18,6 +18,7 @@ defmodule AeMdw.Db.Name do
   alias AeMdw.Db.Model
   alias AeMdw.Db.Format
   alias AeMdw.Db.State
+  alias AeMdw.Db.Sync.InnerTx
   alias AeMdw.Db.Util, as: DbUtil
   alias AeMdw.Names
   alias AeMdw.Validate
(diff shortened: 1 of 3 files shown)


thepiwo (approved):

lib/ae_mdw/db/name.ex
```diff
@@ -186,6 +187,16 @@ defmodule AeMdw.Db.Name do
             name_hash == :aens_transfer_tx.name_hash(tx)
           end)
           |> :aens_transfer_tx.account_id()
+
+        {_block_hash, tx_type, _signed_tx, tx_rec}
+        when tx_type in [:ga_meta_tx, :paying_for_tx] ->
+          {_mod, tx_rec} =
```
sborrazas, line 193: Can we pattern-match `:name_claim_tx` here using `:aetx.specialize_type()` at the end?

That way if the transaction belongs to the wrong type we get a pattern-match error instead of an error triggered by the `:aens_claim_tx/account_id/1`

jyeshe: sure, replaced the `specialize_callback` with `specialize_type` for that.

sborrazas (approved):

PR #1019, opened 2022-11-15
chore: divide swagger v2 docs into separate resource files

sborrazas:
Plus, added some linting for the swagger/README as part of the CI.

The next step would be to update the swagger v2 docs.


thepiwo (approved):
lgtm, haven't tried it out

jyeshe (approved):

PR #1020, opened 2022-11-16
fix: avoid double aex9 event balance update

jyeshe:
Ignores aex9 events for the purpose of event balance update on contract creation.


thepiwo (approved):
cool! so does this prioritize events over dry-run results or other way round?

jyeshe:
> cool! so does this prioritize events over dry-run results or other way round?

It uses always the dry-run result for contract creation as the Mint was not mandatory. But if you prefer priorizing the events it can switch after a certain height.

sborrazas (approved):

PR #1022, opened 2022-11-16
docs: add new block endpoints on swagger v2 docs

sborrazas:
refs #998

### DIFF
--- a/docs/swagger_v2/blocks.spec.yaml
+++ b/docs/swagger_v2/blocks.spec.yaml
@@ -1,5 +1,5 @@
 schemas:
-  Block:
+  KeyBlock:
     example:
       beneficiary: ak_2MR38Zf355m6JtP13T3WEcUcSLVLCxjGvjk6zG95S2mfKohcSS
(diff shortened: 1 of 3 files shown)


docs/swagger_v2/blocks.spec.yaml
```diff
@@ -1,5 +1,5 @@
 schemas:
-  Block:
+  KeyBlock:
```
thepiwo, line 2: a change to the old endpoint name will work as "breaking" in terms of how the aepps consume them currently I guess

thepiwo: maybe @davidyuk can give some input on that?

jyeshe (approved):

docs/swagger_v2/blocks.spec.yaml
davidyuk: changing `operationId` is a breaking change for apps that using `swagger-client`, yes

thepiwo (approved):

PR #1030, opened 2022-11-21
refactor: move ws subscription to specific module

### LINKED ISSUE #1010: replace riverside with built-in phoenix websocket
As requested and discussed, riverside library dependency will be removed in favor of Phoenix built-in websocket support.

jyeshe:
Closes #1010


lib/ae_mdw_web/websocket/broadcaster.ex
```diff
@@ -172,7 +160,9 @@ defmodule AeMdwWeb.Websocket.Broadcaster do
   end
 
   defp broadcast(channel, msg) do
-    AeMdwWeb.Websocket.SocketHandler.channel_broadcast(channel, msg)
```
thepiwo, line 175: wasn't that way simpler before in only relaying the message once, or was the other action "implicit"?

lib/ae_mdw_web/websocket/subscriptions.ex
```diff
@@ -0,0 +1,169 @@
...
+
+  require Ex2ms
+
+  @type channel :: <<_::256>> | String.t()
+  @subs_pids :subs_pids
+  @subs_channel_pid :subs_channel_pid
+  @subs_pid_channel :subs_pid_channel
+  @eot :"$end_of_table"
+
+  @known_channels ["KeyBlocks", "MicroBlocks", "Transactions"]
```
thepiwo, line 21: why is this needed, will subscribing to single accounts/contracts still work?

lib/ae_mdw_web/websocket/broadcaster.ex
jyeshe: If you look to the Riverside code it relays it once because it has the concept of channels. From the previous commit after removing riverside it needs to relay the message multiple times (Phoenix websocket is just a transport):
https://github.com/aeternity/ae_mdw/blob/4b69441dfeb5425ec63bef8e63d1c44dfa90fd42/lib/ae_mdw_web/websocket/socket_handler.ex#L37-L41

Clarifying a bit more, the difference here is that all ets/channel code that was scattered throughout 3 modules were delegated to a specific module. The `AeMdwWeb.Websocket.Subscriptions` was created for this otherwise the `SocketHandler` would continue having protocol handling code mixed with subscription.

This way the semantics of latest `broadcast` would be:
```
    channel 
    |> Subscriptions.subscribers() # for all subscribers of this channel
    |> Enum.each(&SocketHandler.send(&1, msg)) # send for each one of them
```

thepiwo: thanks for the clarification

lib/ae_mdw_web/websocket/subscriptions.ex
thepiwo: thanks for the clarification

thepiwo (approved):

sborrazas (approved):

PR #1037, opened 2022-11-23
feat: add /names/:id/claims endpoint

sborrazas:


lib/ae_mdw/names.ex
```diff
@@ -504,6 +559,39 @@ defmodule AeMdw.Names do
     |> Enum.map(&render_name_info(state, &1, opts))
   end
 
+  defp render_claim(state, plain_name, {{height, _mbi} = block_index, txi}) do
+    tx_hash = Txs.txi_to_hash(state, txi)
+    block_hash = Blocks.block_index_to_hash(state, block_index)
+
+    claim_aetx =
+      case Db.get_tx_data(tx_hash) do
```
jyeshe, line 567: Looks like an opportunity for reuse creating a function to fetch the tx:
https://github.com/aeternity/ae_mdw/blob/master/lib/ae_mdw/activities.ex#L475-L495
https://github.com/aeternity/ae_mdw/blob/master/lib/ae_mdw/db/name.ex#L173-L200 (here instead comparing the hash would compare the name as well and return the tx)

lib/ae_mdw/names.ex
```diff
@@ -504,6 +559,39 @@ defmodule AeMdw.Names do
...
+          :aetx_sign.tx(signed_tx)
+
+        {_block_hash, :contract_call_tx, _signed_tx, _tx_rec} ->
+          state
+          |> Contracts.fetch_int_contract_calls(txi, "AENS.claim")
+          |> Enum.find_value(fn Model.int_contract_call(tx: aetx) ->
+            {:name_claim_tx, tx} = :aetx.specialize_type(aetx)
+            name = :aens_claim_tx.name(tx)
+
+            plain_name == name
```
jyeshe, line 578: It would return `true` instead of the `tx`. May we have also a coverage for internal claim?

lib/ae_mdw/names.ex
```diff
@@ -535,4 +623,34 @@ defmodule AeMdw.Names do
...
+
+  defp serialize_nested_cursor(nil), do: nil
+
+  defp serialize_nested_cursor({{{height, mbi}, txi}, is_reverse?}),
+    do: {"#{height}-#{mbi}-#{txi}", is_reverse?}
+
+  defp deserialize_nested_cursor(nil), do: nil
+
+  defp deserialize_nested_cursor(cursor_bin) do
+    case Regex.run(~r/\A(\d+)-(\d+)-(\d+)\z/, cursor_bin, capture: :all_but_first) do
```
jyeshe, line 648: :+1:

jyeshe (changes requested):

lib/ae_mdw/names.ex
sborrazas: Good find, thanks. Done

thepiwo (approved):

lib/ae_mdw/names.ex
jyeshe: Thanks!

jyeshe (approved):

PR #1041, opened 2022-11-24
chore: remove unnecessary smart_record dependency

sborrazas:


jyeshe (approved):
I believe `SmartRecord` was used to avoid making 3 copies of the oracle record while rendering it. 

Approving considering the oracle formats are gas limited and there will be a profiling that can detect mem peak when oracle use increases.

Issue #195, opened 2021-06-17
improve docker builds


marc0olo:
think there are things that can be improved:

- [x] image size (node image uses ~145mb whereas the mdw including the node requires ~555mb)
- [x] file structure (I would appreciate if it reflects a similar folder structure as the node image at least for the node related stuff)
- [x] specify a specific user and that the containers of the image aren't running with root

ideally we have a more or less generic approach that can be applied to other plugins that make use of https://github.com/aeternity/ae_plugin

sborrazas:
Additionally, we need to build the docker image using [Elixir releases](https://hexdocs.pm/mix/Mix.Tasks.Release.html) which will leave out the overhead from downloading the libraries repos. It will also optimize the image for faster startup.

jyeshe:
Done after:
- #1190 
- #1197 
- #1199

PR #1043, opened 2022-11-25
ci: publish docker image on release

venimus:
Publish releases on dockehub

### DIFF
--- a/.github/workflows/dockerhub-release.yml
+++ b/.github/workflows/dockerhub-release.yml
@@ -1,0 +1,52 @@
+name: Dockerhub Release Pipeline
+
+on:
+  release:
+    types: [created]
+
+env:
+  ENV: "prd"
+  APP: "ae_mdw"
+
+jobs:
+  dockerhub:
+    name: "Publish Release to Dockehub"
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 100
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v2
+        # Only worked for us with this option on
+        with:
+          install: true
+
+      - name: Login to Dockerhub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USER }}
+          password: ${{ secrets.DOCKERHUB_PASS }}
+
+      - name: Extract metadata for docker
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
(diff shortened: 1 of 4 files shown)


thepiwo:
the need for network_id during build is being removed here: https://github.com/aeternity/ae_mdw/pull/1045

thepiwo (approved):

.github/workflows/dockerhub-release.yml
```diff
@@ -0,0 +1,51 @@
+name: Dockerhub Release Pipeline
+
+on:
+  release:
+    types: [created]
+
+env:
+  ENV: "prd"
```
sborrazas, line 8: On the app config we're calling the prod environment "prod". Perhaps it would make sense to name it the same way?

marc0olo:
@venimus what's the tagging pattern now? do we include somehow both, the node and mdw version?

venimus:
for release: <version> and latest
for dev: 'master' (for latest master, referred in gh-actions as "edge"), pr-<pr-number> latest commit in a PR, and <sha> for pr-commits
Node is downloaded when the container is built the version is specified in the mdw config. If you want to bump it you have to release a new mdw version (which is what makes sense)

venimus:
for release: `<version>` and `latest`
for dev: `master` (for latest master, referred in gh-actions as "edge"), `pr-<number>` latest commit in a PR, and for pr-commits `<sha>`
Node is downloaded when the container is built the version is specified in the mdw config. If you want to bump it you have to release a new mdw version (which is what makes sense)

.github/workflows/dockerhub-release.yml
venimus: it makes sense, but it is not directly used, not sure if would have any impact on the build process. I think it should be removed anyway

jyeshe: We would have some work to do on the mdw in regards to prod config and releases. It's in the scope of #195

marc0olo:
ok I see, is there any possibility to also make a tag for the node version in a meaningful way to know what node version the mdw runs with?

or do we consider a release of the mdw always to be tight to a specific node version? @thepiwo @sborrazas @jyeshe

thepiwo:
any docker build is tied to a node version

marc0olo:
so, do we want to indicate the node version somehow in the official release of the image or not? personally I would like to have that indicator somehow.

thepiwo:
I don't mind, for most node versions its not as we have a choice

venimus:
I guess it might be included in README or release notes, though the later would probably be indicated if there is a commit that is picked up from `release-please`. Tagging the docker container would overcomplicate things i guess. Git tags could be used too, but that would mean to "move"/reuse tags which usually causes problems with CI/CD (and when pulling in general)

sborrazas (approved):

PR #1045, opened 2022-11-25
chore: set network_id as runtime config

sborrazas:


thepiwo:
in my setup I have been running without setting this in the past, both mainnet and testnet, what is it needed for anyways?

sborrazas:
> in my setup I have been running without setting this in the past, both mainnet and testnet, what is it needed for anyways?

We need that config for this reason: https://github.com/aeternity/ae_mdw/blob/master/lib/ae_mdw/aexn_contracts.ex#L159, but I'm not really sure why we can't use the core fork_network value instead. 

I'll look into it

thepiwo (approved):

jyeshe:
> > in my setup I have been running without setting this in the past, both mainnet and testnet, what is it needed for anyways?
> 
> We need that config for this reason: https://github.com/aeternity/ae_mdw/blob/master/lib/ae_mdw/aexn_contracts.ex#L159, but I'm not really sure why we can't use the core fork_network value instead.
> 
> I'll look into it

It's an old config needed for https://github.com/aeternity/ae_mdw/pull/523. I am not using ad-hoc images anymore.

PR #1052, opened 2022-11-30
  feat: include tx_hash on NameClaim activity

sborrazas:


thepiwo (approved):
can we discuss a way were these model changes are consistent across endpoints?

sborrazas:
> can we discuss a way were these model changes are consistent across endpoints?

this is very much related to the `unify reference of transactions internally` feature. I think these two fields we're adding here (`tx` and `tx_hash`) are going to be present anyway. but when implementing that feature we should discuss ways in which we can expose whether the `tx_hash` points to a contract call tx or an actual name claim

test/ae_mdw_web/controllers/activities_controller_test.exs
```diff
@@ -767,29 +767,33 @@ defmodule AeMdwWeb.ActivitiesControllerTest do
                  "height" => ^height2,
                  "type" => "NameClaimEvent",
                  "payload" => %{
```
jyeshe, line 769: Could we have also the other two added fields, specially the `tx_hash`?

test/ae_mdw_web/controllers/activities_controller_test.exs
```diff
@@ -767,29 +767,35 @@ defmodule AeMdwWeb.ActivitiesControllerTest do
...
-                   "account_id" => ^account,
-                   "fee" => 333_333,
-                   "name" => ^plain_name,
-                   "name_fee" => 33333,
-                   "name_salt" => 3333,
-                   "nonce" => 333,
-                   "ttl" => 3_333_333,
-                   "type" => "NameClaimTx"
+                   "tx_hash" => "th_2CdVYuqtpcjoshDw2otjbLEyj8SpxjMP9MCgpn1oU9zGaqvUn4",
```
jyeshe, line 770: can we pattern match please to avoid looking up for the hash?

jyeshe: The `micro_time` could be easily tested the same way please.

sborrazas: Done

jyeshe (approved):

PR #1054, opened 2022-11-30
chore: put nft contract limits in a map

jyeshe:
Returns nft contract limits under `limits` object in way that if there is no limit only one field is returned.

### DIFF
--- a/lib/ae_mdw/aex141.ex
+++ b/lib/ae_mdw/aex141.ex
@@ -114,7 +114,7 @@ defmodule AeMdw.Aex141 do
     end
   end
 
-  @spec fetch_limits(State.t(), pubkey()) :: limits()
+  @spec fetch_limits(State.t(), pubkey()) :: limits() | nil
   def fetch_limits(state, contract_pk) do
     Model.nft_contract_limits(
       token_limit: token_limit,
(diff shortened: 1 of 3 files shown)


lib/ae_mdw/aex141.ex
```diff
@@ -127,12 +127,14 @@ defmodule AeMdw.Aex141 do
         {:ok, m_limits} -> m_limits
       end
 
-    %{
-      token_limit: token_limit,
-      template_limit: template_limit,
-      limit_txi: txi,
-      limit_log_idx: log_idx
-    }
+    if token_limit != nil or template_limit != nil do
```
jyeshe, line 130: @marc0olo this is the condition proposed to show the map, otherwise it shows `limits: nil`

thepiwo (approved):

sborrazas (approved):

lib/ae_mdw/aex141.ex
marc0olo: I was just thinking about explicitness vs. implicitness when it comes to differentiation between the `mintable` extension and the `mintable_templates` extension.

from my perspective there is always `token_limit` **OR** `template_limit` as those limit types cannot/should not be mixed among these extensions.

for me this is generally fine. if we'd want to be more explicit we'd add extension related attributes in their own section instead of delivering these values in a generic way without relation to any extension. that is what I was thinking of. but I don't have a strong opinion.

jyeshe: it's fine, thanks. it's only matter of indexing using the same record to keep one limit or the other.

marc0olo: agree, doesn't make sense to split the limit info into different sections. too little differentiation. domain knowledge is required anyway 😅

marc0olo (approved):

PR #1059, opened 2022-12-02
chore: allow network id for local environment

jyeshe:
## What

Enable setting network id for dev and test environments

## Why

Required when running locally with `make shell` to allow using `uat_backup_v1_full_latest.tar.zst`

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -1,5 +1,7 @@
 import Config
 
+config :aecore, network_id: System.get_env("NETWORK_ID", "ae_mainnet")
+
 config :ae_mdw, AeMdwWeb.Endpoint,
   http: [port: 4000],
   debug_errors: true,
--- a/config/test.exs
+++ b/config/test.exs
@@ -8,6 +8,8 @@ if System.get_env("INTEGRATION_TEST") != "1" do
   config :ae_mdw, AeMdw.Db.RocksDb, data_dir: "test_data.db"
 end
 
+config :aecore, network_id: System.get_env("NETWORK_ID", "ae_mainnet")
+
 # HTTP
 config :ae_mdw, AeMdwWeb.Endpoint,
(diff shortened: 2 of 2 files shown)


thepiwo (approved):

sborrazas:
What's the reason for requiring this value? Wouldn't setting it at the `aeternity.yaml` level suffice?

Right now, locally I'm able to set:
```
fork_management:
    network_id: ae_uat
```
on my `docker/aeternity.yaml`, start it with `iex --sname ... -S mix phx.server` and I get testnet blocks synced

jyeshe:
> What's the reason for requiring this value? Wouldn't setting it at the `aeternity.yaml` level suffice?
> 
> Right now, locally I'm able to set:
> 
> ```
> fork_management:
>     network_id: ae_uat
> ```
> 
> on my `docker/aeternity.yaml`, start it with `iex --sname ... -S mix phx.server` and I get testnet blocks synced

The `docker/aeternity.yaml` is not considered by the shell outside the docker image.

sborrazas (approved):
Since the `fork_network` value overrides the `:network_id` config when `:network_id` is `nil` it won't really affect regular docker users. But it might be confusing to have both the `aeternity.yaml` and the `NETWORK_ID` config, specially because when they are different the result might be unexpected. 

Something to keep in mind in the future, thanks for clarifying

jyeshe:
> Since the `fork_network` value overrides the `:network_id` config when `:network_id` is `nil` it won't really affect regular docker users. But it might be confusing to have both the `aeternity.yaml` and the `NETWORK_ID` config, specially because when they are different the result might be unexpected.
> 
> Something to keep in mind in the future, thanks for clarifying

no problem, the variable is not an argument (ARG) anymore to the docker enviroment, it's only available locally for dev.

PR #1062, opened 2022-12-02
fix: name of sender parameter of get_aex9_pair_transfers

davidyuk:
the parameter with `recipient` name is defined below


thepiwo (approved):

sborrazas (changes requested):
All v2 swagger config needs to be edited [here](https://github.com/aeternity/ae_mdw/tree/master/docs/swagger_v2) first. and then autogenerated with `./scripts/swagger-docs.py >priv/static/swagger/swagger_v2.yaml`

As per these instructions https://github.com/aeternity/ae_mdw#auto-generated-documentation

PR #1068, opened 2022-12-05
refactor: log mutations and error msg when database commit fails

jyeshe:


sborrazas:
I would try to avoid error handling if it's going to fail anyway. The error we're currently having also happens locally, there's no need to log mutations on the server. Perhaps we can leave this branch if anyone needs to run it locally?

jyeshe:
> I would try to avoid error handling if it's going to fail anyway. The error we're currently having also happens locally, there's no need to log mutations on the server. Perhaps we can leave this branch if anyone needs to run it locally?

There is already an error handling, it's only bubbled up.
image

I think it also makes the code more clear about what a database transaction is by:
- Instantiating a `TxnDbStore` like any other `Store` with `new()`
- Avoiding a high order function and a template function that passes an input to the high order function
- Avoiding mixed closures between the anonymous high order and `commit` functions

More importantly it would be always be available to any developer debugging a transaction database commit failure and if an error happens on the server it's fully logged to be able to compare with the local result.

sborrazas:
> > I would try to avoid error handling if it's going to fail anyway. The error we're currently having also happens locally, there's no need to log mutations on the server. Perhaps we can leave this branch if anyone needs to run it locally?
> 
> There is already an error handling, it's only bubbled up. image
> 
> I think it also makes the code more clear about what a database transaction is by:
> 
> * Instantiating a `TxnDbStore` like any other `Store` with `new()`
> * Avoiding a high order function and a template function that passes an input to the high order function
> * Avoiding mixed closures between the anonymous high order and `commit` functions
> 
> More importantly it would be always be available to any developer debugging a transaction database commit failure and if an error happens on the server it's fully logged to be able to compare with the local result.

The higher-order function allows for mis-using transactions in general, which is the benefit it provides, by doing it in a functional way. We don't need to remove this benefit in order to log errors, we can keep both:
```elixir
case TxnDbStore.transaction(&exec_muts(&1, muts)) do
  :ok -> ...
  :error ->
    Log.error(muts)
    raise "failed  #{length(muts)}..." # ...
end
```

jyeshe:
> The higher-order function allows for mis-using transactions in general, which is the benefit it provides, by doing it in a functional way. We don't need to remove this benefit in order to log errors, we can keep both:
> 
> ```elixir
> case TxnDbStore.transaction(&exec_muts(&1, muts)) do
>   :ok -> ...
>   :error ->
>     Log.error(muts)
>     raise "failed  #{length(muts)}..." # ...
> end
> ```

The point is: writing a high order function and a template function when **two lines** of code solves the problem is an overkill solution. Only by stating it does in a functional way doesn't make other alternatives less functional. Actually the current code is clearly not functional by using a variable without receiving it as a parameter.

PR #1071, opened 2022-12-08
chore: index only event-based aex9 balances

### LINKED ISSUE #969: Index AEX-N primarly using events

jyeshe:
Closes #969

### DIFF
--- a/lib/ae_mdw/aex9.ex
+++ b/lib/ae_mdw/aex9.ex
@@ -47,10 +47,10 @@ defmodule AeMdw.Aex9 do
       state = get_store_state(state, contract_pk)
 
       state
-      |> Collection.stream(Model.Aex9Balance, {contract_pk, <<>>})
+      |> Collection.stream(Model.Aex9EventBalance, {contract_pk, <<>>})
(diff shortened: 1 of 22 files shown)


thepiwo (approved):
Always love to see more deletions than additions in a refactor, does this also resolve the issues we had in migration?

sborrazas (approved):

jyeshe:
> Always love to see more deletions than additions in a refactor, does this also resolve the issues we had in migration?

Yes, instead of adding synchronization for the optimistic lock conflict, all dry-run code related to contract calls could be removed since it's not a requirement anymore. More kind a `chore` task, thanks for asking.

PR #1072, opened 2022-12-09
feat: add v2 websocket implementation

sborrazas:
To connect to the v2 websocket a param needs to be passsed when establishing the WS connection: `/websocket?version=2`

These new websocket messages will:

* Contain the additional block/tx information that the middleware already provides in the `/v2/key-blocks/:height`, `/v2/micro-blocks/:hash` and `/v2/txs/:hash` endpoints.
* Order the blocks emitted in the order from the chain: for each generation the key-block is emitted before the generation micro-blocks.

### DIFF
--- a/lib/ae_mdw_web/websocket/broadcaster.ex
+++ b/lib/ae_mdw_web/websocket/broadcaster.ex
@@ -91,7 +91,7 @@ defmodule AeMdwWeb.Websocket.Broadcaster do
   defp do_broadcast_key_block(block, :v2, :mdw) do
     header = :aec_blocks.to_header(block)
     height = :aec_headers.height(header)
-    state = State.new()
+    state = State.mem_state()
 
     case Blocks.fetch_key_block(state, "#{height}") do
       {:ok, block} ->
@@ -136,7 +136,7 @@ defmodule AeMdwWeb.Websocket.Broadcaster do
   defp do_broadcast_micro_block(block, :v2, :mdw) do
     header = :aec_blocks.to_header(block)
     {:ok, hash} = :aec_headers.hash_header(header)
-    state = State.new()
+    state = State.mem_state()
 
     case Blocks.fetch_micro_block(state, hash) do
       {:ok, block} ->
@@ -169,7 +169,7 @@ defmodule AeMdwWeb.Websocket.Broadcaster do
   end
 
   defp do_broadcast_txs(block, :v2, :mdw) do
-    state = State.new()
+    state = State.mem_state()
 
     block
     |> :aec_blocks.txs()
@@ -177,13 +177,13 @@ defmodule AeMdwWeb.Websocket.Broadcaster do
       tx_hash = :aetx_sign.hash(tx)
 
       case Txs.fetch(state, tx_hash, true) do
-        {:ok, tx} ->
-          mdw_msg = encode_message(tx, "Transactions", :mdw)
+        {:ok, mdw_tx} ->
+          mdw_msg = encode_message(mdw_tx, "Transactions", :mdw)
           broadcast("Transactions", :v2, mdw_msg)
 
           tx
           |> get_ids_from_tx()
-          |> Enum.each(&broadcast(&1, :v2, encode_message(tx, "Object", :mdw)))
+          |> Enum.each(&broadcast(&1, :v2, encode_message(mdw_tx, "Object", :mdw)))
 
         {:error, _reason} ->
           :ok
--- a/lib/ae_mdw_web/websocket/broadcaster.ex
+++ b/lib/ae_mdw_web/websocket/broadcaster.ex
@@ -4,16 +4,21 @@ defmodule AeMdwWeb.Websocket.Broadcaster do
   """
   use GenServer
 
+  alias AeMdw.Blocks
+  alias AeMdw.Db.State
   alias AeMdw.EtsCache
+  alias AeMdw.Node.Db
+  alias AeMdw.Txs
   alias AeMdwWeb.Websocket.Subscriptions
   alias AeMdwWeb.Websocket.SocketHandler
 
   require Ex2ms
 
-  @dialyzer {:no_return, broadcast: 2}
-
   @hashes_table :broadcast_hashes
   @expiration_minutes 120
+
+  @typep source() :: :node | :mdw
+  @typep version() :: Subscriptions.version()
 
   @spec start_link(any()) :: GenServer.on_start()
   def start_link(_arg), do: GenServer.start_link(__MODULE__, :ok, name: __MODULE__)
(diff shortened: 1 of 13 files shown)


sborrazas:
@thepiwo that's a good idea, would like to implement it that way. Thoughts @jyeshe ?

lib/ae_mdw_web/websocket/broadcaster.ex
```diff
@@ -24,72 +29,81 @@ defmodule AeMdwWeb.Websocket.Broadcaster do
...
-
+  def handle_info(_msg, state) do
     {:noreply, state}
   end
 
   #
   # Private functions
   #
-  defp do_broadcast_key_block(block, source) do
+  defp do_broadcast_key_block(block, :v2, :mdw) do
```
jyeshe, line 91: would invert using `:v1` in the parameter with specific implementation and leave default (without pattern matching) for `:v2` to allow `:v3` continue from `:v2`

lib/ae_mdw_web/websocket/broadcaster.ex
```diff
@@ -111,16 +125,30 @@ defmodule AeMdwWeb.Websocket.Broadcaster do
...
     end
   end
 
-  defp do_broadcast_micro_block(block, source) do
+  defp do_broadcast_micro_block(block, :v2, :mdw) do
+    header = :aec_blocks.to_header(block)
+    {:ok, hash} = :aec_headers.hash_header(header)
+    state = State.new()
+
+    case Blocks.fetch_micro_block(state, hash) do
```
jyeshe, line 141: Needs to be adapted to work for broadcasting in-memory sync.

sborrazas:
> 

I don't think this will be a problem, since I doubt we even make it to more than 10-20 connections total, that `10_000` connections limit can be reduced to `5_000` each and I don't think it would affect the service

lib/ae_mdw_web/websocket/subscriptions.ex
```diff
@@ -130,40 +134,48 @@ defmodule AeMdwWeb.Websocket.Subscriptions do
...
     end
   end
 
-  defp validate_unsubscribe(pid, channel) do
-    with {:ok, channel_key} <- channel_key(channel),
-         true <- exists?(pid, channel) do
-      {:ok, channel_key}
+  defp validate_unsubscribe(pid, version, channel) do
+    with {:ok, channel_key} <- channel_key(version, channel),
+         {:ok, channel} <- match(pid, channel_key) do
```
jyeshe, line 155: The `channel` returned by `match` is exactly the same value as the input of the unsubscribe. So since it's already available in the function the `exists?` function is enough (keeping the code simpler).

lib/ae_mdw_web/websocket/subscriptions.ex
```diff
@@ -48,41 +50,43 @@ defmodule AeMdwWeb.Websocket.Subscriptions do
...
     end
   end
 
-  @spec unsubscribe(pid(), channel()) ::
+  @spec unsubscribe(pid(), version(), channel()) ::
           {:ok, [channel()]} | {:error, :invalid_channel | :not_subscribed}
-  def unsubscribe(pid, channel) do
-    with {:ok, channel_key} <- validate_unsubscribe(pid, channel) do
+  def unsubscribe(pid, version, channel) do
+    with {:ok, channel_key, channel} <- validate_unsubscribe(pid, version, channel) do
```
jyeshe, line 68: The channel returned by `validate_unsubscribe` is the same value as the parameter (additional out = in).

lib/ae_mdw_web/websocket/subscriptions.ex
```diff
@@ -48,41 +50,43 @@ defmodule AeMdwWeb.Websocket.Subscriptions do
...
       :ets.delete_object(@subs_channel_pid, {channel_key, pid})
-      :ets.delete_object(@subs_pid_channel, {pid, channel})
+      :ets.delete_object(@subs_pid_channel, {pid, {channel_key, channel}})
```
jyeshe, line 70: In `subscribed_channels` the return of `@subs_pid_channel` it would have a different format (includes the version):
`[{{_version, _channel_key}, channel}]`

Seems a good moment for extending the test cases to cover it.

jyeshe (changes requested):

jyeshe:
> I don't think this will be a problem, since I doubt we even make it to more than 10-20 connections total, that `10_000` connections limit can be reduced to `5_000` each and I don't think it would affect the service
> 
> edit: forgot to mention @jyeshe @thepiwo

These limits should be considered for security reasons, not for the actual connections of known clients. If this a concern, it becomes 50% precise as it is with a single websocket endpoint. If the concern is a number closer to the internal clients, it should be part of the PR otherwise causes this distortion on the limits.

lib/ae_mdw_web/websocket/subscriptions.ex
sborrazas: The problem is the variable down below is named incorrectly, it should have been `{_channel_key, channel}` instead, since we don't really need to pattern match on the `{_version, _channel}` tuple. Changing it

lib/ae_mdw_web/websocket/subscriptions.ex
sborrazas: It might not be the same because right now the subscribe/unsubscribe allows you to subscribe to a channel like `ak_wTPFpksUJFjjntonTvwK4LJvDw11DPma7kZBneKbumb8yPeFq` and unsubscribe  to `ok_wTPFpksUJFjjntonTvwK4LJvDw11DPma7kZBneKbumb8yPeFq`. Same address, different type

jyeshe:
> > I don't think this will be a problem, since I doubt we even make it to more than 10-20 connections total, that `10_000` connections limit can be reduced to `5_000` each and I don't think it would affect the service
> > edit: forgot to mention @jyeshe @thepiwo
> 
> These limits should be considered for security reasons, not for the actual connections of known clients. If this a concern, it becomes 50% precise as it is with a single websocket endpoint. If the concern is a number closer to the internal clients, it should be part of the PR otherwise causes this distortion on the limits.

Alternatives:
b) remove the limit if it's not important
c) if a rough 50% or 200% value is acceptable add a note on the `Endpoint` config explaining it is split with other endpoints.
d) open another issue to investigate it

lib/ae_mdw_web/websocket/subscriptions.ex
jyeshe: It would be the same because when there is a subscription for `ak_wTPFpksUJFjjntonTvwK4LJvDw11DPma7kZBneKbumb8yPeFq` it is an alias for the binary that is matched with the value in the transactions (that is not the encoded pubkey). In this sense we might consider that `ak_wTPFpksUJFjjntonTvwK4LJvDw11DPma7kZBneKbumb8yPeFq` and `ok_wTPFpksUJFjjntonTvwK4LJvDw11DPma7kZBneKbumb8yPeFq` would be aliases for same subscription.

lib/ae_mdw_web/websocket/broadcaster.ex
sborrazas: The problem with this is that there's 4 possible outcomes (`node/v1`, `mdw/v1`, `node/v2` and `mdw/v2`) and the representation only differs with `mdw/v2` which is why it has a special clause

lib/ae_mdw_web/websocket/subscriptions.ex
jyeshe: Besides the naming the number of elements returned as well because what is returned is the value of the record without the key:
```
    @subs_pid_channel
    |> :ets.match({pid, :"$1"})
```
`[{_channel_key, channel}]` instead of `[{{_version, _channel_key}, channel}]` and since the channel key can be easily derived wouldn't store it but it's ok if you really want to keep in memory for `unsubscribe_all`.

lib/ae_mdw_web/websocket/subscriptions.ex
```diff
@@ -130,40 +134,48 @@ defmodule AeMdwWeb.Websocket.Subscriptions do
...
       {:error, _reason} -> {:error, :invalid_channel}
-      false -> {:error, :not_subscribed}
+      :none -> {:error, :not_subscribed}
     end
   end
 
-  defp channel_key(channel) when channel in @known_channels, do: {:ok, channel}
-  defp channel_key(channel), do: Validate.id(channel)
+  # :ets.delete_object(@subs_channel_pid, {channel_key, pid})
+  # :ets.delete_object(@subs_pid_channel, {pid, {channel_key, channel}})
```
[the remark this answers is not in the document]
sborrazas: Woops, done

lib/ae_mdw_web/websocket/subscriptions.ex
sborrazas: The channel out value is not always the same as the in value, since it will always return the value of the first subscription

lib/ae_mdw_web/websocket/subscriptions.ex
sborrazas: Exactly, but in order to remove the value from ets we need to know the exact address in which the user was subscribed to

sborrazas:
> > > I don't think this will be a problem, since I doubt we even make it to more than 10-20 connections total, that `10_000` connections limit can be reduced to `5_000` each and I don't think it would affect the service
> > > edit: forgot to mention @jyeshe @thepiwo
> > 
> > 
> > These limits should be considered for security reasons, not for the actual connections of known clients. If this a concern, it becomes 50% precise as it is with a single websocket endpoint. If the concern is a number closer to the internal clients, it should be part of the PR otherwise causes this distortion on the limits.
> 
> Alternatives: b) remove the limit if it's not important c) if a rough 50% or 200% value is acceptable add a note on the `Endpoint` config explaining it is split with other endpoints. d) open another issue to investigate it

The previous `transmission`, `max_connection`, etc values in the config weren't really being used, since these were values that were used exclusively by riverside, not phoenix. I think your `b)` alternative will just do for now, if we start seeing issues with the server performance we can always start looking into it further

lib/ae_mdw_web/websocket/subscriptions.ex
jyeshe: There are only two representations for a channel:
A) encoded String, received as a parameter on `unsubscribe` and passed to `validate_unsubscribe`
B) as a raw binary, that is a channel key

Since it already returns the channel key and `validate_unsubscribe` receives the channel as a parameter there is no need to add it to the return of `validate_unsubscribe`. The `validate_subscribe` can be improved to avoid double subscriptions for same object `ak_wTPFpksUJFjjntonTvwK4LJvDw11DPma7kZBneKbumb8yPeFq` and `ok_wTPFpksUJFjjntonTvwK4LJvDw11DPma7kZBneKbumb8yPeFq` (until `ok_wTPFpksUJFjjntonTvwK4LJvDw11DPma7kZBneKbumb8yPeFq` is handled only for Oracles which should be part of another PR please if you wanna work on that right away)

lib/ae_mdw_web/websocket/subscriptions.ex
jyeshe: The address is actual the binary which is the same for `ak_wTPFpksUJFjjntonTvwK4LJvDw11DPma7kZBneKbumb8yPeFq` and `ok_wTPFpksUJFjjntonTvwK4LJvDw11DPma7kZBneKbumb8yPeFq`

jyeshe: The address is actuallly the binary which is the same for `ak_wTPFpksUJFjjntonTvwK4LJvDw11DPma7kZBneKbumb8yPeFq` and `ok_wTPFpksUJFjjntonTvwK4LJvDw11DPma7kZBneKbumb8yPeFq`

lib/ae_mdw_web/websocket/broadcaster.ex
jyeshe: It is in regards to a possible and likely `:v3` where the key block might be the same as the `:v2`.

If the `:v1` is the pattern matching, `:v3` would go to the default as `:v2`

lib/ae_mdw_web/websocket/subscriptions.ex
sborrazas: The subs_pid_channel table stores values such as:
```
{{<#PID1>, {:v1, <<1, 2, 3, ..>>}}, "ak_123.."} 
```
since the key is the binary representation you can't really subscribe to the same pubkey twice, it will return an error.

When unsubscribing from the record, though, I need both the binary and encoded representation, which is why `validate_unsubscribe` returns the original encoded address the user subscribed to

If this is not what you had in mind, would you mind posting the full-code of what you would like it to be? I can't seem to understand otherwise

jyeshe: > The channel out value is not always the same as the in value, since it will always return the value of the first subscription

It was mentioned that it would return the first but the `validate_subscribe` on [master](https://github.com/aeternity/ae_mdw/blob/7d671009dd6d58cf658eedef41210edbf9ecbe7b/lib/ae_mdw_web/websocket/subscriptions.ex#L140-L142) already doesn't allow `ok_wTPFpksUJFjjntonTvwK4LJvDw11DPma7kZBneKbumb8yPeFq` to be subscribed when `ak_wTPFpksUJFjjntonTvwK4LJvDw11DPma7kZBneKbumb8yPeFq` already is. 

So the first entry would be the only subscribed channel available as checked by:
https://github.com/aeternity/ae_mdw/blob/7d671009dd6d58cf658eedef41210edbf9ecbe7b/lib/ae_mdw_web/websocket/subscriptions.ex#L164

If the channel subscribed was `ak_wTPFpksUJFjjntonTvwK4LJvDw11DPma7kZBneKbumb8yPeFq` I think it's fair and expected that the `unsubscribe` works only with same string and not `ok_wTPFpksUJFjjntonTvwK4LJvDw11DPma7kZBneKbumb8yPeFq`. Following this behavior, the `validate_unsubscribe` doesn't need to return the `channel` because it's already the input.

jyeshe (changes requested):
Subscriptions code needs to be adapted only in regards to the version until `ak_wTPFpksUJFjjntonTvwK4LJvDw11DPma7kZBneKbumb8yPeFq` and `ok_wTPFpksUJFjjntonTvwK4LJvDw11DPma7kZBneKbumb8yPeFq` is handled differently by enhancing the `Object` subscriptions to be sensitive to transaction type.

sborrazas:
Made changes as requested

lib/ae_mdw_web/websocket/subscriptions.ex
```diff
@@ -48,41 +50,43 @@ defmodule AeMdwWeb.Websocket.Subscriptions do
...
+    |> :ets.match({channel_key, :"$1"})
     |> List.flatten()
   end
 
   @spec subscribed_channels(pid()) :: [channel()]
   def subscribed_channels(pid) do
     @subs_pid_channel
     |> :ets.match({pid, :"$1"})
-    |> List.flatten()
+    |> Enum.map(fn [{{_version, _channel_key}, channel}] -> channel end)
```
jyeshe, line 89: This line works but it is confusing because the version is part of the channel_key.

lib/ae_mdw_web/websocket/subscriptions.ex
jyeshe: I see now the number of elements is ok, the confusion is because the version is part of channel key and in this `Enum.map` they are separate as `[{{_version, _channel_key}, channel}]`

lib/ae_mdw_web/websocket/subscriptions.ex
jyeshe: @sborrazas I understand the confusion that is happening. It was due to the `match` function returning the channel while if `exists?` it just had to detect a match or not. Thanks going through it avoiding the return of the channel. 

Please find here the code requested with minimum changes needed:
https://github.com/aeternity/ae_mdw/commit/b13edb5fb1eac4d9ad4184d23b8261115b78456d
- `validate_unsubscribe` returns only the `channel_key`
- `exists?` depends only channel key
- `subscribed_channels` only flattens the list
- does not allow duplicate subscription of account and oracle with same pubkey (as you've fixed before, added only a test case)
- since oracle key returns `:already_subscribed` for same account pubkey, it allows unsubscribing it (same as above).

sborrazas:
Made changes as requested

jyeshe (approved):
Thanks for finding out the config and going through a shared solution :pray:

PR #1078, opened 2022-12-14
feat: track nft template edition supply 

jyeshe:
## What

- Adds endpoint `/aex141/:contract_id/templates/:template_id/tokens`
- Adds supply info to `/aex141/:contract_id/templates` edition response field

## Why

- Final request from @marc0olo to allow comparing the template edition limit to the actual supply (refs #629)

## Notes

Docs to be updated on a following PR

### DIFF
--- a/lib/ae_mdw/aex141.ex
+++ b/lib/ae_mdw/aex141.ex
@@ -9,8 +9,9 @@ defmodule AeMdw.Aex141 do
   alias AeMdw.Db.State
   alias AeMdw.Error
   alias AeMdw.Error.Input, as: ErrInput
-  alias AeMdw.Util
+  alias AeMdw.Stats
   alias AeMdw.Txs
+  alias AeMdw.Util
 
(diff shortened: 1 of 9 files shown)


lib/ae_mdw/aex141.ex
```diff
@@ -221,12 +267,41 @@ defmodule AeMdw.Aex141 do
...
+        tx_hash: encode(:tx_hash, tx_hash),
+        log_idx: log_idx
+      }
+    end)
+  end
+
+  defp render_template_edition(state, contract_pk, template_id, limit) do
+    stats_key = Stats.nft_template_tokens_key(contract_pk, template_id)
+
+    if nil != limit or :not_found != State.get(state, Model.Stat, stats_key) do
```
sborrazas, line 297: We can use `State.exists?/3` here

jyeshe: done, thanks

sborrazas (approved):

PR #1079, opened 2022-12-14
chore: add union of model-specific State typespecs

sborrazas:


lib/ae_mdw_web/views/aexn_view.ex
```diff
@@ -43,7 +43,7 @@ defmodule AeMdwWeb.AexnView do
            height: AeMdw.Blocks.height(),
            last_tx_hash: Txs.tx_hash(),
            last_log_idx: AeMdw.Contracts.log_idx(),
-           amount: integer()
+           amount: float()
```
jyeshe, line 46: It is an integer(), in ættos

lib/ae_mdw/db/origin.ex
```diff
@@ -45,7 +45,7 @@ defmodule AeMdw.Db.Origin do
     and created through core hard-forks. These contracts will have a negative
     txi where the number is the index of the preloaded contracts.
   """
-  @spec tx_index(State.t(), creation_txi_locator()) :: {:ok, Txs.txi() | -1} | :not_found
+  @spec tx_index(State.t(), creation_txi_locator()) :: {:ok, integer()} | :not_found
```
jyeshe, line 48: What would be the motivation behind if the Txs.txi() was used in so many places and the function returns a transaction index?

lib/ae_mdw_web/views/aexn_view.ex
sborrazas: Type checking is complaining because it's being specified as a `float()` on the `aex9_event_balance` record:
https://github.com/aeternity/ae_mdw/blob/1bedf79e4f95b0aeb2043a9614e128138c72fc29/lib/ae_mdw_web/views/aexn_view.ex#L100
https://github.com/aeternity/ae_mdw/blob/1bedf79e4f95b0aeb2043a9614e128138c72fc29/lib/ae_mdw/db/model.ex#L384

lib/ae_mdw/db/origin.ex
sborrazas: `tx_index` returns a negative integer when the contract is one of the preset contracts: https://github.com/aeternity/ae_mdw/blob/1bedf79e4f95b0aeb2043a9614e128138c72fc29/lib/ae_mdw/db/origin.ex#L55

jyeshe: Two alternatives to make it more clear from my perspective: 
a) if a negative index is accessible in the database (present somehow in `Model.tx`), then the `Txs.txi()` should include them.
b) if that's not the case, they should be returned in a tagged value instead of negative integer. I understand in this case it wouldn't be part of the scope of the PR but it's something for us to keep in mind when doing a refactoring.

sborrazas: When the returned value of `tx_index` is for example `-8`, it means that the contract is the 8th preset contract. I agree though that this isn't really friendly, we should def do something about it. But the intention of this PR isn't a refactor, but to include more helpful typespecs of the current (unchanged) implementation

jyeshe: Yes, totally fine to be fixed later.

jyeshe (approved):

PR #1080, opened 2022-12-14
fix: remove dockerization from ci tests run

sborrazas:
Fixes the following error:

Screen Shot 2022-12-14 at 12 05 46 PM


jyeshe (approved):
Approving as a temporary solution to unblock CI since it's important to fix it with docker in order to have tests including the image environment and setup.

Issue #1032, opened 2022-11-21
Add missing transaction details data for channels


michele-franchi:
Currently not all transactions data is returned from `/v2/txs/:transactionId` endpoint.

It would be useful to include the following data for these "Channel" transactions types to have a more detailed overview of these transaction types:

ChannelCreateTx:
- Sent by

ChannelCloseMutualTx:
- Responder
- Initiator

ChannelCloseSoloTx:
- Round
- Initiator
- Initiator amount
- Responder
- Responder amount

ChannelSlashTx:
- Round

ChannelSettleTx:
- Initiator
- Initiator amount

ChannelSnapshotSoloTx:
- Round
- Initiator
- Initiator amount
- Responder
- Responder amount

ChannelForceProgressTx:
- Initiator 
- Initiator amount
- Responder 
- Responder amount

sborrazas:
I've now added the `/v2/channels/:id` page for being able to retrieve most of this information. The round number updated on every snapshot and slash transaction is still missing. It will be added by @jyeshe in the future

jyeshe:
@michele-franchi I am closing this considering it done as the issue involving channels is for another reason. The fields that you will find missing are for the reasons:
- `ChannelCreateTx`: If there is no particular case missing the sender would be the initiator (after the creation, other transactions could be initiated by delegates). Actually the sender could be by someone `paying_for` the channel creation but it would be associated already in `/txs`. Free to open a specific issue if there is a specific requirement to make any additional relation with `paying_for`, `ga_meta` or other sender context. 
- `ChannelCloseSoloTx`, `ChannelSnapshotSoloTx` and `ChannelForceProgressTx`: The amounts are available only at microblock that might have multiple channel transactions in a way that could lead to imprecision if we put in these transactions. It was added a parameter to `/channels/:channel` though to allow querying the channel state including the amounts and reserve at a specific micro block.
-  `ChannelSettleTx`: please use initiator_amount_final

PR #1083, opened 2022-12-16
feat: add node details to channel page

jyeshe:
refs #1032

### DIFF
--- a/lib/ae_mdw/channels.ex
+++ b/lib/ae_mdw/channels.ex
@@ -63,7 +63,7 @@ defmodule AeMdw.Channels do
       type_count(state, :channel_settle_tx, from_txi, next_txi)
   end
 
-  @spec fetch_channel(state(), binary()) :: {:ok, channel()} | {:error, Error.t()}
+  @spec fetch_channel(state(), pubkey()) :: {:ok, channel()} | {:error, Error.t()}
   def fetch_channel(state, channel_pk) do
     case locate(state, channel_pk) do
       {:ok, channel, source} ->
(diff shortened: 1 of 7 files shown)


lib/ae_mdw/channels.ex
```diff
@@ -113,22 +120,37 @@ defmodule AeMdw.Channels do
...
+      responder: encode_account(responder_pk),
+      state_hash: encode(:state, state_hash),
       last_updated_height: last_updated_height,
       last_updated_tx_hash: tx_hash,
       last_updated_tx_type: tx_type,
       updates_count: length(updates),
       active: is_active?
     }
+
+    with true <- node_details?,
```
sborrazas, line 141: Is there a reason why we're unable to add this `get_channel_at_hash` information on the `/v2/channels` endpoint for each channel?

lib/ae_mdw/channels.ex
```diff
@@ -62,15 +64,17 @@ defmodule AeMdw.Channels do
   end
 
   @spec fetch_channel(state(), binary()) :: {:ok, channel()} | {:error, Error.t()}
```
sborrazas, line 66: We should change signature here to be `(state(), pubkey())`

lib/ae_mdw/channels.ex
jyeshe: It would be better after a profiling to check the impact on memory and performance since it needs to load multiple block state trees for different block hashes (one for each channel updated on different micro block).

lib/ae_mdw/channels.ex
jyeshe: done

sborrazas (approved):

lib/ae_mdw/channels.ex
jyeshe: It demands profiling to check the impact on memory and performance since it needs to load multiple block state trees for different block hashes (one for each channel updated on different micro block). On a `/v2/channels` refactoring would be a better moment for that.

PR #1084, opened 2022-12-19
ci: enable docker image tests

jyeshe:
Adds visibility to what happens running the middleware in a docker container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,10 +34,12 @@ RUN ${NODEDIR}/bin/aeternity check_config /home/aeternity/aeternity.yaml
 COPY config ./ae_mdw/config
 COPY lib ./ae_mdw/lib
 COPY priv ./ae_mdw/priv
+COPY test ./ae_mdw/test
 COPY mix.exs ae_mdw
 COPY mix.lock ae_mdw
 COPY Makefile ae_mdw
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -34,10 +34,12 @@ RUN ${NODEDIR}/bin/aeternity check_config /home/aeternity/aeternity.yaml
 COPY config ./ae_mdw/config
 COPY lib ./ae_mdw/lib
 COPY priv ./ae_mdw/priv
+COPY test ./ae_mdw/test
 COPY mix.exs ae_mdw
 COPY mix.lock ae_mdw
 COPY Makefile ae_mdw
 COPY docker/entrypoint.sh ae_mdw/entrypoint.sh
+COPY scripts/test.sh ae_mdw/test.sh
```
sborrazas, line 42: Do we want to add tests to the published docker image that users are going to run on their production servers? Maybe we can figure out a way to run these tests without having to add them by using shared volumes specified in the docker-compose

jyeshe: Wouldn't worry with the test files for now because if a user runs the `docker-compose up` it already duplicates every file, including the `/test` under the `/app`. Nevertheless added it as volumes, so the image itself won' have the files.

sborrazas (approved):

PR #1086, opened 2022-12-19
chore(ci): separate docker compose by env

jyeshe:


docker-compose.yml
```diff
@@ -1,24 +1,20 @@
...
       dockerfile: ./Dockerfile
-    ports:
-      - "4000:4000" #MDW's default port
-      - "4001:4001" #MDW's websocket default port
-      - "3113:3113" #Node's default internal API port
-      - "3013:3013" #Node's default external API port
-      - "3014:3014" #Node's channels default websocket port
     entrypoint: /home/aeternity/node/ae_mdw/entrypoint.sh
     volumes:
-      - ${PWD}/data/mnesia:/home/aeternity/node/local/rel/aeternity/data/mnesia
```
venimus, line 15: we need that

docker-compose.yml
```diff
@@ -1,24 +1,20 @@
...
   ae_mdw:
+    healthcheck:
+      test: ["CMD", "curl", "-f", "localhost:4000/status"]
+      interval: 1m
+      timeout: 50s
+      retries: 3
     build:
       context: .
       dockerfile: ./Dockerfile
-    ports:
```
venimus, line 7: i think that is also needed

docker-compose.yml
jyeshe: ports as well, right?

docker-compose.yml
```diff
@@ -14,11 +19,7 @@ services:
...
       - ${PWD}/data/mdw.db:/home/aeternity/node/local/rel/aeternity/data/mdw.db
-      - ${PWD}/docker/aeternity.yaml:/home/aeternity/aeternity.yaml
+      - ${HOME}/aeternity.yaml:/home/aeternity/aeternity.yaml
       - ${PWD}/priv:/home/aeternity/node/ae_mdw/priv
-      - ${PWD}/test:/home/aeternity/node/ae_mdw/test
-      - ${PWD}/scripts/test.sh:/home/aeternity/node/ae_mdw/test.sh
-      - ${PWD}:/app
+      - ${PWD}/log:/home/aeternity/node/ae_mdw/log
     expose: [ 3013, 3014, 3113, 4001, 4000 ]
-    environment:
```
sborrazas, line 23: Would it make sense to add `- MIX_ENV=prod` here?

docker-compose.yml
```diff
@@ -1,24 +1,20 @@
...
-      - "4000:4000" #MDW's default port
-      - "4001:4001" #MDW's websocket default port
-      - "3113:3113" #Node's default internal API port
-      - "3013:3013" #Node's default external API port
-      - "3014:3014" #Node's channels default websocket port
     entrypoint: /home/aeternity/node/ae_mdw/entrypoint.sh
     volumes:
-      - ${PWD}/data/mnesia:/home/aeternity/node/local/rel/aeternity/data/mnesia
       - ${PWD}/data/mdw.db:/home/aeternity/node/local/rel/aeternity/data/mdw.db
       - ${PWD}/docker/aeternity.yaml:/home/aeternity/aeternity.yaml
```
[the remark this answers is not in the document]
jyeshe: done

docker-compose.yml
venimus: i think defaults are not needed (unless there is a possibility that someone change them)

venimus (approved):

docker-compose.yml
jyeshe: To be done when we move to releases in order to allow adding dynamic prod configs.

sborrazas: > i think defaults are not needed (unless there is a possibility that someone change them), or if it is not the default of course

The default right now is `dev` and that's what's being run right now in prod

venimus: then it makes sense to add it

docker-compose.yml
```diff
@@ -1,6 +1,11 @@
...
 services:
   ae_mdw:
+    healthcheck:
+      test: ["CMD", "curl", "-f", "localhost:4000/status"]
+      interval: 1m
+      timeout: 50s
+      retries: 3
     build:
       context: .
       dockerfile: ./Dockerfile
```
venimus, line 11:
```suggestion
    image: aeternity/ae_mdw:latest
```

venimus: I think that would make more sense (once we have proper tagging of images)

jyeshe: We are close to a release, it makes sense.

docker-compose.yml
venimus: I guess it also makes sense to add it to the builder https://github.com/aeternity/ae_mdw/blob/master/.github/workflows/dockerhub.yml#L38

docker-compose.yml
venimus: this is for deployment anyway, so it wont mess up with anything because it won't be deployed earlier

docker-compose.yml
jyeshe: The point is that `dev.exs` has always being used and I will start working on the Elixir releases that will change the way config works. I will provide an initial prod config though.

sborrazas (approved):

jyeshe:
To be continued in the next PR as there is no point adding now the MIX_ENV to the environment and removing it with releases (the config will be runtime without mix).

docker-compose.yml
jyeshe: Thanks @venimus, will do after our first release (after stable latest is available)

PR #1088, opened 2022-12-20
feat: introduce {bi, {txi, local_idx}} for precise internal txs refs

sborrazas:
New transactions created by contract calls were only referenced by the txi. With this new implementation, we can now point exactly to the internal contract call that created the transaction, without having to search for it.

**NOTE**: requies a full re-sync to index all of the existing contract call indexes for claims/updates/transfers/etc


lib/ae_mdw/db/sync/name.ex
```diff
@@ -251,7 +252,8 @@ defmodule AeMdw.Db.Sync.Name do
     m_name_exp = Model.expiration(index: {expire, plain_name})
     m_owner = Model.owner(index: {owner, plain_name})
     m_name_owner_deactivation = Model.owner_deactivation(index: {owner, expire, plain_name})
-    %{tx: %{name_fee: name_fee}} = read_raw_tx!(state, txi)
+    name_claim_tx = DbUtil.read_node_tx(state, txi_idx)
```
jyeshe, line 255: this is a minor, a refactoring opportunity to use pipes since `name_claim_tx` is not used elsewhere

lib/ae_mdw/db/util.ex
```diff
@@ -154,6 +156,33 @@ defmodule AeMdw.Db.Util do
     end
   end
 
+  @spec read_node_tx(state(), Txs.txi_idx()) :: Node.tx()
+  def read_node_tx(state, {txi, -1}) do
```
jyeshe, line 160: nice refactoring :100:

lib/ae_mdw/db/name.ex
```diff
@@ -118,33 +115,10 @@ defmodule AeMdw.Db.Name do
...
-            tx
-          end)
-          |> Enum.find(fn tx ->
-            name_hash == :aens_update_tx.name_hash(tx)
-          end)
-          |> :aens_update_tx.pointers()
-      end
-
-    pointers
+  def pointers(state, Model.name(updates: [{_block_index, txi_idx} | _rest_updates])) do
```
jyeshe, line 118: Could we take the opportunity to extend the test cases to cover the internal calls and `paying_for` (for `ga_meta` would be nice too but we have on `name_controller_test` a thorough one)?

lib/ae_mdw/db/sync/contract.ex
```diff
@@ -139,62 +144,44 @@ defmodule AeMdw.Db.Sync.Contract do
...
+      {local_idx, "AENS.transfer", :name_transfer_tx, _aetx, tx, _tx_hash} ->
+        NameTransferMutation.new(tx, {call_txi, local_idx}, block_index)
 
+      {local_idx, "AENS.revoke", :name_revoke_tx, _aetx, tx, _tx_hash} ->
         tx
         |> :aens_revoke_tx.name_hash()
-        |> NameRevokeMutation.new(call_txi, block_index)
+        |> NameRevokeMutation.new({call_txi, local_idx}, block_index)
+
+      {_local_idx, fname, _tx_type, _aetx, _tx, _tx_hash} when fname in @ignored_tx_calls ->
```
jyeshe, line 175: I would agree to make a full match to any other event like before in such a way that I would care only about the events that need to be handled (the ignored ones would be the complement set that might change). An pattern match error would be caught by a sync but would save me some time detecting which new one should match.

jyeshe (commented):
What I like about this PR is that it allows a direct tracking if a name or oracle transition happened due to a specific transaction or after a contract call. I miss only two things:
- This info could be available not only for internal tracking but to the user
- For each mutation changed, would expect the respective test suite to contemplate the new input.

lib/ae_mdw/db/sync/name.ex
sborrazas: Done

lib/ae_mdw/db/sync/contract.ex
sborrazas: Pattern-matching error display the structure that it wasn't pattern-matched to. added a `raise` anyway

jyeshe: It was matching any other event before:
https://github.com/aeternity/ae_mdw/blob/bc54cd63044bb88906637aeb0445c4ea4e132d5c/lib/ae_mdw/db/sync/contract.ex#L153
To ignore internal calls, we would need just like that, for any other that doesn't have to be handled.

sborrazas: I'd rather raise an exception, if we mistakenly add a clause that doesn't match the `tx_type` (e.g. `{local_idx, "ANES.revoke", :name_foo_tx, ...}` we would never find out about it

jyeshe: `fname=AENS.revoke` wouldn't match either for `fname in @ignored_tx_calls` but it's okay, as new events are introduced some tests would be able to cover it.

jyeshe (approved):

PR #1089, opened 2022-12-20
chore: specialize build envs

jyeshe:
- Prepare for Elixir releases with explicit prod build env (runtime will be from `runtime.exs`)
- Reintroduce `AETERNITY_CONFIG` env var as it's needed by node on runtime


scripts/test.sh
```diff
@@ -2,6 +2,6 @@
 
 set -xe
 
-elixir --sname aeternity@localhost -S mix test
+MIX_ENV=test elixir --sname aeternity@localhost -S mix test
```
sborrazas, line 5: This isn't really needed, since it's set by `mix test` automatically

jyeshe: done, made implicit again

docker-compose-dev.yml
```diff
@@ -4,6 +4,9 @@ services:
     build:
       context: .
       dockerfile: ./Dockerfile
+      args:
+        MIX_ENV: dev
+    image: aeternity/ae_mdw_dev:latest
```
venimus, line 9: we don't have this repository `ae_mdw_dev` in dockerhub

venimus: I guess that needs a rebase

venimus (approved):

PR #1090, opened 2022-12-20
feat: return event name when it is aexn

jyeshe:
Set `event_name` on `/contracts/logs` response if it's an AEX-N event.

### DIFF
--- a/lib/ae_mdw/aexn_contracts.ex
+++ b/lib/ae_mdw/aexn_contracts.ex
@@ -12,6 +12,7 @@ defmodule AeMdw.AexnContracts do
 
   import AeMdw.Util.Encoding, only: [encode_contract: 1]
 
+  @type event_name() :: String.t()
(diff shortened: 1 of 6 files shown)


lib/ae_mdw/aexn_contracts.ex
```diff
@@ -110,6 +111,12 @@ defmodule AeMdw.AexnContracts do
     end
   end
 
+  @spec event_name(AeMdw.Contracts.event_hash()) :: String.t() | nil
```
sborrazas, line 114: Since we're using it in several places, can we use a new type for these names? `event_name()` maybe

sborrazas (approved):

lib/ae_mdw/aexn_contracts.ex
jyeshe: Only two places but I know you like named type specs so here it is (:

PR #1091, opened 2022-12-20
chore: remove old makefile

jyeshe:


scripts/do.sh
```diff
@@ -0,0 +1,23 @@
...
+  "shell")
+    iex --sname $NAME -S mix phx.server
+    ;;
+
+  "remsh")
+    iex --sname console --remsh $NAME
+    ;;
+
+  "docker-shell")
+    docker-compose run --rm --workdir=/app --entrypoint="" ae_mdw /bin/bash
```
sborrazas, line 13: Can we add `--service-ports` here?

jyeshe: done

sborrazas (approved):

PR #1093, opened 2022-12-21
fix: handle inner name_update pointers

jyeshe:
Yet another binary pointer to be handled: https://testnet.aeternity.io/mdw/txi/6011255 
(GaMetaTx with inner :name_update_tx having a binary non-text pointer)


sborrazas (approved):
nice find!

PR #1095, opened 2022-12-23
fix(ci): publish docker on tag or master push

jyeshe:


.github/workflows/dockerhub.yml
```diff
@@ -2,8 +2,10 @@ name: Dockerhub Pipeline
 
 on:
   push:
+    branches:
+      - master
```
venimus, line 6: that is actually redundant, because it generates tags only on default branch, and if there are no tags it won't push to dockerhub. But it is ok for clarity

venimus: Though I thought that we removed `type=ref,event=pr` here https://github.com/aeternity/ae_mdw/pull/1095/files#diff-79be07d4634c4c8e630c58e010635d3621a19f63e92861570af1570120df38b6R30

jyeshe: Even if there was a commit on master without tags? Since there is a master [here](https://hub.docker.com/r/aeternity/ae_mdw/tags) it might the case to also push under any condition or remove the master.

jyeshe:
@venimus I am closing this PR in favor of one of yours that publishes a tagged release image built with docker-compose.

.github/workflows/dockerhub.yml
venimus: it is supposed to push any commit to `master` as `master` and `v*` only when tagged

PR #1098, opened 2022-12-23
feat: add key block hash to activities

sborrazas:


lib/ae_mdw/activities.ex
```diff
@@ -165,15 +166,37 @@ defmodule AeMdw.Activities do
...
     end
   end
 
+  defp render_activities(state, account_pk, activities_locators_data) do
+    {activities_locators_data, _acc} =
+      Enum.map_reduce(activities_locators_data, {nil, nil}, fn
+        {{height, _txi, _local_idx}, _data} = locator, {height, enc_kb_hash} = kb_info ->
+          {{enc_kb_hash, locator}, kb_info}
+
+        {{height, _txi, _local_idx}, _data} = locator, _kb_info ->
```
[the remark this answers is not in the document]
jyeshe: Could you double check the intention here? `kb_info` is always a tuple of 2 elements and the `locator` has the same  format for both clauses matching always with the first one.

lib/ae_mdw/activities.ex
```diff
@@ -165,15 +166,37 @@ defmodule AeMdw.Activities do
...
+
+          {{enc_kb_hash, locator}, {height, enc_kb_hash}}
+      end)
+
+    Enum.map(activities_locators_data, fn {key_block_hash, {{height, txi, _local_idx}, data}} ->
+      {type, payload} = render_payload(state, account_pk, height, txi, data)
+
+      %{
+        height: height,
+        key_block_hash: key_block_hash,
```
jyeshe, line 193: Depending on the requirement for key block information, instead of `key_block_hash`, or additionally to it, could be the `block_hash` with the micro block like `/txs`, to indicate the exact its location.

sborrazas: This is a good idea, I'm going to implement it this way

sborrazas: Done

lib/ae_mdw/activities.ex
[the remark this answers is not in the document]
jyeshe: solved by microblock inclusion

jyeshe (approved):

PR #1099, opened 2022-12-26
chore: remove tx hashes handling on int contract calls

sborrazas:


sborrazas:
> whats the reason to remove it, isn't this a breaking change?

this isn't a breaking change, it's more of a removal of a tx_hash that was being handled internally throughout contract call events which is always the same tx_hash as the actual contract call. this was very confusing, so on this PR I removed the tx_hash altogether. but this is only done internally, there's no change in the endpoints

jyeshe (approved):

PR #1100, opened 2022-12-27
chore: use master instead of latest to pull docker image

sborrazas:
@venimus is the intention to use this `master` tag instead of latest? Also, I wasn't able to find the `ae_mdw_dev` image on dockerhub, do we need to name it?

### DIFF
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -6,7 +6,7 @@ services:
       dockerfile: ./Dockerfile
       args:
         MIX_ENV: dev
-    image: aeternity/ae_mdw:latest
+    image: aeternity/ae_mdw:master
     ports:
       - "4000:4000" #MDW's default port
       - "4001:4001" #MDW's websocket default port
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -20,5 +20,6 @@ services:
       - ${PWD}/docker/aeternity.yaml:/home/aeternity/aeternity.yaml
       - ${PWD}/priv:/home/aeternity/node/ae_mdw/priv
       - ${PWD}:/app
+      - ${PWD}/docker/gitconfig:/root/.gitconfig
     environment:
       - AETERNITY_CONFIG=/home/aeternity/aeternity.yaml
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -6,7 +6,7 @@ services:
       dockerfile: ./Dockerfile
       args:
         MIX_ENV: dev
-    image: aeternity/ae_mdw_dev:latest
+    image: aeternity/ae_mdw_dev:master
     ports:
       - "4000:4000" #MDW's default port
       - "4001:4001" #MDW's websocket default port
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -6,7 +6,7 @@ services:
       dockerfile: ./Dockerfile
       args:
         MIX_ENV: dev
-    image: aeternity/ae_mdw_dev:master
+    image: aeternity/ae_mdw:latest
     ports:
       - "4000:4000" #MDW's default port
       - "4001:4001" #MDW's websocket default port
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -6,7 +6,7 @@ services:
       dockerfile: ./Dockerfile
       args:
         MIX_ENV: dev
-    image: aeternity/ae_mdw:master
+    image: aeternity/ae_mdw_dev:latest
     ports:
       - "4000:4000" #MDW's default port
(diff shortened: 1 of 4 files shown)


venimus:
Latest is the latest version tagged image. “Master” is the latest push to master which is essentially the dev branch. There were pr branches but was dropped. I think it is the correct way to use master image for development (as it is here in the pr) and “latest” for public deployments.

docker-compose.yml
```diff
@@ -6,7 +6,7 @@ services:
       interval: 1m
       timeout: 50s
       retries: 3
-    image: aeternity/ae_mdw:latest
+    image: aeternity/ae_mdw:master
```
venimus, line 9: This should stay “latest”

venimus (changes requested):

docker-compose.yml
sborrazas: The latest tag doesn't exist on dockerhub, and it needs to be built locally every time. And if users want to run `aeternity/ae_mdw` they won't be able to. We should probably figure out a way to build on a per-version basis (not branch) and create latest tag as well. I'll revert it back to `latest`

docker-compose-dev.yml
```diff
@@ -6,7 +6,7 @@ services:
       dockerfile: ./Dockerfile
       args:
         MIX_ENV: dev
-    image: aeternity/ae_mdw_dev:master
+    image: aeternity/ae_mdw:latest
```
venimus, line 9: I think that this should be master, so you can either pull the latest master or build it

sborrazas: Good idea. Done

venimus (approved):

jyeshe (changes requested):
I would rather continue with the purpose we've discussed where the `docker-compose-dev.yml` is used for feature and bug fixing development (not for the `master`).

venimus:
> I would rather continue with the purpose we've discussed where the `docker-compose-dev.yml` is used for feature and bug fixing development (not for the `master`).

I don't really get the workflow, but IMHO those things are not really related to the image tag. My logic is that anyway that would go to the `master` branch eventually. This won't be pushed to dockerhub anyway, so you could tag it as you think is most appropriate, just being `latest` doesn't make much sense to me, as it refers to the stable release and development is never the stable.

jyeshe:
> > I would rather continue with the purpose we've discussed where the `docker-compose-dev.yml` is used for feature and bug fixing development (not for the `master`).
> 
> I don't really get the dev workflow, but IMHO those things are not really related to the image tag. My logic is that anyway that any code changes would go to the `master` branch eventually. This won't be pushed to dockerhub anyway, so you could tag it as you think is most appropriate, just being `latest` doesn't make much sense to me, as it refers to the stable release and development is never the stable.

When both `build` and `image` is provided for an image that doesn't exist on dockerhub (`ae_mdw_dev`), the `docker-compose` creates the image locally with the name provided by the `image` field. It's helpful to separate what is `ae_mdw_dev` (for dev branches) from `ae_mdw` (for master and release tags).

docker/gitconfig
```diff
@@ -0,0 +1,2 @@
+[safe]
```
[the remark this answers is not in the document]
sborrazas: Ever since we made these docker updates, I'm starting to get these errors when developing locally:
```
fatal: detected dubious ownership in repository at '/app'
                                                         To add an exception for this directory, call:

                                                                                                      	git config --global --add safe.directory /app
                                                                                                                                                     ** (MatchError) no match of right hand side value: {"", 128}
    (stdlib 3.14.2.2) erl_eval.erl:453: :erl_eval.expr/5
    (stdlib 3.14.2.2) erl_eval.erl:126: :erl_eval.exprs/5
    (stdlib 3.14.2.2) erl_eval.erl:449: :erl_eval.expr/5
    (elixir 1.13.4) lib/code.ex:404: Code.validated_eval_string/3
    (elixir 1.13.4) lib/config.ex:260: Config.__eval__!/3
    (mix 1.13.4) lib/mix/tasks/loadconfig.ex:54: Mix.Tasks.Loadconfig.load_compile/1
```
this happens when `git log` in the `config/config.exs` file is executed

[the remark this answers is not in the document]
jyeshe: This error was actually one of the motivations to get started working again on the Docker image.
The `fatal: detected dubious ownership in repository at '/app'` was happening here: 
https://user-images.githubusercontent.com/750314/207631812-d0dc9dd8-54d2-4899-a03f-892fa9dbcc90.png
https://github.com/aeternity/ae_mdw/pull/1080

Locally what I am doing and working fine is run the dev image with `docker-compose -f docker-compose-dev.yml up` and then connect to it with a remote shell.

jyeshe (approved):

PR #1104, opened 2022-12-29
feat: include source tx_hash on nested names endpoints

sborrazas:


jyeshe (commented):
The tx hash seems helpful, the tx type is already returned under `tx`:
image

sborrazas:
> The tx hash seems helpful, the tx type is already returned under `tx`: image

The `tx_type` does not always match the `tx_type` from the tx object. for instance, with name claims `tx_type` can either be `NameClaimTx` or `ContractCallTx`

jyeshe:
> > The tx hash seems helpful, the tx type is already returned under `tx`: image
> 
> The `tx_type` does not always match the `tx_type` from the tx object. for instance, with name claims `tx_type` can either be `NameClaimTx` or `ContractCallTx`

I see the `tx` is not actual chain transaction, might be an internal one. Having `tx_type` and `tx.type` that might differ it's a bit confusing to me. If not for a specific reason, indicating if it's from a contract call or not with a boolean might to be more clear.

jyeshe (commented):
I see, might be an internal tx not the chain one. Having tx_type and tx.type that might differ it's a bit confusing to me. If not for a specific reason, indicating if it's from a contract call or not with a boolean might to be more clear.

sborrazas:
> I see, might be an internal tx not the chain one. Having tx_type and tx.type that might differ it's a bit confusing to me. If not for a specific reason, indicating if it's from a contract call or not with a boolean might to be more clear.

Renamed the field to `source_tx_hash` and `source_tx_type`. Also added boolean `internal_source` to indicate whether it's internal or not

PR #1106, opened 2023-01-04
feat: query channel reserve at a hash

### LINKED ISSUE #1105: Query state channel reserve
Additionally to the channel transactions info, it's requested that the Mdw should expose the balance (reserve) of the channel.

jyeshe:
Closes #1105 
Refs #1032

### DIFF
--- a/lib/ae_mdw/channels.ex
+++ b/lib/ae_mdw/channels.ex
@@ -7,6 +7,7 @@ defmodule AeMdw.Channels do
   alias AeMdw.Collection
   alias AeMdw.Db.Model
   alias AeMdw.Db.State
+  alias AeMdw.Db.Util, as: DbUtil
   alias AeMdw.Error
   alias AeMdw.Error.Input, as: ErrInput
   alias AeMdw.Txs
@@ -19,6 +20,7 @@ defmodule AeMdw.Channels do
 
   @typep state() :: State.t()
   @typep pubkey() :: AeMdw.Node.Db.pubkey()
+  @typep block_hash() :: String.t()
   @typep pagination() :: Collection.direction_limit()
   @typep range() :: {:gen, Range.t()} | nil
   @typep cursor() :: Collection.pagination_cursor()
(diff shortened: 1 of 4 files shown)


lib/ae_mdw/channels.ex
```diff
@@ -175,4 +177,34 @@ defmodule AeMdw.Channels do
...
+          param_block_hash
+        end
+
+      {:error, _reason} ->
+        update_block_hash
+    end
+  end
+
+  defp get_block_index(state, param_block_hash) do
+    block_hash = Validate.id!(param_block_hash)
```
sborrazas, line 200: Perhaps we can send the decoded block_hash binary to avoid decoding it twice

lib/ae_mdw_web/controllers/channel_controller.ex
```diff
@@ -22,10 +22,22 @@ defmodule AeMdwWeb.ChannelController do
   end
 
   @spec channel(Conn.t(), map()) :: Conn.t()
-  def channel(%Conn{assigns: %{state: state}} = conn, %{"id" => id}) do
+  def channel(%Conn{assigns: %{state: state}} = conn, %{"id" => id} = params) do
+    block_hash = params["block_hash"]
+
     with {:ok, channel_pk} <- Validate.id(id, [:channel]),
-         {:ok, channel} <- Channels.fetch_channel(state, channel_pk) do
+         true <- valid_optional_block_hash?(block_hash),
```
sborrazas, line 29: This is the first decoding that we do, we can use its value by returning `{:ok, block_hash}`. In addition, if the parameter is not a valid `block_hash` the whole function returns `false` which is not handled by the FallbackController and it returns a 500 error

sborrazas (changes requested):

lib/ae_mdw/channels.ex
jyeshe: It was avoiding an additional parameter since the type is also needed and it's a few microseconds operation but did the optimization though.

lib/ae_mdw_web/controllers/channel_controller.ex
jyeshe: done

thepiwo (approved):

sborrazas (approved):

PR #1114, opened 2023-01-09
feat: add offchain rounds to channel transactions

jyeshe:
refs #1032

### DIFF
--- a/lib/ae_mdw/db/format.ex
+++ b/lib/ae_mdw/db/format.ex
@@ -286,6 +286,26 @@ defmodule AeMdw.Db.Format do
     |> update_in(["previous"], fn prevs -> Enum.map(prevs, &expand_name_info(state, &1)) end)
   end
 
+  @payload_index 3
+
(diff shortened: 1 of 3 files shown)


thepiwo (approved):

sborrazas (approved):
LGTM. Maybe, in the future, we can also contribute and add new getters on the node source, similar to how it's done in https://github.com/aeternity/aeternity/blob/v6.7.0/apps/aens/src/aens_claim_tx.erl#L33
to avoid having to dig inside the opaque term `:aesc_slash_tx.tx()`

jyeshe:
> similar to how it's done in

Thanks for the suggestion. Now that we'll be able to contribute more directly to the Node gonna suggest these additional getters.

jyeshe:
> LGTM. Maybe, in the future, we can also contribute and add new getters on the node source, similar to how it's done in https://github.com/aeternity/aeternity/blob/v6.7.0/apps/aens/src/aens_claim_tx.erl#L33 to avoid having to dig inside the opaque term `:aesc_slash_tx.tx()`

Thanks for the suggestion. Now that we'll be able to contribute more directly to the Node gonna suggest these additional getters.

PR #1120, opened 2023-01-12
feat: add channel participants to channel txs

jyeshe:
refs #1032

### DIFF
--- a/lib/ae_mdw/db/format.ex
+++ b/lib/ae_mdw/db/format.ex
@@ -13,7 +13,6 @@ defmodule AeMdw.Db.Format do
   alias AeMdw.Names
   alias AeMdw.Node
   alias AeMdw.Txs
-  alias AeMdw.Validate
   alias AeMdw.Db.Sync.InnerTx
 
   require Model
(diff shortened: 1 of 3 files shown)


lib/ae_mdw/db/format.ex
```diff
@@ -286,31 +288,44 @@ defmodule AeMdw.Db.Format do
...
   end
 
-  defp custom_encode(_state, :channel_slash_tx, tx, tx_rec, _signed_tx, _txi, _block_hash) do
-    tx_rec
-    |> elem(@payload_index)
-    |> put_channel_offchain_round(tx)
+  defp custom_encode(state, :channel_slash_tx, tx, tx_rec, _signed_tx, _txi, _block_hash) do
+    tx
+    |> put_channel_offchain_round(elem(tx_rec, @payload_index))
+    |> put_channel_participants(state)
```
sborrazas, line 318: Can we use the already decoded value inside the tx for the channel pk? E.g.
```elixir
|> put_channel_participants(:channel_slash_tx.channel_pubkey(tx), state)
```

jyeshe: With the `tx` map serialized by the node, we can get in a generic way for the 7 transaction types.

sborrazas: On each `custom_encode` clause we know the transaction type and already do different things per transaction type. I think it will look better and it will type-check once we get the aeternity types in MDW. But you don't need to change it if you don't want to

sborrazas (approved):

lib/ae_mdw/db/format.ex
jyeshe: The type received is not the one with the getter, i.e `:channel_slash_tx.channel_pubkey(tx)` wouldn't work. There would be a need for additional conversion.

jyeshe:
Besides not needing the conversion of the tx type to the module with the getter, the `put_channel_participants` is generic and safe enough. No worries regarding the type or the value, it comes from the Node.

lib/ae_mdw/db/format.ex
jyeshe: Done using `signed_tx` with `aesc_utils`. As it was already assured to be safe by the Node, I would consider it a micro optimization but it will be fun to play around with this.

thepiwo (approved):

PR #1121, opened 2023-01-12
feat: update node to version 6.7.0

sborrazas:
closes #1121


lib/ae_mdw/db/contract.ex
```diff
@@ -143,7 +143,7 @@ defmodule AeMdw.Db.Contract do
 
   @spec logs_write(state(), txi(), txi(), Contract.call()) :: state()
   def logs_write(state, create_txi, txi, call_rec) do
-    contract_pk = :aect_call.contract_pubkey(call_rec)
+    {_id_tag, contract_pk} = call_rec |> :aect_call.contract_id() |> :aeser_id.specialize()
```
jyeshe, line 146: This wouldn't be needed since the `:aect_call` has the `contract_pubkey` getter.

lib/ae_mdw/db/format.ex
```diff
@@ -117,7 +117,7 @@ defmodule AeMdw.Db.Format do
   end
 
   defp custom_raw_data(state, :contract_call_tx, tx, tx_rec, signed_tx, block_hash) do
-    contract_pk = :aect_call_tx.contract_pubkey(tx_rec)
+    {_id_tag, contract_pk} = tx_rec |> :aect_call_tx.contract_id() |> :aeser_id.specialize()
```
jyeshe, line 120: Maybe we could have a helper function to get only the desired data since the `id_tag` is ignored in all places.

lib/ae_mdw/db/contract.ex
sborrazas: You're right, done

lib/ae_mdw/db/format.ex
sborrazas: Good call, added `Db.id_pubkey/1` to access the pubkey from the aeser id

jyeshe (approved):

thepiwo (approved):

Issue #1122, opened 2023-01-12
mainnet Websocket subscription Bug


Lfh2msn:
@thepiwo 
wss://mainnet.aeternity.io/mdw/v2/websocket
Subscribing to an account in the official mdw, or Transactions, is problematic. I initiate a normal transfer on the chain, mdw ws does not inform me.

Sometimes I can get a tx from the monitoring system, but I never receive my designated account or TX hash.

WeTrue has been using websocket heavily for a long time, and the problem occurred during a recent mdw upgrade.

ak_dMyzpooJ4oGnBVX35SCvHspJrq55HAAupCwPQTDZmRDT5SSSW
This is the WeTrue receiving account, and you can try to subscribe and initiate another transfer.

thepiwo:
@sborrazas and @jyeshe are looking into it, there has been some problems with the deployment that we are fixing currently

jyeshe:
@Lfh2msn the delivery of the transactions was normalized on mainnet and testnet. There might be some jitter though. If you run on premise, the recommended way, the fix is available on the latest published [image](https://hub.docker.com/r/aeternity/ae_mdw/tags).

Issue #1033, opened 2022-11-22
Unlimited websockets subscriptions per client


jyeshe:
@thepiwo as requested the limit control on the amount of subscriptions per socket was removed from #1030 (0f9e1888) and postponed until having a better observability on how it would affect client apps. 

It will need to be tackled somehow another time since the lack of this feature allows a memory usage exploit (subscriptions objects are kept in memory).

For the moment on AWS API Gateway there is a throttling option but might applied only to route-level:
https://docs.aws.amazon.com/apigateway/latest/developerguide/websocket-api-protect.html

When it will be possible to be activated, the implementation on 0f9e1888 diff easily solves it for any environment and on a socket basis.

jyeshe:
@venimus could we have some protection as https://docs.aws.amazon.com/apigateway/latest/developerguide/websocket-api-protect.html for the websocket endpoints (v1 and v2)?

PR #1132, opened 2023-01-17
refactor: enqueue a block only once to ws broadcasting

jyeshe:
## What

- Enqueue a block or block + transactions only once to avoid filling GenServer mailbox with data to be discarded (same block that is synced multiple times is enqueued once).
- Enqueue only the microblock header (without transactions) for "MicroBlocks" subscription
- Skips fetching and serializing multiple transactions when there is no "Transactions" or "Object" subscription
- Simplify blocks serialization by reusing `Db.prev_block_type`

## Why

refs #1122 

- Faster delivery (it takes less time to get to broadcast the last blocks specially when the Mdw height lags behind Node height when recovering or on a full sync)
- Smaller websocket footprint

### DIFF
--- a/lib/ae_mdw_web/websocket/subscriptions.ex
+++ b/lib/ae_mdw_web/websocket/subscriptions.ex
@@ -73,8 +73,8 @@ defmodule AeMdwWeb.Websocket.Subscriptions do
     end
   end
 
-  @spec has_subscribers(version(), channel) :: boolean()
-  def has_subscribers(version, channel) do
+  @spec has_subscribers?(version(), channel) :: boolean()
+  def has_subscribers?(version, channel) do
     {:ok, channel_key} = channel_key(version, channel)
 
     case :ets.match(@subs_channel_pid, {channel_key, :"$1"}, 1) do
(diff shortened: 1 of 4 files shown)


lib/ae_mdw_web/websocket/subscriptions.ex
```diff
@@ -73,6 +73,27 @@ defmodule AeMdwWeb.Websocket.Subscriptions do
...
+    {:ok, channel_key} = channel_key(version, channel)
+
+    case :ets.match(@subs_channel_pid, {channel_key, :"$1"}, 1) do
+      {_some, _continuation} -> true
+      @eot -> false
+    end
+  end
+
+  @spec has_object_subscribers(version()) :: boolean()
+  def has_object_subscribers(version) do
```
sborrazas, line 87: Can we add `?` to the end of the function names for these?

sborrazas (approved):
Good call on improving the amount of messages sent to the broadcaster. I recently found that, even when blocks are only sent once only, the broadcaster reduces the overall application performance (by a lot) because of all the data being sent. I think we need to talk to the explorer team to avoid having to send this much data.

lib/ae_mdw_web/websocket/subscriptions.ex
jyeshe: Done

jyeshe:
> Good call on improving the amount of messages sent to the broadcaster. I recently found that, even when blocks are only sent once only, the broadcaster reduces the overall application performance (by a lot) because of all the data being sent. I think we need to talk to the explorer team to avoid having to send this much data.

For "Transactions" sub unfortunately we can't do much per connection but for objects we have it pending https://github.com/aeternity/ae_mdw/issues/1033. I agree it should be investigated though how many connections it opens that makes the same subscriptions (e.g if there is more than one per browser tab subscribing to "Transactions").

thepiwo (approved):

PR #1133, opened 2023-01-17
chore: remove no longer needed oracle queries nonce fix

sborrazas:


thepiwo (approved):

jyeshe (commented):
LGTM. Approving to merge after 1.37.1 goes well otherwise it will be an addtional PR to revert.

sborrazas:
> LGTM. Approving to merge after 1.37.1 goes well otherwise it will be an additional PR to revert.

Agreed, this change is a slightly fragile one, I'm currently running a full sync of mainnet with it just to make sure

sborrazas:
This works just fine with a full mainnet sync, I think we're good to roll with it

jyeshe (commented):
Would there be a case with two "Oracle.query" in the same transaction on mainnet
to make sure it performs an increment replace [this](https://github.com/aeternity/ae_mdw/blob/7ba828a6f616fc03a0224ad86db445e230c8a7e8/lib/ae_mdw/db/sync/contract.ex#L191) one?

sborrazas:
> Would there be a case with two "Oracle.query" in the same transaction on mainnet to make sure it performs an increment to replace [this](https://github.com/aeternity/ae_mdw/blob/7ba828a6f616fc03a0224ad86db445e230c8a7e8/lib/ae_mdw/db/sync/contract.ex#L191) one?

Just ran a sync on testnet and the results are the same. The nonce obtained was the correct one with [this](https://github.com/aeternity/aeternity/pull/4027) fix merged

jyeshe (approved):

PR #1136, opened 2023-01-18
chore: add MIX_ENV=test for being able to run tests

sborrazas:
Currently all `mix test` tasks (unit and integration) produce an error when running the docker-dev container with the MIX_ENV=dev. This is not necessary, since MIX_ENV will default to `test` when running `mix test`.


jyeshe (changes requested):
I would rather stick to use the same `Dockerfile` for development and prod and specially the same for CI testing and production otherwise it might run in a different environment giving no clue whatsoever of what happens in advance. Or in the best case we gonna need to keep two files to be compatible.

For this reason, the Dockerfile receives the `MIX_ENV` as argument. To run the tests inside the dev container please add once these lines to `docker-compose-dev.yml`: 
```
- ${PWD}/lib:/home/aeternity/node/ae_mdw/lib
- ${PWD}/scripts:/home/aeternity/node/ae_mdw/scripts
- ${PWD}/test:/home/aeternity/node/ae_mdw/test
```
With `MIX_ENV` set by `/scripts/do.sh` as `MIX_ENV=test elixir --sname $NAME -S mix test`, the tests run fine under `/home/aeternity/node/ae_mdw/$`.

sborrazas:
> I would rather stick to use the same `Dockerfile` for development and prod and specially the same for CI testing and production otherwise it might run in a different environment giving no clue whatsoever of what happens in advance. Or in the best case we gonna need to keep two files to be compatible.
> 
> For this reason, the Dockerfile receives the `MIX_ENV` as argument. To run the tests inside the dev container please add once these lines to `docker-compose-dev.yml`:
> 
> ```
> - ${PWD}/lib:/home/aeternity/node/ae_mdw/lib
> - ${PWD}/scripts:/home/aeternity/node/ae_mdw/scripts
> - ${PWD}/test:/home/aeternity/node/ae_mdw/test
> ```
> 
> With `MIX_ENV` set by `/scripts/do.sh` as `MIX_ENV=test elixir --sname $NAME -S mix test`, the tests run fine under `/home/aeternity/node/ae_mdw/$`.

Perhaps we can do this by removing the `MIX_ENV` from the docker build, and instead, use `MIX_ENV=${MIX_ENV} ./entrypoint.sh` and leave the default envs for the rest of the tasks for all environments

jyeshe:
> Perhaps we can do this by removing the `MIX_ENV` from the docker build, and instead, use `MIX_ENV=${MIX_ENV} ./entrypoint.sh` and leave the default envs for the rest of the tasks for all environments

Better continue publishing the image built with `prod` and `elixir --sname $NAME -S mix test` fails under `/app` whether or not with `MIX_ENV=test`

jyeshe (commented):
Better continue publishing the image built with `prod` and `elixir --sname $NAME -S mix test` fails under `/app` whether or not with `MIX_ENV=test`

jyeshe (approved):
Thanks for looking for the cause of the test error on `/app`. With this and letting it recreate the `VERSION` it works.

PR #1137, opened 2023-01-19
fix: look for pointee also on previous name record

jyeshe:


lib/ae_mdw/db/name.ex
```diff
@@ -220,6 +220,17 @@ defmodule AeMdw.Db.Name do
     |> :aec_trees.ns()
   end
 
+  defp pointee_at(state, Model.name(index: name, updates: updates, previous: previous), ref_txi)
+       when previous != nil do
+    case List.last(updates) do
+      {_block, update_txi} when ref_txi >= update_txi ->
```
sborrazas, line 226: `Model.name()` `updates` attribute contains `bi_txi_idx()` tuples, so it would not be a `txi`, but a `{txi, idx}` tuple

sborrazas (changes requested):

sborrazas (approved):

PR #1139, opened 2023-01-19
refactor: remove duplicated v2 ws enqueuing

jyeshe:
- Specialize only the order of key blocks (before or after micro blocks+transactions)
- Skips enqueuing micro blocks when there is no subscription

### DIFF
--- a/lib/ae_mdw_web/websocket/broadcaster.ex
+++ b/lib/ae_mdw_web/websocket/broadcaster.ex
@@ -68,8 +68,7 @@ defmodule AeMdwWeb.Websocket.Broadcaster do
     {:ok, hash} = block |> :aec_blocks.to_header() |> :aec_headers.hash_header()
 
     if not already_processed?({:txs, hash, source}) do
-      versions = if broadcast_transaction?(:v1), do: [:v1], else: []
-      versions = if broadcast_transaction?(:v2), do: versions ++ [:v2], else: versions
+      versions = Enum.filter(~w(v1 v2)a, &broadcast_transaction?/1)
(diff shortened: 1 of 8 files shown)


lib/ae_mdw_web/websocket/broadcaster.ex
```diff
@@ -278,7 +290,22 @@ defmodule AeMdwWeb.Websocket.Broadcaster do
     |> Enum.uniq()
   end
 
+  defp get_subscribed_versions(channel) do
+    Enum.reduce([:v1, :v2], [], fn version, acc ->
```
sborrazas, line 294: Same here 
```elixir
Enum.filter([:v1, :v2], &Subscriptions.has_subscribers?(&1, channel))
```

scripts/do.sh
```diff
@@ -19,6 +19,6 @@ case $1 in
     ;;
 
   "test")
-    elixir --sname $NAME -S mix test
+    MIX_ENV=test elixir --sname $NAME -S mix test
```
sborrazas, line 22: Not part of this PR I guess

sborrazas (commented):
Nice performance improvement for the broadcaster!

lib/ae_mdw_web/websocket/broadcaster.ex
jyeshe: done

scripts/do.sh
jyeshe: done

sborrazas (approved):

PR #1151, opened 2023-01-27
fix: fix bugs found through integration tests

sborrazas:


jyeshe (commented):
More than welcome update. I think should keep the `fname_prefix =  "Oracle"`. There is an example based on this filter in the readme. Currently the endpoint is failing to return the internal calls (returns an empty array while with `Oracle.query` it works):
https://testnet.aeternity.io/mdw/v2/contracts/calls?function=Oracle

sborrazas:
> More than welcome update. I think should keep the `fname_prefix = "Oracle"`. There is an example based on this filter in the readme. Currently the endpoint is failing to return the internal calls (returns an empty array while with `Oracle.query` it works): https://testnet.aeternity.io/mdw/v2/contracts/calls?function=Oracle

This decision to use full names (to avoid returning the wrong results when scoping) was taken some time ago https://github.com/aeternity/ae_mdw/pull/1005. I'll create a following PR either reverting this feature so that it supports prefixes, I figured out a way we can do this without adding any new index. But for now these integration tests map the current functionality.

jyeshe:
> > More than welcome update. I think should keep the `fname_prefix = "Oracle"`. There is an example based on this filter in the readme. Currently the endpoint is failing to return the internal calls (returns an empty array while with `Oracle.query` it works): https://testnet.aeternity.io/mdw/v2/contracts/calls?function=Oracle
> 
> This decision to use full names (to avoid returning the wrong results when scoping) was taken some time ago #1005. I'll create a following PR either reverting this feature so that it supports prefixes, I figured out a way we can do this without adding any new index. But for now these integration tests map the current functionality.

The decision on #1005 was not to use full names but to not allow filtering both scope and function prefix: 
"As per discussed with @thepiwo, we're simply going to disable filtering by both scope and function prefix" (sic)

We should keep the prefix test for the current functionality that expects function prefixes to work (:

sborrazas:
> 

The decision to approve the PR which doesn't allow to filter by prefixes was made by everyone on the team, and that's the functionality we have right now. I simply updated the integration tests to map the current, expected and agreed behavior.

Either way, I'll update the functionality so that we can filter by both scope and prefix on a separate PR. Once I do that, I'll update integration tests to map that functionality as well.

sborrazas:
> > The decision to approve the PR which doesn't allow to filter by prefixes was made by everyone on the team, and that's the functionality we have right now. I simply updated the integration tests to map the current, expected and agreed behavior.
> > Either way, I'll update the functionality so that we can filter by both scope and prefix on a separate PR. Once I do that, I'll update integration tests to map that functionality as well.
> 
> The expected and agreed behaviour was not to introduce a breaking change on the API. Any other behavior it's very clear it's a bug that was introduced by the PR which indicates we need to be even more rigorous about the test coverage. I don't see a point faking results on integration tests when they are not blocking the development while still indicate an issue.

I hadn't seen the already implemented behavior for returning an error when scope and function params are provided. I was going to add a test case for it, but it will get removed on the next PR of mine. Reverted the failing test back to continue failing. Right now it's the only integration test that's failing, but it will be fixed on the next PR. Thank you for looking into it

jyeshe (approved):

PR #1162, opened 2023-02-02
Add NODE_URL docker build argument

dincho:
More often than not, the mdw docker image needs to be build with dev/unreleased node versions i.e. master (https://builds.aeternity.io/aeternity-latest-ubuntu-x86_64.tar.gz). 

This change allows that as otherwise default URL pattern (NODE_VERSION) does not.


jyeshe (approved):

jyeshe:
LGTM. The rule for the linter is based on conventional commits ex: "ci: add NODE_URL docker build argument"

sborrazas (approved):

PR #1164, opened 2023-02-02
fix: use path to decide websocket version

jyeshe:


lib/ae_mdw_web/websocket/socket_handler.ex
```diff
@@ -19,10 +19,8 @@ defmodule AeMdwWeb.Websocket.SocketHandler do
   end
 
   @impl Phoenix.Socket.Transport
-  def connect(%{params: params}) do
-    version_bin = Map.get(params, "version", "1")
-
-    case Map.fetch(@versions, version_bin) do
+  def connect(%{connect_info: %{version: endpoint_version}}) do
+    case Map.fetch(@versions, endpoint_version) do
```
sborrazas, line 23: Since the version si no longer provided by the user, we can use `Map.fetch!/2` instead, we would never need to return an error

jyeshe: Resolved, worked with the atom directly.

sborrazas (approved):

PR #1169, opened 2023-02-06
fix: encode blocks on the /v2/blocks endpoint using formatter

sborrazas:
This is causing errors with blocks like https://mainnet.aeternity.io/mdw/v2/blocks/444851


jyeshe (approved):
Nice catch! One optional refactor and looks good to me.

PR #1179, opened 2023-02-09
feat: specialize ws subscription by source

### LINKED ISSUE #1165: Allow subscribing to a websockets data source (node or mdw)
On v1 the mdw broadcasting was introduced. On v2 there will be an additional optional parameter on the subscription to allow receiving the payloads after the node sync and/or mdw sync.

jyeshe:
closes #1165

### DIFF
--- a/lib/ae_mdw_web/websocket/broadcaster.ex
+++ b/lib/ae_mdw_web/websocket/broadcaster.ex
@@ -12,6 +12,8 @@ defmodule AeMdwWeb.Websocket.Broadcaster do
   alias AeMdwWeb.Websocket.Subscriptions
   alias AeMdwWeb.Websocket.SocketHandler
 
+  import AeMdw.Util.Encoding, only: [encode: 2]
+
   require Ex2ms
 
   @hashes_table :broadcast_hashes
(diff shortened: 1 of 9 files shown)


lib/ae_mdw_web/websocket/broadcaster.ex
```diff
@@ -109,111 +116,71 @@ defmodule AeMdwWeb.Websocket.Broadcaster do
...
-      ser_tx = :aetx_sign.serialize_for_client(header, tx)
-
-      ser_tx
-      |> encode_message("Transactions", source)
-      |> broadcast_tx(tx_pids)
-
-      obj_msg = encode_message(ser_tx, "Object", source)
+  defp serialize_tx(tx, _block, :mdw, :v2) do
+    tx_hash = :aetx_sign.hash(tx)
+    Txs.fetch(State.mem_state(), tx_hash, true)
```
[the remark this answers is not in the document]
jyeshe: It's not expensive because persisent_term get performs well, it's used for every endpoint request, but makes sense to move it to outside of the loop.

lib/ae_mdw_web/websocket/broadcaster.ex
```diff
@@ -109,111 +116,72 @@ defmodule AeMdwWeb.Websocket.Broadcaster do
...
-    |> :aec_blocks.txs()
-    |> Enum.each(fn tx ->
-      ser_tx = :aetx_sign.serialize_for_client(header, tx)
-
-      ser_tx
-      |> encode_message("Transactions", source)
-      |> broadcast_tx(tx_pids)
-
-      obj_msg = encode_message(ser_tx, "Object", source)
+  defp serialize_tx(tx, %State{} = state) do
```
sborrazas, line 166: We can instead tag terms to avoid looking into opaque term structures. E.g. `{:state, state} | {:block, block}`

jyeshe: It's not looking to data, only to struct type but refactored to help your best friend dialyzer.

sborrazas: Thanks

sborrazas (approved):

PR #1190, opened 2023-02-17
chore: use mix releases for prod docker images

sborrazas:
This is done using [Mix.Task.Release](https://hexdocs.pm/mix/Mix.Tasks.Release.html) together with the
[auto-generated Dockerfile](https://hexdocs.pm/phoenix/Mix.Tasks.Phx.Gen.Release.html)

This helped reduce the image size to half its size, and it might also help with BEAM code loading efficiency.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -93,7 +93,7 @@ RUN mix release
 # the compiled release and other runtime necessities
 FROM ${RUNNER_IMAGE}
 
-RUN apt-get update -y && apt-get install -y libstdc++6 openssl libncurses5 locales libncurses5 libsodium-dev jq libgmp10 curl \
+RUN apt-get update -y && apt-get install -y libstdc++6 openssl libncurses5 locales libncurses5 libsodium-dev libgmp10 \
   && ldconfig \
   && apt-get clean && rm -f /var/lib/apt/lists/*_*
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,30 @@
-FROM hexpm/elixir:1.13.4-erlang-23.3.4.17-debian-buster-20220801
-# Add required files to download and compile only the dependencies
+# Find eligible builder and runner images on Docker Hub. We use Ubuntu/Debian instead of
+# Alpine to avoid DNS resolution issues in production.
+#
+# https://hub.docker.com/r/hexpm/elixir/tags?page=1&name=ubuntu
+# https://hub.docker.com/_/ubuntu?tab=tags
+#
+#
+# This file is based on these images:
+#
+#   - https://hub.docker.com/r/hexpm/elixir/tags - for the build image
+#   - https://hub.docker.com/_/debian?tab=tags&page=1&name=bullseye-20220801-slim - for the release image
+#   - https://pkgs.org/ - resource for finding needed packages
+#   - Ex: hexpm/elixir:1.13.4-erlang-23.3.4.17-debian-bullseye-20210902-slim
+#
+ARG ELIXIR_VERSION=1.13.4
+ARG OTP_VERSION=23.3.4.17
+ARG DEBIAN_VERSION=bullseye-20220801-slim
 
-# Install other required dependencies
+ARG BUILDER_IMAGE="hexpm/elixir:${ELIXIR_VERSION}-erlang-${OTP_VERSION}-debian-${DEBIAN_VERSION}"
+ARG RUNNER_IMAGE="debian:${DEBIAN_VERSION}"
+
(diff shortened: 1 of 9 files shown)


Dockerfile
```diff
@@ -1,7 +1,30 @@
...
+ARG RUNNER_IMAGE="debian:${DEBIAN_VERSION}"
+
+FROM ${BUILDER_IMAGE} as builder
+
+# install build dependencies
+RUN apt-get update -y && apt-get install -y build-essential git \
+    && apt-get clean && rm -f /var/lib/apt/lists/*_*
 
-# Install other required dependencies
 RUN apt-get -qq update && apt-get -qq -y install git curl libncurses5 libsodium-dev jq libgmp10 \
```
jyeshe, line 28: small detail that with the generated `apt-get update` above, this is not needed.

jyeshe (commented):
Considering the `docker-compose-dev.yml` replicates the local setup to `/app` dir I was wondering if we could automate its generation to avoid maintaining Dockerfile and Dockerfile.dev. This way we make sure that the tests, that switched to Dockerfile.dev runs in the same setup as the `prod`'s `Dockerfile`.

Except for the node lines, the file is almost the same as generated by `mix phx.gen.release --docker` so we could append some parts to include the Node.

Dockerfile
sborrazas: Done, merged both installations into one

sborrazas:
> Considering the `docker-compose-dev.yml` replicates the local setup to `/app` dir I was wondering if we could automate its generation to avoid maintaining Dockerfile and Dockerfile.dev. This way we make sure that the tests, that switched to Dockerfile.dev runs in the same setup as the `prod`'s `Dockerfile`.
> 
> Except for the node lines, the file is almost the same as generated by `mix phx.gen.release --docker` so we could append some parts to include the Node.

The dockerfile generated was the file generated by phx.gen.release merged with things needed by the node. This docker image is an empty docker image that contains very little useful things for development (almost none). You can't run elixir or any of the elixir/erlang utilities. This is because of the multi-stage dockerfile usage, which ends up generating a very minimal setup

jyeshe:
Yeah, I know it doesn't have mix for example. I mean automate the generation of Dockerfile.dev from the Dockerfile. Changing the runner and replacing the cmd with entrypoint, if not much many other changes in order to have the dev and test as closer as possible to prod.

sborrazas:
> Yeah, I know it doesn't have mix for example. I mean automate the generation of Dockerfile.dev from the Dockerfile. Changing the runner and replacing the cmd with entrypoint, if not much many other changes in order to have the dev and test as closer as possible to prod.

How would you suggest reusing one Dockerfile from the other? I think that dev and release usages are extremely different for it to be worth to perform hacks only to be able to re-use the elixir version, but I'm open to suggestions and code examples

jyeshe:
> > Yeah, I know it doesn't have mix for example. I mean automate the generation of Dockerfile.dev from the Dockerfile. Changing the runner and replacing the cmd with entrypoint, if not much many other changes in order to have the dev and test as closer as possible to prod.
> 
> How would you suggest reusing one Dockerfile from the other? I think that dev and release usages are extremely different for it to be worth to perform hacks only to be able to re-use the elixir version, but I'm open to suggestions and code examples

No hacks suggested, as the proposal above, wouldn't changing the runner and command be sufficient? This way the diff could go from 65 lines to possibly 2.

The goal would be that from dev to prod, and mainly from test to prod, the difference could be having or not a mix and the config that is already separated for each env.

sborrazas:
> > > Yeah, I know it doesn't have mix for example. I mean automate the generation of Dockerfile.dev from the Dockerfile. Changing the runner and replacing the cmd with entrypoint, if not much many other changes in order to have the dev and test as closer as possible to prod.
> > 
> > 
> > How would you suggest reusing one Dockerfile from the other? I think that dev and release usages are extremely different for it to be worth to perform hacks only to be able to re-use the elixir version, but I'm open to suggestions and code examples
> 
> No hacks suggested, as the proposal above, wouldn't changing the runner and command be sufficient? This way the diff could go from 65 lines to possibly 2.
> 
> The goal would be that from dev to prod, and mainly from test to prod, the difference could be narrowed down to having or not a mix and the config that is already separated for each env.

There's too many differences between dev/release, e.g.: the entrypoint, local hex and local rebar installation, the apt-get packages, the workdir. It wouldn't just be a matter of using a different image on the second docker stage

jyeshe (approved):
Approving for prod. I can work on unifying the Dockerfile for dev and test.

PR #1199, opened 2023-02-23
chore: replace Dockerfile.dev using mix runner

jyeshe:


sborrazas (approved):
LGTM we can then use scripts to reuse dockerfile functionality and keep dockerfiles separate as discussed later on

jyeshe:
> LGTM we can then use scripts to reuse dockerfile functionality and keep dockerfiles separate as discussed later on

As discussed without hackish includes and duplicated commands otherwise it will be better to keep a single Dockerfile for all envs.

Issue #108, opened 2020-11-17
Name swagger file consistent with Aeternity node


davidyuk:
The aeternity node's swagger file named `/api`, but the middleware's swagger file for some reason has path `/swagger/swagger.json`. Proposing to put it to the same place as the node's swagger file for consistency.

https://mainnet.aeternity.io/api
https://mainnet.aeternity.io/mdw/swagger/swagger.json

Also, the same naming using the compiler and the old middleware:
https://compiler.aepps.com/api
https://mainnet.aeternal.io/middleware/api

thepiwo:
I agree, this should be adjusted

thepiwo:
closing in favour of general swagger improvements issue

PR #1219, opened 2023-03-06
docs: use /api and /v2/api path for swagger spec files to match node

sborrazas:
refs #108
refs #674

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -82,14 +82,15 @@ COPY lib lib
 COPY scripts scripts
 COPY docs docs
 
-# Generate swagger V2 file
-RUN scripts/swagger-docs.py >priv/static/swagger/swagger_v2.yaml
-
 # Compile the release
 RUN mix compile
 
 # Changes to config/runtime.exs don't require recompiling the code
 COPY config/runtime.exs config/
+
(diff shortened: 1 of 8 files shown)


jyeshe (commented):
There is a also `code:priv_dir` that gives the priv directly but no big deal. Missing more the test cases to avoid manual testing.

sborrazas:
> There is also `code:priv_dir` that gives the priv directly but no big deal. Missing more the test cases to avoid manual testing.

Good call on the `priv_dir/1`. Also, I added unit/integration test cases for it

Dockerfile
```diff
@@ -82,15 +82,16 @@ COPY lib lib
...
-
 # Compile the release
 RUN mix compile
 
 # Changes to config/runtime.exs don't require recompiling the code
 COPY config/runtime.exs config/
 
+# Generate swagger V2 file
+RUN mix run --no-start -e 'IO.puts(Mix.Project.config[:version])' >AEMDW_VERSION
+RUN scripts/swagger-docs.py >priv/static/swagger/swagger_v2.yaml
```
jyeshe, line 93: :clap:

jyeshe (approved):
:gear:

PR #1222, opened 2023-03-07
chore: make web config runtime config

sborrazas:


config/dev.exs
```diff
@@ -1,33 +1,12 @@
...
 config :ae_mdw, :enable_livedashboard, true
 
 config :ae_mdw, TelemetryMetricsStatsd,
   host: "localhost",
   port: 8125
 
-# Endpoints
-protocol_opts = [max_request_line_length: 1_024, max_skip_body_length: 1_024]
-
-config :ae_mdw, AeMdwWeb.Endpoint,
```
jyeshe, line 16: Some of these are needed in dev, used differently from prod. I see there is a boolean flag on `debug_errors` in the `runtime.exs`. Could we add the same for `live_view`? Or just leave the endpoint declaration here in `dev.exs`.

config/runtime.exs
```diff
@@ -2,13 +2,39 @@ import Config
 
 env = config_env()
 
+# Chain
+config :aecore, network_id: System.get_env("NETWORK_ID", "ae_mainnet")
```
jyeshe, line 6: This is dev only without docker, keeping it in `dev.exs` would avoid conflicts in prod.

docker-compose-test.yml
```diff
@@ -14,4 +14,6 @@ services:
     volumes:
       - ${PWD}:/app
     environment:
-      - AETERNITY_CONFIG=/home/aeternity/aeternity.yaml
+      AETERNITY_CONFIG: /home/aeternity/aeternity.yaml
+      PORT: 4002
+      WS_PORT: 4003
```
jyeshe, line 19: Same as first comment.

.github/workflows/test-action.yml
```diff
@@ -43,6 +43,9 @@ jobs:
 
       - name: Execute coverage
         run: elixir --sname aeternity@localhost -S mix coveralls
+        env:
+          PORT: 4002
+          WS_PORT: 4003
```
jyeshe, line 48: This is not necessary because tests don't use the http server and ws is only accessed by test cases.

jyeshe (changes requested):
More flexibility for other setups, nice! Nothing operational that affects prod only some chore comments.

config/runtime.exs
sborrazas: Move this back to dev.config. Thanks

.github/workflows/test-action.yml
sborrazas: If I don't add them, these tests fail: https://github.com/aeternity/ae_mdw/blob/master/test/ae_mdw/websocket/socket_handler_test.exs#L10

config/dev.exs
sborrazas: Back to dev too

sborrazas:
> All good for me except that test ports could be setup in a single file.

moved compile-time config for tests back to test.exs

jyeshe (approved):

Issue #1141, opened 2023-01-23
APM instrumentation


jyeshe:
## Goal ## 
Turn errors and performance bottlenecks visible by instrumenting the Mdw to collect logs, metrics and spans.

## Parameters ##
- OpenTelemetry has become the main standard for this and could be used as much as there are corresponding stable libraries for each of the monitoring concerns.
- DataDog would be the initial APM service to be supported

## Notes ##
For now only collecting traces is stable with [OpenTelemetry.](https://opentelemetry.io/docs/instrumentation/erlang/) For logs and metrics there are alternatives using phoenix telemetry and exporters like FluentD.

dincho:
The infrastructure in development supports only Prometheus. Not sure where did you read about OpenTelemetry becoming a main standard, but that's not quite true IMO.

dincho:
In regards to logs, you should not worry about it but support console output + switchable machine readable format, usually JSON.

jyeshe:
> The infrastructure in development supports only Prometheus. Not sure where did you read about OpenTelemetry becoming a main standard, but that's not quite true IMO.

I get your point. Not referring to the de facto standard because there are multiple APM services which still use different protocols but for the APM that work on push mode (DataDog being our first APM to be supported), OpenTelemetry is currently the most discussed standardized approach (forums and Code BEAM talks for references) to export data on push mode. This on the protocol side while internally, to instrument the app there are also another set of libraries called [telemetry*](https://github.com/beam-telemetry) that are part of mainstream.

jyeshe:
> In regards to logs, you should not worry about it but support console output + switchable machine readable format, usually JSON.

Okay, could we have them exported? If not we can discuss using free tool. One of the goals would be to receive notifications when a 500 error, which we can also instrument on the app if you prefer. To be able to query for logs would be helpful in some situations to avoid asking for them from time to time (on a release basis at least).

jyeshe:
> The new infrastructure does not implement yet any tracing caps, but In future we'll add https://www.jaegertracing.io

No problem, meanwhile I will wrap them up as metrics for the duration of some inner workings.

dincho:
> Okay, could we have them exported? If not we can discuss using free tool. One of the goals would be to receive notifications when a 500 error, which we can also instrument on the app if you prefer. To be able to query for logs would be helpful in some situations to avoid asking for them from time to time (on a release basis at least).

[...]

PR #1225, opened 2023-03-09
chore: aggregate request metric per route

jyeshe:
refs #1141

### DIFF
--- a/lib/ae_mdw_web/plugs/request_span.ex
+++ b/lib/ae_mdw_web/plugs/request_span.ex
@@ -1,0 +1,34 @@
+defmodule AeMdwWeb.Plugs.RequestSpan do
+  @moduledoc """
+  Emits telemetry request duration events associating them to a route.
+  """
+
+  alias Plug.Conn
+
+  @type opts() :: []
+
+  @spec init(opts()) :: opts()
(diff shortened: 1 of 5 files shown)


lib/ae_mdw_web/plugs/request_span.ex
```diff
@@ -0,0 +1,34 @@
...
+    |> Conn.assign(:start_time, System.monotonic_time(:millisecond))
+    |> Conn.register_before_send(&emit_event_callback/1)
+  end
+
+  @spec emit_event_callback(Conn.t()) :: Conn.t()
+  def emit_event_callback(conn) do
+    duration = System.monotonic_time(:millisecond) - conn.assigns.start_time
+
+    with %{route: route} <-
+           Phoenix.Router.route_info(AeMdwWeb.Router, conn.method, conn.request_path, conn.host),
```
sborrazas, line 25: Can we pattern-match these out of the `Conn` struct? That way we get compiler errors if there's any typo

jyeshe: Sure, tests would fail but changed to pattern matching style.

sborrazas (approved):
thanks 🙏

PR #1237, opened 2023-03-16
fix: handle return type for failed GAMeta txs as well

sborrazas:


lib/ae_mdw/db/sync/transaction.ex
```diff
@@ -395,19 +395,23 @@ defmodule AeMdw.Db.Sync.Transaction do
     auth_id = :aega_meta_tx.auth_id(tx)
 
     case :aec_chain.get_ga_call(owner_pk, auth_id, block_hash) do
```
jyeshe, line 397: may we replace it by a `with`?

sborrazas: Good call. Done

jyeshe (approved):

PR #1251, opened 2023-04-03
chore: fix credo `refactor` errors

sborrazas:


lib/ae_mdw_web/validate.ex
```diff
@@ -1,88 +0,0 @@
-defmodule AeMdwWeb.Validate do
```
jyeshe, line 1: :pray:

jyeshe (approved):

Issue #755, opened 2022-06-29
Expose wealth ordered by account (top 100-500)


sborrazas:
refs #124

thepiwo:
as discussed this is quite complicated to be implemented, thus we won't consider it until it gains more importance to stakeholders

PR #1273, opened 2023-04-21
feat: wealth endpoint

jyeshe:
## What

Tracks wealth of 100 wallets with biggest balances.

## Why
refs #755 

## Notes
Next there will be: 
- the config of amount of wallets
- consider internal transfers and rewards
- automated tests of async tasks


lib/ae_mdw/sync/server.ex
```diff
@@ -298,15 +299,29 @@ defmodule AeMdw.Sync.Server do
...
+        |> maybe_enqueue_accounts_balance(block, block_index)
+        |> State.commit_db(block_mutations, clear_mem?)
       end)
 
     broadcast_blocks(gens_mutations)
 
     new_state
   end
 
+  defp maybe_enqueue_accounts_balance(state, _block, {_kbi, -1}) do
```
sborrazas, line 313: Key blocks would affect miner's balances, would it make sense to include them too?

sborrazas (commented):
Rest LGTM

lib/ae_mdw/sync/server.ex
jyeshe: Yes, gonna be implemented incrementally. Next there will be these items from the Notes:
- the config of amount of wallets
- consider internal transfers and rewards
- automated tests of async tasks

sborrazas: Cool, thanks for explaining

sborrazas (approved):

PR #1274, opened 2023-04-24
feat: consider existing oracles on inactive ones count

jyeshe:
Avoids double counting inactive oracles by checking if it was the first time the oracle was created (without previous).
Plus, decrements inactive ones that are reactivated.

### DIFF
--- a/lib/ae_mdw/db/mutations/stats_mutation.ex
+++ b/lib/ae_mdw/db/mutations/stats_mutation.ex
@@ -202,6 +202,8 @@ defmodule AeMdw.Db.StatsMutation do
 
     token_supply_delta = AeMdw.Node.token_supply_delta(height - 1)
     auctions_expired = get(state, :auctions_expired, 0)
+    new_oracles_expired = get(state, :new_oracles_expired, 0)
+    old_oracles_registered = get(state, :old_oracles_registered, 0)
 
     Model.total_stat(
       index: height,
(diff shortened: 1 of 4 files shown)


lib/ae_mdw/db/mutations/stats_mutation.ex
```diff
@@ -202,6 +202,8 @@ defmodule AeMdw.Db.StatsMutation do
 
     token_supply_delta = AeMdw.Node.token_supply_delta(height - 1)
     auctions_expired = get(state, :auctions_expired, 0)
+    new_oracles_expired = get(state, :new_oracles_expired, 0)
+    old_oracles_registered = get(state, :old_oracles_registered, 0)
```
sborrazas, line 206: Can we keep the notation used before (without the `old/new` prefix)?

sborrazas (commented):
I think we should re-purpose this PR as a feature, I don't think we should label it as a bug

lib/ae_mdw/db/mutations/stats_mutation.ex
jyeshe: Sorry, the description of the problem was missing. These are for:
- `new_oracles_expired`: avoids double counting inactive oracles by checking if it was the first time the oracle was created (without previous).
- `old_oracles_registered`: decrements inactive ones that are reactivated.

sborrazas: I hadn't realized oracles are mapped 1-to-1 to the address of the creator (unlike contracts or channels), meaning that the owner can create multiple oracles (not simultaneously, I guess?). Thank you for clearing it up

sborrazas (approved):

janmichek:
@sborrazas @jyeshe 
I am a bit confused here.

By the stats
https://mainnet.aeternity.io/mdw/v2/totalstats?limit=1
The sum of active and inactive is 33

From the listing, I see 57
https://mainnet.aeternity.io/mdw/v2/oracles?limit=100

Can you clarify, please?

jyeshe:
> @sborrazas @jyeshe I am a bit confused here.
> 
> By the stats https://mainnet.aeternity.io/mdw/v2/totalstats?limit=1 The sum of active and inactive is 33
> 
> From the listing, I see 57 https://mainnet.aeternity.io/mdw/v2/oracles?limit=100
> 
> Can you clarify, please?

Thank you for the feedback @janmichek. I am still working on the issue #1278.

PR #1277, opened 2023-04-28
feat: add channels updates nested endpoint

sborrazas:


lib/ae_mdw/channels.ex
```diff
@@ -102,6 +135,39 @@ defmodule AeMdw.Channels do
     end
   end
 
+  defp build_nested_streamer(state, updates, range, cursor) do
```
jyeshe, line 138: This is a naming nit pick as the output data comes from the list that was already fully fetched and is not transformed in a lazy way. Maybe sth like `build_updates_filter`?

test/ae_mdw_web/controllers/channel_controller_test.exs
```diff
@@ -407,4 +419,165 @@ defmodule AeMdwWeb.ChannelControllerTest do
...
+      with_mocks [
+        {DbUtil, [:passthrough],
+         [
+           read_node_tx_details: fn
+             _state, ^txi_idx1 ->
+               {channel_deposit_tx1, :channel_deposit_tx, tx_hash1, :channel_deposit_tx,
+                block_hash}
+
+             _state, ^txi_idx2 ->
+               {channel_settle_tx2, :channel_settle_tx, tx_hash2, :contract_call_tx, block_hash}
```
jyeshe, line 502: 🙏

test/ae_mdw_web/controllers/contract_controller_test.exs
```diff
@@ -978,16 +978,18 @@ defmodule AeMdwWeb.Controllers.ContractControllerTest do
...
+               {contract_create_tx2, :contract_create_tx, tx_hash2, :contract_call_tx, block_hash}
 
              _state, {^txi3, -1} ->
-               {contract_create_tx3, tx_hash3, :contract_create_tx, block_hash}
+               {contract_create_tx3, :contract_create_tx, tx_hash3, :contract_create_tx,
+                block_hash}
 
              _state, {^txi4, -1} ->
-               {ga_attach_tx, tx_hash4, :ga_attach_tx, block_hash}
+               {ga_attach_tx, :contract_create_tx, tx_hash4, :ga_attach_tx, block_hash}
```
jyeshe, line 992: With `{txi, -1}` it handles only `ga_meta_tx` and `paying_for` inner types. This would be a fix only on the test. I think `ga_attach` should continue without inner type tx.

jyeshe (commented):
LGTM. Only two minor details.

test/ae_mdw_web/controllers/contract_controller_test.exs
sborrazas: Could you explain a little bit further? What would you want the behavior to be here?

jyeshe: It's just about the consistency of the mocking with the behavior of `read_node_tx_details` that doesn't return inner type for `ga_attach_tx`:
https://github.com/aeternity/ae_mdw/blob/master/lib/ae_mdw/db/util.ex#L177-L184

`:ga_attach_tx` in place of `:contract_create_tx` would be a more realistic test.

sborrazas: With these branch changes, it does return `contract_create_tx` https://github.com/aeternity/ae_mdw/blob/channel-updates/lib/ae_mdw/db/util.ex#L178

jyeshe: Please notice the condition `tx_type in [:ga_meta_tx, :paying_for_tx]` (`tx_type = :ga_attach_tx` doesn't have inner tx)

sborrazas: You're right, I was confusing meta with attach. Done now, thanks!

jyeshe (approved):

Issue #1295, opened 2023-05-03
Oracle query and associated response


michele-franchi:
It would be nice to have a way to get the associated response of a query, if available, and ideally in a single endpoint.

With the following data:
- Query Tx
- Queried At
- Respond Tx
- Respond At
- Status: Response pending, Response available, Response expired, Query timed out (should be calculated based on the TTL)

I see two cases:

1. Query sent, no response sent
- Query id
- Query fee
- Query TTL

2. The query has a response
- Query id
- Query fee
- Query
- Response
- Response TTL

I'm not suggesting a structure, just the minimum ideal information that could be returned from the paginated endpoint.

sborrazas:
This is now done and deployed to prod

janmichek:
Seems like there are still some values missing to achieve what we want
https://user-images.githubusercontent.com/15363559/236007821-6a18fee4-65f4-4cb3-89a8-8553c19935de.png

Namely:
- Query block height
- Response block height 
- Status

sborrazas:
> Seems like there are still some values missing to achieve what we want https://user-images.githubusercontent.com/15363559/236007821-6a18fee4-65f4-4cb3-89a8-8553c19935de.png
> 
> Namely:
> 
> * Query block height
> * Response block height
> * Status

@janmichek height is now part of queries/responses. It is also included on the oracle detail for which `height < expire_height ? active : inactive`

PR #1314, opened 2023-05-12
feat: include oracle responses on oracle queries endpoints

sborrazas:
refs #1295

### DIFF
--- a/priv/migrations/20230509120001_index_oracle_queries_responses.ex
+++ b/priv/migrations/20230509120001_index_oracle_queries_responses.ex
@@ -1,0 +1,51 @@
+defmodule AeMdw.Migrations.IndexOracleQueriesResponses do
+  @moduledoc """
+  Indexes new Model.oracle_query response_txi_idx attribute.
+  """
+
+  alias :aeser_api_encoder, as: Enc
+  alias AeMdw.Db.WriteMutation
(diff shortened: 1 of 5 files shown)


priv/migrations/20230509120001_index_oracle_queries_responses.ex
```diff
@@ -0,0 +1,51 @@
...
+              Model.oracle_query(
+                index: index,
+                txi_idx: txi_idx,
+                response_txi_idx: response_txi_idx
+              )
+
+            WriteMutation.new(Model.OracleQuery, oracle_query)
+
+          :not_found ->
+            IO.puts("""
```
jyeshe, line 36: Could we log these so we can have a record of them on staging? When someone asks we could search for them on the log file.

jyeshe (commented):
Looking good with a single minor request.

priv/migrations/20230509120001_index_oracle_queries_responses.ex
sborrazas: These responses have already been logged here https://github.com/aeternity/ae_mdw/blob/master/lib/ae_mdw/db/mutations/oracle_response_mutation.ex#L66 and can be found on the logs. This is just to inform during migrating

jyeshe (approved):

PR #1326, opened 2023-05-16
feat: index inner contract creations for /contracts endpoint

sborrazas:


jyeshe (commented):
Could we have some test cases?

jyeshe (approved):

PR #1340, opened 2023-05-24
feat: include inactive channels on channels listing

sborrazas:
refs #1320

### DIFF
--- a/lib/ae_mdw/channels.ex
+++ b/lib/ae_mdw/channels.ex
@@ -32,7 +32,9 @@ defmodule AeMdw.Channels do
   @type channel_update() :: map()
 
   @table_active Model.ActiveChannel
-  @table_activation Model.ActiveChannelActivation
+  @table_active_activation Model.ActiveChannelActivation
+  @table_inactive Model.InactiveChannel
+  @table_inactive_activation Model.InactiveChannelActivation
 
   @channel_tx_mod %{
     :channel_create_tx => :aesc_create_tx,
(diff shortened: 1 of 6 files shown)


lib/ae_mdw/channels.ex
```diff
@@ -131,7 +133,13 @@ defmodule AeMdw.Channels do
 
   defp build_streamer(state, scope, cursor) do
     fn direction ->
-      Collection.stream(state, @table_activation, direction, scope, cursor)
+      [@table_active_activation, @table_inactive_activation]
```
jyeshe, line 136: what you think about using the inverse order? when it goes on the backwards default direction it would render the active ones firstly (maybe it's helpful on heights with many inactive ones).

jyeshe (commented):
Neat complement! Maybe only change the order of the tables.

lib/ae_mdw/channels.ex
sborrazas: It already returns results in backwards activation order. So the channels that were activated last will show up first by default

jyeshe: it shows the inactive firstly (on backwards) because of this order: `[@table_active_activation, @table_inactive_activation]`

sborrazas: `Collection.merge/2` merges results in the right order, it grabs the first key from both tables and then merges them accordingly, iteratively. I think what you're thinking of is `Stream.concat/2`

jyeshe (commented):
Some or all inactive records are shown before active ones for the same height. If possible with the same index it would be interesting to show the active ones firstly.

lib/ae_mdw/channels.ex
sborrazas: Since the table index is `{activation_height, channel_pk}`, when on equal heights, the first result shown is always the one following the `channel_pk` order:
```
ch1 = Validate.id!("ch_2dE8Co8wHju7qb46ezjBwUpvDr6gnQxr2vEjxpWwEsFC4jBurr")
ch2 = Validate.id!("ch_NAbNxGCpXSAzTGoXxKPnEmeWLzKYT5oN5fQqD6x1b4DQ77J5C")

ch2 < ch1
```
which is why on default order `ch_2dE8Co8wHju7qb46ezjBwUpvDr6gnQxr2vEjxpWwEsFC4jBurr` is shown first

jyeshe: Okay the `Collection.merge` doesn't preserve the order then. I will leave it to the final users if it's an important request.

PR #1367, opened 2023-05-30
feat: allow filtering channels by active/inactive

sborrazas:


jyeshe (commented):
Looks good. Missed only the swagger for 100%

jyeshe (approved):

PR #1369, opened 2023-05-30
chore: track internal calls for wealth

jyeshe:


lib/ae_mdw/sync/async_tasks/wealth_rank_accounts.ex
```diff
@@ -0,0 +1,110 @@
...
+    |> Enum.filter(&is_struct(&1, IntCallsMutation))
+    |> Enum.flat_map(fn %IntCallsMutation{int_calls: int_calls} ->
+      int_calls
+      |> Enum.filter(fn {_idx, fname, _type, _aetx, _tx} ->
+        fname in ["Chain.spend", "Call.amount"]
+      end)
+      |> Enum.flat_map(fn {_idx, _fname, _type, aetx, _tx} ->
+        aetx_accounts(aetx)
+      end)
+    end)
```
sborrazas, line 81: Perhaps we can have `Sync.Block.block_mutations/4` return these addresses to avoid having to inspect so many internal data structures. Something like
```elixir  
@typep block_mutations() :: {Blocks.block_index(), block(), [Mutation.t()], [pubkey()]}
```
where the last item from the tuple would be all of the updated addresses

sborrazas (commented):
Are we planning to do a full sync to retrieve all of the addresses that weren't considered for the wealth?

lib/ae_mdw/sync/async_tasks/wealth_rank_accounts.ex
jyeshe: Honestly I think the return of `Sync.Block.block_mutations/4` is complex enough:
```
@typep block_mutations() :: {Blocks.block_index(), block(), [Mutation.t()]}
@type height_mutations() :: {Blocks.height(), [block_mutations()]}

@spec blocks_mutations(...) :: {[height_mutations()], Txs.txi()}
```
I think it returns what is is asked for the function so I would rather filter the mutations.

jyeshe:
> Are we planning to do a full sync to retrieve all of the addresses that weren't considered for the wealth?

Along with Node 6.8.1

lib/ae_mdw/sync/async_tasks/wealth_rank_accounts.ex
sborrazas: We can rename the function or return a more explicit data structure using structs. But if you want to leave it like this I won't block it

sborrazas:
> > Are we planning to do a full sync to retrieve all of the addresses that weren't considered for the wealth?
> 
> Along with Node 6.8.1

6.8.1 wouldn't require a sync from scratch, it does not add any new information to the database

jyeshe:
> > > Are we planning to do a full sync to retrieve all of the addresses that weren't considered for the wealth?
> > 
> > 
> > Along with Node 6.8.1
> 
> 6.8.1 wouldn't require a sync from scratch, it does not add any new information to the database

With a new Node I would consider it highly desirable.

sborrazas (approved):

Issue #1364, opened 2023-05-30
`type: string` missed near `enum`


davidyuk:
$.components.schemas.Auction.properties.info.properties.status

> The enum schema 'Auction-info-status' with an undefined type and enum values is ambiguous. This has been auto-corrected to 'type:string'
      
$.components.schemas.Transfer.properties.kind

> The enum schema 'Transfer-kind' with an undefined type and enum values is ambiguous. This has been auto-corrected to 'type:string'

$.paths./names/search.get.responses.200.content.application/json.schema.allOf[0].properties.data.items.allOf[1].properties.status

 >  The enum schema 'get-200-application-json-allOf-properties-items-1-status' with an undefined
      type and enum values is ambiguous. This has been auto-corrected to 'type:string'

$.paths./txs.get.parameters[0].schema.items
>  The enum schema 'get-0-itemsItem' with an undefined type and enum values is ambiguous.
      This has been auto-corrected to 'type:string'
      
https://swagger.io/docs/specification/data-models/enums/

sborrazas:
This has been fixed and can be accessed in https://mainnet.aeternity.io/mdw/v2/api

PR #1376, opened 2023-06-02
docs: add type:string to enum types that are missing it

sborrazas:
refs #1364

### DIFF
--- a/docs/swagger_v2/activities.spec.yaml
+++ b/docs/swagger_v2/activities.spec.yaml
@@ -57,7 +57,6 @@ schemas:
           - reward_dev
           - reward_block
           - fee_refund_oracle
-          - fee_lock_name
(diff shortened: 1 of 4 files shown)


docs/swagger_v2/activities.spec.yaml
```diff
@@ -57,7 +57,6 @@ schemas:
           - reward_dev
           - reward_block
           - fee_refund_oracle
-          - fee_lock_name
```
jyeshe, line 60: noticed this is different from the `type: string` error, is it a wanted change?

sborrazas: It's not, reverted it to include it on separate PR. Thanks

jyeshe (approved):

PR #1382, opened 2023-06-05
chore: add transactions count to websocket keyblock

### LINKED ISSUE #1291: Websocket serves different data than the REST endpoints
Hey folks,

# Background: 
I'm implementing a live keyblocks, microblocks and transactions browser app that presents recently minted objects in the blockchain. On initialization of my browser I am fetching the state from REST API and then apply further updates through websocket subscription. Unfortunately, these sources of data don't serve identical content which leads to misrepresenting the real state of the blockchain

# Issues:

## Sending websocket message before an object is indexed in the db (?): 
From my observation, it looks like the WebSocket messages are sent immediately after receiving information about it from a Node. Indexing that data in the database seems to take between tens of ms to a few seconds. 

### Issue 1 - wrong transaction count
As a result, often a websocket client receives information about transactions, while REST API `/keyblocks` endpoint still does not account for their existence.

image
image

I didn't spot a similar problem with the microblocks count

### Issue 2 - missing properties
Keyblock and Microblock messages received through the websocket do not contain microblock_count and transaction_count properties. As far as I know, we are aiming towards unifying the responses from REST API and the websocket, so this is a major discrepancy that requires websocket messages to be augmented on the client side. 

### Issue 3 - receiving transactions and microblocks which are later moved to another keyblock
Just before minting a new keyblock websocket usually sends up to a few microblocks and transactions assigned to the old keyblock, which are later orphaned and moved to the freshly minted keyblock. It would be nice to either prevent such a situation completely or at least send a message informing that given transactions and keyblocks were reassigned to another keyblock.

## Adding indexing information to REST endpoints
Since a request to REST API might take a different time than a websocket message to start subscribing events related to a given object, there always will be a chance for unsynchronized communication. Therefore: 

### Issue 4 - there is no way to sync data between REST endpoint and websocket messages with 100% accuracy
I think this issue might be resolved by adding information to /keyblocks and /microblocks endpoints about which microblock and transaction has been indexed recently so we can compare it with the first message received from the websocket. If there is a mismatch, we can perform another request to REST API to sync everything.

---

As I mentioned multiple problems with the current APIs, I guess we should treat this issue as a general discussion about these problems and then create more granular issues to implement the results of that discussion.

jyeshe:
closes #1291

### DIFF
--- a/lib/ae_mdw/application.ex
+++ b/lib/ae_mdw/application.ex
@@ -15,7 +15,7 @@ defmodule AeMdw.Application do
   alias AeMdw.NodeHelper
   alias AeMdw.Sync.Watcher
   alias AeMdw.Util
-  alias AeMdwWeb.Websocket.Broadcaster
+  alias AeMdwWeb.Websocket.BroadcasterCache
 
   require Model
 
@@ -176,8 +176,7 @@ defmodule AeMdw.Application do
   defp init(:aesync), do: Application.ensure_all_started(:aesync)
 
   defp init(:tables) do
-    {ets_table, ets_expiration} = Broadcaster.ets_config()
-    EtsCache.new(ets_table, ets_expiration)
+    BroadcasterCache.init()
 
     AeMdw.Sync.AsyncTasks.Stats.init()
     AeMdw.Sync.AsyncTasks.Store.init()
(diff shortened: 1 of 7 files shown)


sborrazas (approved):

janmichek:
Hi, I am taking over this issue. I think this was a mistaken requirement. The count of transactions per keyblock should be served in the microblock-triggered message. 
When the keyblock is mined and the message fired, tx count and microblocks count will always be 0 / null. Right?

image

janmichek:
ping

Issue #1278, opened 2023-04-28
Count inactive names and oracles once per object


jyeshe:
`/v2/totalstats` inherited from `v1` the inactive names and oracles counters that are incremented per expiration. However both names and oracles might expire multiple times (oracle considering the pubkey, not the oracle id). 

`/v2/totalstats` needs to be improved to consider only new expirations and decrement the inactive counter when the object is reactivated.

lukeromanowicz:
Looking at the /v2/oracles endpoint it currently returns only 58 oracles, while the totalstats endpoint reports:

active_oracles:32
inactive_oracles: 340

I'm afraid one of the endpoint is working incorrectly

jyeshe:
> Looking at the /v2/oracles endpoint it currently returns only 58 oracles, while the totalstats endpoint reports:
> 
> active_oracles:32
> inactive_oracles: 340
> 
> I'm afraid either one of the endpoints is working incorrectly or I misuse/misunderstand how they work and what they return

We need a full resync to see the correct result. Thanks for double checking Luke.

jyeshe:
@lukeromanowicz cc @michele-franchi the Mdw is updated with a counting that depends on actual keys (not on operations) so it should be fine by now (for latest generations, after 824600, for previous ones it's needed to wait for a full resync).

PR #1397, opened 2023-06-14
fix: consider all oracles expirations

jyeshe:
refs #1278

### DIFF
--- a/lib/ae_mdw/db/mutations/oracles_expiration_mutation.ex
+++ b/lib/ae_mdw/db/mutations/oracles_expiration_mutation.ex
@@ -78,7 +78,6 @@ defmodule AeMdw.Db.OraclesExpirationMutation do
           |> SyncOracle.cache_through_write(Model.InactiveOracleExpiration, m_exp)
           |> SyncOracle.cache_through_delete(Model.ActiveOracle, pubkey)
           |> State.inc_stat(:oracles_expired)
-          |> maybe_increment_new_oracles_expired(Model.oracle(m_oracle, :previous) == nil)
         else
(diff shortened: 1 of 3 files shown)


sborrazas (commented):
Should this change include a migration?

jyeshe:
> Should this change include a migration? 
Reindexation will be required to consider the ` old_oracles_registered`.

sborrazas:
> > Should this change include a migration?
> 
> Reindexation will be required to consider the `old_oracles_registered`.

I would imagine there are not that many oracle registrations/expirations, perhaps we can iterate through registrations/expirations and build a count for each generation and then update the total/delta? Similar to how it was done here for example: https://github.com/aeternity/ae_mdw/commit/d965275c5173e8406ceb6f546bb054a38f30df2d#diff-6bec92ed38dd3e93f9c2600420a2eca3d0b410d013b67b4621c61cfbfca967f1

jyeshe:
> > > Should this change include a migration?
> > 
> > 
> > Reindexation will be required to consider the `old_oracles_registered`.
> 
> I would imagine there are not that many oracle registrations/expirations, perhaps we can iterate through registrations/expirations and build a count for each generation and then update the total/delta? Similar to how it was done here for example: [d965275#diff-6bec92ed38dd3e93f9c2600420a2eca3d0b410d013b67b4621c61cfbfca967f1](https://github.com/aeternity/ae_mdw/commit/d965275c5173e8406ceb6f546bb054a38f30df2d#diff-6bec92ed38dd3e93f9c2600420a2eca3d0b410d013b67b4621c61cfbfca967f1)

The issue is that we need to simulate the registrations, registrations of only new oracles and expirations (some are registered and expires multiple times).

Issue #1371, opened 2023-05-31
Display `approximate_expiration_time` together with expiration


sborrazas:
This is a follow up of #831 

This applies to:

* Names
* Auctions
* Oracles

and should be added to each of the endpoints that expose their expiration

sborrazas:
Add the following fields to include the date:
- `approximate_last_updated_time` in `channels/{stateChannelId}`
- `approximate_last_updated_time` in `channels` endpoint
- `approximate_activated_time` in `names` (with expand=true) endpoint
- `approximate_auction_end_time` in `names/auctions` endpoint
- `approximate_expiration_time` and `approximate_auction_end_time` in `names/{name}` endpoint
- `approximate_created_time` in `accounts/{nameHash}/activities` (for cases if `micro_time` is not available) endpoint
- `approximate_created_time`  in  `contracts/logs?contract_id={contractId}` endpoint
- `approximate_activated_time` in `oracles` endpoint
- `approximate_registered_time` in `oracles/{oracleId}` endpoint

PR #1403, opened 2023-06-19
feat: display approximate expiration time for auctions

sborrazas:
refs #1371


jyeshe (commented):
May we have a test case for an active auction, where the `approximate_expire_time` is an estimated (different from block's `time_in_msecs`)?

jyeshe (approved):

PR #1404, opened 2023-06-19
fix: only bootstrap accounts for configured hardforks

seanhinde:
For hyperchains it's likely we start directly with ceres so there won't be account bootstrap files for the earlier heights.

This PR was sponsored by the ACF


jyeshe (approved):

jyeshe:
To pass CI lint check the 'Include' on last commit needs to be renamed to 'include'

PR #1260, opened 2023-04-07
feat: track aex9 token holders count per contract

jyeshe:
Resolves #1257

Issue #1257, opened 2023-04-05
Expose number of aex9 token holders in token information endpoint


lukeromanowicz:
Hey,
currently to get the number of aex9 token holders we need to paginate through all of the `/v2/aex9/:contract_id/balances` pages which on the frontend is very far from ideal. It's easy to imagine a token airdropped to multiple accounts that will generate a long balances list resulting in killing the device that tries to go through that list (not to mention the time and bandwidth needed to transfer all of that information). 

Could we expose the number of token holders in `/v2/aex9/:contract_id` to save the hustle and optimize the performance on the client-side? :)

jyeshe:
Aex9 compliance to standard to be classified by #1410

michele-franchi:
@jyeshe I see a PR for the fix was merged but I noticed a difference between the count `holders` and the balances endpoint.
Example:
- https://mainnet.aeternity.io/mdw/v2/aex9/ct_xtk8rSz9suPb6D6VLquyfVji25FcnFRDjn3dnn5mmvHsPiESt returns `"holders": 38`
- https://mainnet.aeternity.io/mdw/v2/aex9/ct_xtk8rSz9suPb6D6VLquyfVji25FcnFRDjn3dnn5mmvHsPiESt/balances?limit=100 returns 40 entries

janmichek:
Hi, this does not seem to be resolved yet.

The example above still shows 63 entries and 62 in `holders` field.

Here is another example

![image](https://github.com/aeternity/ae_mdw/assets/15363559/b136c4fc-0bf4-4466-8126-7ec40b321822)

janmichek:
@jyeshe ping

jyeshe:
> @jyeshe ping

Need to hold...

janmichek:
Hello, do we have some updates on this?

janmichek:
There is a difference between length for balance endpoint for V2 and V3. Currently we are using v3. Should I downgrade to v2 so the length and count are the same?

https://mainnet.aeternity.io/mdw/v3/aex9/ct_xtk8rSz9suPb6D6VLquyfVji25FcnFRDjn3dnn5mmvHsPiESt/balances?by=amount&limit=100

https://mainnet.aeternity.io/mdw/v2/aex9/ct_xtk8rSz9suPb6D6VLquyfVji25FcnFRDjn3dnn5mmvHsPiESt/balances?limit=100

vatanasov:
[...]

PR #1409, opened 2023-06-20
chore: return nil for invalid amount of aex9 holders

jyeshe:
What/Why

Negative number is confusing for users: https://github.com/aeternity/ae_mdw/pull/1260#issuecomment-1591093770
Returns nil (for non-applicable) instead.
refs #1257

### DIFF
--- a/lib/ae_mdw_web/views/aexn_view.ex
+++ b/lib/ae_mdw_web/views/aexn_view.ex
@@ -135,7 +135,7 @@ defmodule AeMdwWeb.AexnView do
 
     num_holders =
       with num when num < 0 <- Stats.fetch_aex9_holders_count(state, contract_pk) do
-        "NA"
+        nil
       end
 
     %{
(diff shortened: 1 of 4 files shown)


lib/ae_mdw_web/views/aexn_view.ex
```diff
@@ -142,7 +147,7 @@ defmodule AeMdwWeb.AexnView do
       extensions: extensions,
       initial_supply: initial_supply,
       event_supply: event_supply,
-      holders: Stats.fetch_aex9_holders_count(state, contract_pk)
+      holders: num_holders
```
sborrazas, line 150: perhaps we can use `nil` instead, since it would be a bit confusing for it to be either a string or a number

jyeshe: The purpose is to catch user's attention as much as possible. But we can start with `nil` documenting it. If it continues to pop up like the balance before I change for `NA`.

jyeshe: Actually another thing that would have being useful for a while, is to tag contracts as compliant or not. Opened here: https://github.com/aeternity/ae_mdw/issues/1410

sborrazas: Agreed, that would be very useful

sborrazas (approved):

Issue #1418, opened 2023-06-26
Add block height and time to oracle queries


michele-franchi:
It would be great to have block height and time to both query and response in oracle `/queries` endpoint.

sborrazas:
Done

PR #1419, opened 2023-06-26
feat: add micro block time to oracle nested resources

sborrazas:
refs #1418

### DIFF
--- a/lib/ae_mdw/node/db.ex
+++ b/lib/ae_mdw/node/db.ex
@@ -258,6 +258,9 @@ defmodule AeMdw.Node.Db do
     end
   end
 
+  @spec get_block(Blocks.block_hash()) :: key_block() | micro_block()
+  def get_block(block_hash), do: :aec_db.get_block(block_hash)
+
   defp block_accounts_tree(mb_hash) do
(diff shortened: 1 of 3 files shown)


lib/ae_mdw/node/db.ex
```diff
@@ -258,6 +258,9 @@ defmodule AeMdw.Node.Db do
     end
   end
 
+  @spec get_block(Blocks.block_hash()) :: key_block() | micro_block()
+  def get_block(block_hash), do: :aec_db.get_block(block_hash)
```
jyeshe, line 262: :pray: for a motivation to be praised

jyeshe: What you think about making public a function that calls `block_hash |> get_block() |> :aec_blocks.time_in_msecs()`?

jyeshe: Besides encapsulating the Node it could simplify the tests as well.

jyeshe (commented):
LGTM, minor suggest

lib/ae_mdw/node/db.ex
sborrazas: Good call. Done

jyeshe (approved):

Issue #1385, opened 2023-06-09
Some name activities are missing micro_time


lukeromanowicz:
Some name activities are missing micro_time in the event's payload e.g. InternalContractCallEvent. Perhaps there are more cases but that's the only one I spotted so far

![image](https://github.com/aeternity/ae_mdw/assets/46789227/3d713458-b912-466a-b8cb-ed3e7aec55ed)

sborrazas:
@lukeromanowicz the payload depends on the event type. So the fact that there are SOME attributes on some given payload type does not mean they will be present on others.

That said, we can probably move `micro_time` outside the payload, which would be the time at which the `block_hash` happened.

sborrazas:
@lukeromanowicz all activities should now have `block_time` at the outermost level

PR #1432, opened 2023-06-29
feat: add block_time to all activities

sborrazas:
refs #1385

### DIFF
--- a/lib/ae_mdw/activities.ex
+++ b/lib/ae_mdw/activities.ex
@@ -663,10 +663,13 @@ defmodule AeMdw.Activities do
   end
 
   defp render_payload(state, _account_pk, _height, txi, {:claim, local_idx}) do
+    Model.tx(time: micro_time) = State.fetch!(state, Model.Tx, txi)
+
     {claim_aetx, :name_claim_tx, tx_hash, tx_type, _block_hash} =
       DbUtil.read_node_tx_details(state, {txi, local_idx})
 
     payload = %{
+      micro_time: micro_time,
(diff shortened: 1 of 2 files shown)


lib/ae_mdw/activities.ex
```diff
@@ -657,13 +663,10 @@ defmodule AeMdw.Activities do
   end
 
   defp render_payload(state, _account_pk, _height, txi, {:claim, local_idx}) do
-    Model.tx(time: micro_time) = State.fetch!(state, Model.Tx, txi)
-
     {claim_aetx, :name_claim_tx, tx_hash, tx_type, _block_hash} =
       DbUtil.read_node_tx_details(state, {txi, local_idx})
 
     payload = %{
-      micro_time: micro_time,
```
jyeshe, line 666: Better keep it to avoid potential breaking changes (added more than 6 months ago)

test/ae_mdw_web/controllers/activities_controller_test.exs
```diff
@@ -966,10 +1030,10 @@ defmodule AeMdwWeb.ActivitiesControllerTest do
                  "height" => ^height1,
                  "type" => "NameClaimEvent",
                  "block_hash" => ^enc_mb_hash1,
+                 "block_time" => 456,
```
jyeshe, line 1033: may it have a different value for mb_hash1 and mb_hash2?

lib/ae_mdw/activities.ex
sborrazas: Good call, done

jyeshe (approved):

Issue #1454, opened 2023-07-07
`block_time` is not defined in openapi


davidyuk:
It was added in https://github.com/aeternity/ae_mdw/pull/1432.
I'm getting `block_time` in currently-deployed version (1.53.0), but this field definition is also missed in 1.54.1.

thepiwo:
I assume this can be closed @sborrazas

sborrazas:
Yes, thank you

PR #1455, opened 2023-07-07
docs: add block time to newly added places

sborrazas:
refs #1454

### DIFF
--- a/docs/swagger_v2/channels.spec.yaml
+++ b/docs/swagger_v2/channels.spec.yaml
@@ -37,6 +37,10 @@ schemas:
         type: string
         description: The transaction type of the last transaction that updated the channel
         example: ChannelDepositTx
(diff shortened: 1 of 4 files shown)


docs/swagger_v2/channels.spec.yaml
```diff
@@ -37,6 +37,10 @@ schemas:
         type: string
         description: The transaction type of the last transaction that updated the channel
         example: ChannelDepositTx
+      last_updated_time:
+        type: integer
+        description: The block time in which the channel was last updated
```
jyeshe, line 42: maybe "updated on chain" to distinguish

sborrazas: Done

jyeshe (approved):

PR #1459, opened 2023-07-12
feat: add v3 initial endpoints for names/auctions

sborrazas:


lib/ae_mdw/auction_bids.ex
```diff
@@ -95,6 +109,37 @@ defmodule AeMdw.AuctionBids do
...
+    last_bid = Txs.fetch!(state, bi_txi_idx_txi(last_bid))
+    name_ttl = Names.expire_after(expire_height)
+
+    %{
+      name: plain_name,
+      activation_time: DbUtil.block_index_to_time(state, block_index),
+      auction_end: expire_height,
+      approximate_expire_time:
+        DbUtil.height_to_time(state, expire_height, last_gen, last_micro_time),
+      last_bid: put_in(last_bid, ["tx", "ttl"], name_ttl)
```
jyeshe, line 132: let's remove the `tx_index` from all responses?

jyeshe (commented):
Looking good, just might be the time to start replacing the txi with tx_hash for every response.

lib/ae_mdw/auction_bids.ex
sborrazas: Good call, done

jyeshe (approved):

PR #1464, opened 2023-07-18
chore: move nested name records to individual tables

sborrazas:


lib/ae_mdw/auction_bids.ex
```diff
@@ -113,16 +114,21 @@ defmodule AeMdw.AuctionBids do
...
+           expire_height: expire_height
          ),
          last_gen,
          last_micro_time,
          _opts
        ) do
+    last_bid_txi =
+      state
+      |> Name.stream_nested_resource(Model.AuctionBidClaim, plain_name)
+      |> Enum.at(0)
```
jyeshe, line 126: It's already readable now but there are 3 uses of stream and `Enum.at(0)`. Maybe we could have a new function that picks the first?

lib/ae_mdw/db/name.ex
```diff
@@ -112,40 +119,61 @@ defmodule AeMdw.Db.Name do
...
-  def pointers(state, Model.name(updates: [{_block_index, txi_idx} | _rest_updates])) do
+  @spec stream_nested_resource(state(), nested_table(), plain_name(), height()) :: Enumerable.t()
+  def stream_nested_resource(state, table, plain_name, height) do
     state
-    |> DbUtil.read_node_tx(txi_idx)
-    |> :aens_update_tx.pointers()
-    |> Enum.into(%{}, &pointer_kv_raw/1)
-    |> Format.encode_pointers()
+    |> Collection.stream(table, :backward, nil, {plain_name, @max_int, {@max_int, @max_int}})
+    |> Stream.take_while(&match?({^plain_name, ^height, _txi_idx}, &1))
```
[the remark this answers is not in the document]
sborrazas: Done

lib/ae_mdw/db/name.ex
```diff
@@ -155,8 +183,20 @@ defmodule AeMdw.Db.Name do
       nil ->
         {:error, :name_not_found}
 
-      {m_name, _module} ->
-        pointee_at(state, m_name, time_reference_txi)
+      {_m_name, _module} ->
+        update_txi_idx =
+          state
+          |> stream_nested_resource(Model.NameUpdate, plain_name)
+          |> Stream.drop_while(
```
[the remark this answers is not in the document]
sborrazas: Fixed, thanks

lib/ae_mdw/auction_bids.ex
sborrazas: I needed to add 2 new functions for both stream_nested_resource/3 and /4. Or add a single function that only did `Enum.at(stream, 0)`. I don't think it's worth it either way

jyeshe: There is one that is doing it already, the [last_update](https://github.com/aeternity/ae_mdw/blob/1333d624a9a23309fb5ae27a453914199abeb175/lib/ae_mdw/db/name.ex#L149). This [ownership case](https://github.com/aeternity/ae_mdw/blob/1333d624a9a23309fb5ae27a453914199abeb175/lib/ae_mdw/db/name.ex#L169) would be a little bit more readable.

jyeshe (approved):
Approving with a readability alternative.

lib/ae_mdw/auction_bids.ex
sborrazas: You're right, done

PR #1481, opened 2023-07-27
feat: filter owned nfts by collection

### LINKED ISSUE #1479: get NFTs owned by a specific account for a specific contract/collection
I just saw some nft related PR and recently I wanted to fetch this info. I think it would be great being able to fetch this data.

maybe just extend following endpoint with a filter option for a contract/collection?

`GET /v2/aex141/owned-nfts/:account_id`

probably another endpoint would be better as the response format would most likely differ. in case of provided contract id we could just return a list of ids.

currently the response includes contract and owner for each nft:

```json
{
   "data":[
      {
         "contract_id":"ct_2cFUyFZfnCFXrUvW9e3i8sa3w8qU7YRoZDai4DEmSWyr4ty1jB",
         "owner_id":"ak_dwYGBWwLbuMTSmg8BL1STffp1Hnk4qziQYhSE36Y41vYFsyqS",
         "token_id":1
      }
   ],
   "next":null,
   "prev":null
}
```

jyeshe:
resolves #1479

### DIFF
--- a/lib/ae_mdw_web/controllers/aex141_controller.ex
+++ b/lib/ae_mdw_web/controllers/aex141_controller.ex
@@ -124,10 +124,9 @@ defmodule AeMdwWeb.Aex141Controller do
   end
 
   defp validate_optional_pubkey(params, param_name) do
-    case params |> Map.get(param_name) |> Validate.id() do
-      {:ok, pubkey} -> {:ok, pubkey}
-      {:error, {AeMdw.Error.Input.Id, nil}} -> {:ok, nil}
-      {:error, reason} -> {:error, reason}
+    case Map.fetch(params, param_name) do
+      {:ok, id} -> Validate.id(id)
+      :error -> {:ok, nil}
     end
   end
 end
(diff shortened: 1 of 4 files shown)


lib/ae_mdw_web/controllers/aex141_controller.ex
```diff
@@ -108,17 +108,26 @@ defmodule AeMdwWeb.Aex141Controller do
...
+         {:ok, contract_pk} <- validate_optional_pubkey(params, "contract"),
          {:ok, {prev_cursor, nfts, next_cursor}} <-
-           Aex141.fetch_owned_nfts(state, account_pk, cursor, pagination) do
+           Aex141.fetch_owned_nfts(state, account_pk, contract_pk, cursor, pagination) do
       WebUtil.paginate(conn, prev_cursor, nfts, next_cursor)
     end
   end
+
+  defp validate_optional_pubkey(params, param_name) do
+    case params |> Map.get(param_name) |> Validate.id() do
```
sborrazas, line 127: Perhaps we can avoid having `Validate.id` handle any value type and use `Map.fetch(params, param_name)` here to tell if it's present or not

sborrazas (approved):

Issue #1492, opened 2023-08-01
Statistics endpoint definition for transaction types


sborrazas:
The following outlines the specifications for the statistics endpoint, designed to provide chartable data that can be seamlessly displayed in a chart.

These endpoints offer both filtering and pagination functionalities, enabling the efficient retrieval of results. With these features, users can easily access the specific data they need, streamlining the charting process and enhancing the overall user experience.

All endpoints support the following query params:

* `interval` - Specifies the interval between values. This can be `generation`, `daily`, `weekly` or `monthly`. Defaults to `daily`.
* `scope` - If given, scopes the results between the two dates. In the format of `1590883200000-1607817600000`. Defaults to 3 years back from the current date/generation. **TBD** according to desired server capacity.
* `direction` - `forward` or `backward` (default `backward`)

## Endpoints

The structure for all endpoints follow the same structure as any cursor-based paginated endpoint, which includes `prev`, `next` and `data`.
In this case, each data object would include the `start_time` in which that interval started and the `count` of objects resulting up until the next start time.

```
{
  "data": [
    {
      "start_time": 1690901323472,
      "count": 18
    },
    ...
  ]
  "prev": null,
  "next": "/statistics/txs?cursor=<next-cursor>"
}
```

## `/statistics/txs`

Displays the amount of transactions (total) or filtered transaction type on a given timeframe.

Additional query params:

* `tx_type` - If given, it will filter only the transactions of that type (e.g. `tx_type=contract_call`)

## `/statistics/aexn-transfers`

Displays amount of calls made to aexn contracts (both aex9 and ax141).

Additional query params:

* `type` - `type=aex9` or `type=aex141`.

**TODO**: Do we want to keep track of tranfers, or contract calls made to AExN contracts?

## `/statistics/blocks`

Displays amount of key-blocks (or micro-blocks) on the given timeframe, which allows to see the historial rate in which they were being added to the chain.

Additional query params:

* `type` - `type=key` or `type=micro`.

michele-franchi:
Good analysis @sborrazas, I have some questions.

1) I see you mentioned pagination but how would it exactly work? For example if a frontend app wants to display a chart with the amount of transactions of the last 365 days it would mean it has to extract 365 records and if this is done with multiple HTTP requests it would probably takes a lot of time, compared to a single HTTP call.

[...]

PR #1496, opened 2023-08-03
feat: add statistics and /statistics/transactions endpoint

sborrazas:
refs #1492

### DIFF
--- a/lib/ae_mdw/db/sync/stats.ex
+++ b/lib/ae_mdw/db/sync/stats.ex
@@ -3,8 +3,11 @@ defmodule AeMdw.Db.Sync.Stats do
   Update general and per contract stats during the syncing process.
   """
 
+  alias AeMdw.Blocks
   alias AeMdw.Db.Model
   alias AeMdw.Db.State
+  alias AeMdw.Db.StatisticsMutation
+  alias AeMdw.Node
   alias AeMdw.Stats
   alias AeMdw.Util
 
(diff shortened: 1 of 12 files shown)


lib/ae_mdw/stats.ex
```diff
@@ -209,6 +223,87 @@ defmodule AeMdw.Stats do
...
+      {{:transactions, tx_tag}, :day, Util.min_int()},
+      {{:transactions, tx_tag}, :day, Util.max_int()}
+    }
+
+    fn direction ->
+      Collection.stream(state, Model.Statistic, direction, key_boundary, cursor)
+    end
+  end
+
+  defp render_statistic(state, {{:transactions, _tx_tag}, :day, interval_start} = statistic_key) do
```
jyeshe, line 263: For me it's okay if you wanna add `:month` and `:year` to the same PR, but it's your choice.

lib/ae_mdw/db/sync/stats.ex
```diff
@@ -51,6 +59,44 @@ defmodule AeMdw.Db.Sync.Stats do
         &(&1 - 1)
       )
 
+  @spec statistics_mutations(time(), type_counts()) :: StatisticsMutation.t() | nil
+  def statistics_mutations(time, type_counts) do
```
jyeshe, line 63: may we have a unit test for this syncing function and rename it for the purpose of the transactions stats?

lib/ae_mdw/stats.ex
```diff
@@ -30,9 +34,17 @@ defmodule AeMdw.Stats do
...
   @typep aexn_type() :: :aex9 | :aex141
+  @typep query() :: map()
+  @typep pagination() :: Collection.direction_limit()
+  @typep pagination_cursor() :: Collection.pagination_cursor()
 
   @type nft_stats :: %{nfts_amount: non_neg_integer(), nft_owners: non_neg_integer()}
   @typep template_id() :: AeMdw.Aex141.template_id()
+
+  @type statistic_tag() :: {:transactions, Node.tx_type() | :all}
+  @type interval() :: :day | :week | :month
```
jyeshe, line 45: `interval_by` would be more appropriate

jyeshe (commented):
Looking good, minor suggests and missed only sync tests to avoid regression.

lib/ae_mdw/stats.ex
sborrazas: Done

lib/ae_mdw/db/sync/stats.ex
sborrazas: Done

lib/ae_mdw/stats.ex
sborrazas: Yeah, was intending not to implement the filter yet. Also, monthly intervals would need to be rendered differently

jyeshe (approved):

PR #1516, opened 2023-08-09
feat: add week/month interval filter on statistics

sborrazas:


lib/ae_mdw/stats.ex
```diff
@@ -243,23 +251,44 @@ defmodule AeMdw.Stats do
...
     fn direction ->
       Collection.stream(state, Model.Statistic, direction, key_boundary, cursor)
     end
   end
 
+  defp render_statistic(state, {{:transactions, _tx_tag}, :month, interval_start} = statistic_key) do
+    Model.statistic(count: count) = State.fetch!(state, Model.Statistic, statistic_key)
+
+    %{
+      start_date: months_to_unix(interval_start),
```
jyeshe, line 276: how about using an iso date instead of the timestamp? It's compact, makes it already formatted to render on the frontend and is also human readable.

jyeshe (commented):
Works well, good idea to make the month also 0-indexed. Only for the format `start_date`, would pick the ISO.

test/ae_mdw_web/controllers/stats_controller_test.exs
```diff
@@ -165,4 +166,95 @@ defmodule AeMdwWeb.StatsControllerTest do
...
   end
+
+  test "when interval_by = week, it returns the count of transactions for the latest weekly periods",
+       %{
+         conn: conn,
+         store: store
+       } do
+    start_interval1 = 0
+    start_interval2 = 1_000
+    start_interval3 = 2_000
```
jyeshe, line 177: may we use multiples of 52 and add a test for a leap year?

lib/ae_mdw/stats.ex
sborrazas: this is how the node formats time already, so I'd thought I wouldn't change it to avoid having the client implement two separate decoders for datetimes

test/ae_mdw_web/controllers/stats_controller_test.exs
sborrazas: All three day/week/month implementations start counting for days/week/months since epoch, so leap years wouldn't really be an edge-case. I've added a testcase for 1972-1973 for which 1972 is a leap year anyway

lib/ae_mdw/stats.ex
jyeshe: With ISO date there is no need for decoders, it's yyyy-mm-dd. Even though the issue says `start_time`, the name chosen in the PR `start_date` makes it more precise since there is no time related to this stats.

test/ae_mdw_web/controllers/stats_controller_test.exs
jyeshe: With 52 or 53 weeks, it can give a different datetime (timestamp) or date depending on the implementation.
The idea of the test would be not to rely on the implementation and check easily without external computations (on how `1_000` should be represented as date).

jyeshe (approved):

PR #1519, opened 2023-08-10
refactor: remove name cache

jyeshe:
Cache not needed anymore after mem store refactoring.


sborrazas (approved):
Nice! I've been waiting to get rid of this, very unnecessary with the current implementation!

PR #1541, opened 2023-08-18
chore: unify convert_params usage into util function

sborrazas:


lib/ae_mdw/util.ex
```diff
@@ -169,4 +169,17 @@ defmodule AeMdw.Util do
         map
     end
   end
+
+  @spec convert_params(
+          map(),
+          ({binary(), binary()} -> {:ok, {atom(), term()}} | {:error, term()})
+        ) :: {:ok, Keyword.t()} | {:error, term()}
```
jyeshe, line 176: 5 cents on a reusable `convert_params`: instead of returning a `Keyword.t()` it could return a `map()` since each context module instantiates it.

lib/ae_mdw/contracts.ex
```diff
@@ -501,22 +504,23 @@ defmodule AeMdw.Contracts do
   end
 
   defp convert_param(state, {"contract_id", contract_id}),
-    do: {:create_txi, create_txi!(state, contract_id)}
+    do: {:ok, {:create_txi, create_txi!(state, contract_id)}}
```
jyeshe, line 507: Along with tagged `:ok | :error` would use a non-banging `create_txi/2`

lib/ae_mdw/contracts.ex
```diff
@@ -525,11 +529,9 @@ defmodule AeMdw.Contracts do
       |> Enum.flat_map(fn {tx_type, positions} -> Enum.map(positions, &{&1, tx_type}) end)
       |> Enum.group_by(fn {pos, _tx_type} -> pos end, fn {_pos, tx_type} -> tx_type end)
 
-    {:type_pos, {pos_types, Validate.id!(id_val)}}
+    {:ok, {:type_pos, {pos_types, Validate.id!(id_val)}}}
```
jyeshe, line 532: likewise for `Validate.id`

jyeshe (commented):
Nice refactoring, a small suggestion on `convert_params` output and I miss only some tests for each controller affected by newly incorporated parameter validation.  (e.g for `/oracles?state=some_mispelling`, `/contracts/logs?invalid_field=any_value`, etc.)

lib/ae_mdw/util.ex
sborrazas: Done

lib/ae_mdw/contracts.ex
sborrazas: Done

lib/ae_mdw/contracts.ex
sborrazas: Done

jyeshe (approved):

PR #1565, opened 2023-08-25
fix: match entity using nft collection and token id

jyeshe:


lib/ae_mdw/db/sync/active_entities.ex
```diff
@@ -30,29 +30,37 @@ defmodule AeMdw.Db.Sync.ActiveEntities do
...
       _nil_or_false ->
         state
     end
   end
 
-  defp exec_transition(state, create_txi, txi, {:initial, entity_name}) do
+  defp exec_transition(state, create_txi, txi, {:initial, entity_name}, args) do
     m_entity = Model.entity(index: {entity_name, txi, create_txi})
     m_contract_entity = Model.contract_entity(index: {entity_name, create_txi, txi})
+    final_args = args |> Enum.take(2) |> Enum.map(& &1.value)
```
jyeshe, line 43: Later, it will be possible to match args by positions (for now hardcodes that the first two must match)

lib/ae_mdw/db/model.ex
```diff
@@ -663,12 +663,24 @@ defmodule AeMdw.Db.Model do
...
+  #     index: {name, create txi, args}
+  #     txi: initial txi
+  @entity_args_defaults [index: {"", -1, []}, txi: -1]
+  defrecord :entity_args, @entity_args_defaults
+
+  @type entity_args_index() :: {String.t(), txi(), list()}
+  @type entity_args() ::
+          record(:entity_args,
+            index: entity_args_index()
+          )
```
sborrazas, line 676: It's missing `txi` field here. Also, should the list of arguments be part of the index? Can we move that to another record field?

jyeshe: Thanks! Regarding the list on the index, it is needed to be able to fetch the auction for a specific NFT but the list is comprised only by the collection contract id and token id. As we have other records with the index having two pubkeys it won't be a problem.

sborrazas (approved):

Issue #1572, opened 2023-09-01
Add ability to sort NFT's by creation time


janmichek:

jyeshe:
https://mainnet.aeternity.io/mdw/v2/aex141?by=creation&direction=forward

PR #1583, opened 2023-09-08
feat: sort aexn contracts by creation

jyeshe:
solves #1572

### DIFF
--- a/lib/ae_mdw/db/model.ex
+++ b/lib/ae_mdw/db/model.ex
@@ -529,17 +529,20 @@ defmodule AeMdw.Db.Model do
           record(:aexn_contract_symbol, index: aexn_contract_symbol_index())
 
   # AEX-N meta info sorted by txi_idx:
-  #     index: {type, {txi, idx}, pubkey}
-  #     unused: nil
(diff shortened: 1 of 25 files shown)


lib/ae_mdw/db/model.ex
```diff
@@ -528,6 +528,19 @@ defmodule AeMdw.Db.Model do
...
+  # AEX-N meta info sorted by txi_idx:
+  #     index: {type, {txi, idx}, pubkey}
+  #     unused: nil
+  @aexn_contract_creation_defaults [
+    index: {nil, {-1, -1}, nil},
+    unused: nil
+  ]
+  defrecord :aexn_contract_creation, @aexn_contract_creation_defaults
+
+  @type aexn_contract_creation_index() :: {aexn_type(), txi_idx(), pubkey()}
```
sborrazas, line 540: It it possible for the same event (local index) to create two different contracts? Otherwise we can move pubkey out of the index and reduce the index size

jyeshe: Done

sborrazas (approved):

PR #1587, opened 2023-09-13
feat: index and query aex9 transfers by contract and account

### LINKED ISSUE #1580: Get aex9 transfers filtered by contractId and accountId
Currently in the wallet we have an assets list that contain all the aex9 tokens user have balance on (`/v2/aex9/account-balances/${address}`). Not sure about all the ways across aeternity universe of getting such tokens on users account, but main source is of course DEX. And we are searching for the way to show transactions only related to a particular address and particular token, with as less requests as possible.

jyeshe:
closes #1580

PS: migration depends on logs records from resync/staging

### DIFF
--- a/lib/ae_mdw_web/router.ex
+++ b/lib/ae_mdw_web/router.ex
@@ -94,6 +94,14 @@ defmodule AeMdwWeb.Router do
       get "/aex9/:contract_id/balances/:account_id", AexnTokenController, :aex9_token_balance
       get "/aex9/account-balances/:account_id", AexnTokenController, :aex9_account_balances
 
+      get "/aex9/:contract_id/transfers-from/:sender",
+          AexnTransferController,
+          :aex9_contract_transfers
+
+      get "/aex9/:contract_id/transfers-to/:recipient",
+          AexnTransferController,
+          :aex9_contract_transfers
+
       get "/aex9/transfers/from/:sender", AexnTransferController, :aex9_transfers_from
       get "/aex9/transfers/to/:recipient", AexnTransferController, :aex9_transfers_to
 
--- a/lib/ae_mdw_web/router.ex
+++ b/lib/ae_mdw_web/router.ex
@@ -53,6 +53,10 @@ defmodule AeMdwWeb.Router do
       get "/names/auctions/:id/claims", NameController, :auction_claims
       get "/statistics/transactions", StatsController, :transactions_statistics
       get "/statistics/blocks", StatsController, :blocks_statistics
+
+      get "/aex9/:contract_id/transfers",
+          AexnTransferController,
+          :aex9_contract_transfers
     end
 
     scope "/v2" do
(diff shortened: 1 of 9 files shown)


lib/ae_mdw_web/router.ex
```diff
@@ -94,6 +94,14 @@ defmodule AeMdwWeb.Router do
...
       get "/aex9/account-balances/:account_id", AexnTokenController, :aex9_account_balances
 
+      get "/aex9/:contract_id/transfers-from/:sender",
+          AexnTransferController,
+          :aex9_contract_transfers
+
+      get "/aex9/:contract_id/transfers-to/:recipient",
+          AexnTransferController,
+          :aex9_contract_transfers
+
```
sborrazas, line 104: Can we always use query params to filter results that have the same structure and use paths only for nesting resources? e.g. `/aex9/:contract_id/transfers?sender=ak_...`

jyeshe: `v2` mostly uses path params. we should simplify the routes on `v3`

sborrazas: Ideally we should move all routes without requiring users to make any changes, so maybe we can start implementing routes as we would want them to be on v3 as well? Maybe even add this route on v3 exclusively

jyeshe: All token transfers are path dependent for accounts on `v2`. Using a new route pattern I am inclined to add only to `v3` after removing the `call_txi`. Would you have any sugestion of changes on the rendered object for `v3`? (Besides a pending fix on the aexn_type)

sborrazas: Looks good like this now, thanks 👍

sborrazas (approved):

lib/ae_mdw_web/router.ex
jyeshe: @janmichek thank you for asking. It's a route deployed on new endpoint version (v3). A new yaml will be prepared with it.
A request example for transfers from a wallet:
https://mainnet.aeternity.io/mdw/v3/aex9/ct_2BtHZSY9SfmiDQZ9yGqPppyRwKmrhHFa7yUJTpYMsWnJr23r5U/transfers?sender=ak_2UGL6qArnbWx9fEkJyszssMkCjpu3VeLFBPA6MXYV3APfDSz1k
or with recipient wallet:
https://mainnet.aeternity.io/mdw/v3/aex9/ct_2BtHZSY9SfmiDQZ9yGqPppyRwKmrhHFa7yUJTpYMsWnJr23r5U/transfers?recipient=ak_2wC9ev55yWy5uVtTUuUGsyKh5yXKAVDqVEKw1Tj8CziU3vfpgH

janmichek: Ahaaa, thanks for those examples. I thought it's going to work even without `sender` / `recipient` parameter.
For the feature (Dex trades) we are building for aescan we would need to see a different listing like this
https://etherscan.io/token/0xdac17f958d2ee523a2206206994597c13d831ec7#tokenTrade
I will request new feature then.

PR #1594, opened 2023-09-15
feat: filter and count microblock txs by type

### LINKED ISSUE #1471: Add ability to filter by tx type to /micro-blocks/{hash}/txs
The same as for [/txs/](https://petstore.swagger.io/?url=https://mainnet.aeternity.io/mdw/v2/api#/default/get_tx)

Also please add counter of such a count of filtered transactions

jyeshe:
resolves #1471

### DIFF
--- a/lib/ae_mdw/txs.ex
+++ b/lib/ae_mdw/txs.ex
@@ -159,6 +159,73 @@ defmodule AeMdw.Txs do
           add_spendtx_details?()
         ) :: {:ok, page_cursor(), [tx()], page_cursor()} | {:error, Error.t()}
   def fetch_txs(state, pagination, range, query, cursor, add_spendtx_details?) do
+    with {:ok, streamer} <- txs_streamer(state, range, query, cursor) do
+      do_fetch_txs(state, streamer, pagination, add_spendtx_details?)
+    end
+  end
+
+  @spec count_micro_block_txs(state(), binary(), query()) ::
+          {:ok, non_neg_integer()} | {:error, Error.t()}
+  def count_micro_block_txs(state, hash, query) do
+    with {:ok, streamer} <- micro_block_txs_streamer(state, hash, query) do
+      {:ok, Enum.count(streamer.(:forward))}
+    end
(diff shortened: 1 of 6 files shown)


lib/ae_mdw/txs.ex
```diff
@@ -159,40 +159,45 @@ defmodule AeMdw.Txs do
...
+      do_fetch_txs(state, streamer, pagination, add_spendtx_details?)
+    end
+  end
 
-      scope =
-        case range do
-          {:gen, first_gen..last_gen} ->
-            {DbUtil.first_gen_to_txi(state, first_gen), DbUtil.last_gen_to_txi(state, last_gen)}
+  @spec count_micro_block_txs(state(), binary()) :: {:ok, non_neg_integer()} | {:error, Error.t()}
+  def count_micro_block_txs(state, hash) do
```
sborrazas, line 168: Currently the `/micro-blocks/:hash` returns the count of transactions as part of the response, is there a reason we need a separate endpoint?

jyeshe: It will be used on the front-page, to avoid over-fetching until we don't have GraphQL (:

jyeshe: Replaced the streaming with tx indexes math though.

sborrazas: To avoid adding a new endpoint, perhaps we can use the already existing `/txs/count` endpoint, but include a new `block_hash=mh_...` filter, this way we can also start supporting `?type=spend&block_hash=mh_..` as per #1471 requirements

sborrazas (approved):
thanks!

lib/ae_mdw/txs.ex
jyeshe: The filtering would come next and it was avoiding mixing with other params as `/txs/count` accepts many but looks as a good idea. It could continue to be clear and an exception is launched when scope is used. 
PS: only mb_hash is accept since there might thousands of txs on a generation.

PR #1598, opened 2023-09-25
chore: move previous names into separate table

sborrazas:


lib/ae_mdw/sync/async_store_server.ex
```diff
@@ -49,6 +54,14 @@ defmodule AeMdw.Sync.AsyncStoreServer do
     {:noreply, state}
   end
 
+  def handle_cast({:write_mutations_without_async, mutations, done_fn}, state) do
+    _new_state = State.commit_db_without_async(State.new(), mutations)
```
jyeshe, line 58: Once the database transaction is an optimistic one this write operation might be lost/dropped on a conflict with the transaction from the sync server.

test/ae_mdw/db/names_expire_mutation_test.exs
```diff
@@ -11,25 +11,14 @@ defmodule AeMdw.Db.NamesExpirationMutationTest do
     test "inactivates a name that has just expired", %{store: store} do
       plain_name = "some-active-name"
       sync_height = 100_000
-
-      m_previous =
```
jyeshe, line 15: It would be nice to check that a `PreviousName` exists after the expiration.

test/ae_mdw_web/controllers/tx_controller_test.exs
```diff
@@ -1122,21 +1122,15 @@ defmodule AeMdwWeb.TxControllerTest do
           |> Store.put(Model.PlainName, Model.plain_name(index: name_hash2, value: plain_name2))
           |> Store.put(
             Model.ActiveName,
-            Model.name(
```
jyeshe, line 1125: It seems the controller test was incomplete. Can we have the `PreviousName` records on the fixture and check that it appears on the response?

lib/ae_mdw/names.ex
```diff
@@ -770,13 +761,14 @@ defmodule AeMdw.Names do
...
-    name
-    |> Stream.unfold(fn
-      Model.name(previous: nil) -> nil
-      Model.name(previous: previous) -> {previous, previous}
+  defp render_previous(state, plain_name, last_gen, last_micro_time, opts) do
+    key_boundary = {{plain_name, @min_int}, {plain_name, @max_int}}
+
+    state
+    |> Collection.stream(Model.PreviousName, :backward, key_boundary, nil)
+    |> Enum.map(fn Model.previous_name(name: name) ->
```
jyeshe, line 769: This needs a `fetch!` to have the `name` after the `index`.

jyeshe (changes requested):

lib/ae_mdw/sync/async_tasks/migrate.ex
```diff
@@ -15,7 +15,7 @@ defmodule AeMdw.Sync.AsyncTasks.Migrate do
     Enum.each(migrations, fn %Migration{mutations_mfa: {module, mutations_fn, params}} ->
       module
       |> apply(mutations_fn, params)
-      |> AsyncStoreServer.write_mutations(done_fn)
+      |> AsyncStoreServer.write_mutations_without_async(done_fn)
```
jyeshe, line 18: For a generic migration processing, this might cause errors on the sync server transaction because of the optimistic lock (it has happened in the past when the `AsyncStoreServer` used to persist data).

sborrazas: I've now removed the async migration changes. I thought these mutations would run before the server started, my bad

lib/ae_mdw/names.ex
sborrazas: Fixed, thanks

test/ae_mdw_web/controllers/tx_controller_test.exs
sborrazas: I've added previous names on the auctions test case which is where previous names appear

jyeshe (approved):
_()_

PR #1619, opened 2023-10-23
chore: unify pagination returns and cursor serialization

sborrazas:


thepiwo:
what will the cursor then look like?

sborrazas:
> what will the cursor then look like?

the cursor will look the same, this is more about unifying the way we code it since there was a lot of duplication on functions like `serialize_cursor/1`

jyeshe (approved):

Issue #1612, opened 2023-10-17
Names statistics


janmichek:
Please provide data for activated names (or claimed names) charts. Chart displays total number of names on time axis

janmichek:
EDIT: It should display names activated per time unit (day, week, month)

janmichek:
The data seems to be somehow incomplelte
EG: https://mainnet.aeternity.io/mdw/v3/statistics/names?limit=100&interval_by=month
Could you check, please?

sborrazas:
@janmichek this endpoint does not return intervals where count=0 and it looks like there's very few months where names where activated. Can you find a name that was activated outside of those months?

janmichek:
I also noticed this for the other statistics endpoint. 
As a user I would expect the time axis is continuous and it's not skipping days or months when the value at given time period is 0. Why just not print 0 for a time period with 0 events?

I see this chart as non-intuitive
![image](https://github.com/aeternity/ae_mdw/assets/15363559/c3ad8376-3d26-47b1-a5f3-42f25bdc9425)

janmichek:
And it's ultimately not true
![image](https://github.com/aeternity/ae_mdw/assets/15363559/38117db1-029b-499f-94f1-0fb057d5f2e7)

https://mainnet.aeternity.io/mdw/v3/statistics/names?interval_by=day&limit=7
as for 7 days limit I get span values from 02/02 to 02/17

sborrazas:
[...]

PR #1620, opened 2023-10-28
feat: add name activation statistics

sborrazas:
refs #1612


jyeshe (approved):

janmichek:
Thanks for the implementation. How long does it usually take to sync?
The status says both `synced` and `syncing` is true. Not sure what does it mean .
image

sborrazas:
@janmichek `mdw_syncing` means the syncing process is running, so for every new block added it will appear soon on the different endpoints. `mdw_synced` means that mdw is up to date with the latest block

janmichek:
Thanks for the hint @sborrazas . Let me put it this way - how long do you expect it to be syncing till we get complete data from the endpoint

PR #1646, opened 2023-12-13
fix: handle micro-block cursor properly

sborrazas:


jyeshe (commented):
May it be also `nil`?

sborrazas:
> May it be also `nil`?

The Collection.paginate cursor_fn is never called with nil values: https://github.com/aeternity/ae_mdw/blob/master/lib/ae_mdw/collection.ex#L40

jyeshe (approved):

PR #1648, opened 2023-12-14
fix: check return_type instead of ret_value for errors

sborrazas:


jyeshe (changes requested):
Just seen the dialyzer saying smth

jyeshe (approved):

PR #1674, opened 2024-02-01
fix: include all auction claims when closing up an auction

sborrazas:
Requires re-sync

### DIFF
--- a/lib/ae_mdw/db/mutations/name_claim_mutation.ex
+++ b/lib/ae_mdw/db/mutations/name_claim_mutation.ex
@@ -120,32 +120,48 @@ defmodule AeMdw.Db.NameClaimMutation do
         auction_end = height + timeout
         m_auction_exp = Model.expiration(index: {auction_end, plain_name})
 
-        m_auction_bid =
-          Model.auction_bid(
-            index: plain_name,
-            block_index_txi_idx: {block_index, txi_idx},
-            expire_height: auction_end,
-            owner: owner_pk
-          )
-
-        auction_claim = Model.auction_bid_claim(index: {plain_name, auction_end, txi_idx})
-
         state3 =
           IntTransfer.fee(state2, {height, txi_idx}, :spend_name, owner_pk, txi_idx, name_fee)
 
         state4 =
           case State.get(state3, Model.AuctionBid, plain_name) do
             :not_found ->
-              State.inc_stat(state3, :auctions_started)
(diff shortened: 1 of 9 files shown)


lib/ae_mdw/db/mutations/name_claim_mutation.ex
```diff
@@ -120,33 +120,49 @@ defmodule AeMdw.Db.NameClaimMutation do
...
+              auction_claim = Model.auction_bid_claim(index: {plain_name, start_height, txi_idx})
+
+              m_auction_bid =
+                Model.auction_bid(
+                  index: plain_name,
+                  start_height: start_height,
+                  block_index_txi_idx: {block_index, txi_idx},
+                  expire_height: auction_end,
+                  owner: owner_pk
+                )
```
jyeshe, line 164: Not an operational issue but as the purpose here is an update, would use the original record and set only fields that needs to be updated to make it more clear what it is changing.

jyeshe (changes requested):
There are some points where it references `expired_height` instead of `started_height` for an `AuctionBidClaim`:
https://github.com/aeternity/ae_mdw/blob/aaae71a61301210ae7fcb3f16cab43b5ab03fcbe/lib/ae_mdw/auction_bids.ex#L154
https://github.com/aeternity/ae_mdw/blob/aaae71a61301210ae7fcb3f16cab43b5ab03fcbe/lib/ae_mdw/names.ex#L763-L768

We would be great to have the test suite covering this gap.

sborrazas:
> There are some points where it references `expired_height` instead of `started_height` for an `AuctionBidClaim`:
> 
> https://github.com/aeternity/ae_mdw/blob/aaae71a61301210ae7fcb3f16cab43b5ab03fcbe/lib/ae_mdw/auction_bids.ex#L154
> 
> 
> https://github.com/aeternity/ae_mdw/blob/aaae71a61301210ae7fcb3f16cab43b5ab03fcbe/lib/ae_mdw/names.ex#L763-L768
> 
> We would be great to have the test suite covering this gap.

Good find, fixed the /auctions endpoint and updated tests accordingly which started to fail. thanks

lib/ae_mdw/db/mutations/name_claim_mutation.ex
sborrazas: Done

jyeshe (approved):

Issue #1666, opened 2024-01-11
v2/names/<name.chain> shows only the last bid tx idx


davidyuk:
```json
// 20240111232129
// https://testnet.aeternity.io/mdw/v2/names/testtest.chain

{
  "active": false,
  "info": {
    "approximate_auction_end_time": 1707668435657,
    "auction_end": 904987,
    "bids": [
      38159222
    ],
    "last_bid": {
      "block_hash": "mh_Qve2gRJJarURCxZenxDPcHWvzKkZoBNNCRUpHa459RcDnKd9V",
      "block_height": 890107,
      "hash": "th_21gApXt3t8LYoRcjrvT7QviPao7ZKbuz1wiKFnC9V6AbwKKQvZ",
      "micro_index": 2,
      "micro_time": 1704990062622,
      "signatures": [
        "sg_9EXeRtmiNMRLoqnDXVju2gE3hMH6FzRJN3Vd8REg96Zx1Fjwdxw5JS3YmUcDPMVZcaUj6NJNFJsSBq6p1j51FSTNTNZ1D"
      ],
      "tx": {
        "account_id": "ak_2swhLkgBPeeADxVTAVCJnZLY5NZtCFiM93JxsEaMuC59euuFRQ",
        "fee": 16500000000000,
        "name": "testtest.chain",
        "name_fee": 23000000000000000000,
        "name_id": "nm_2M4LHM6Two4ELNSErs5aAnJNQeKUBesiQWNAxq7wckLPybv5eb",
        "name_salt": 0,
        "nonce": 1661,
        "ttl": 1084987,
        "type": "NameClaimTx",
        "version": 2
      },
      "tx_index": 38159222
    }
  },
  "name": "testtest.chain",
  "previous": [
    
  ],
  "status": "auction"
}
```
The `bids` array contains only one value, but there are three bids: the initial claim and two bids.
https://testnet.aescan.io/transactions/th_21gApXt3t8LYoRcjrvT7QviPao7ZKbuz1wiKFnC9V6AbwKKQvZ -- 23ae bid
https://testnet.aescan.io/transactions/th_fK44V3FozgrkxaARSBfscKvJZoQ7c4cR6RinX2c6CZJ11YLDg -- 21ae bid
https://testnet.aescan.io/transactions/th_28b8mZF1FBFXTJCCDrqW8L6G3hdoAuYmGuDPWndvpW5KHJPYRH -- initial claim

[...]

PR #1677, opened 2024-02-07
feat: add v3 name and auction detail endpoint

sborrazas:
refs #1666


jyeshe (approved):
Looks good for the purpose of not returning bids and claims endpoint for names in auction in v3. Just one testing improvement.

jyeshe:
The migration that is blocking the test I can fix it tomorrow.

PR #1691, opened 2024-02-27
fix: process HC seed contracts with the correct format

sborrazas:


lib/ae_mdw/db/origin.ex
```diff
@@ -143,21 +143,12 @@ defmodule AeMdw.Db.Origin do
...
-      |> Enum.map(&(Map.get(&1, "pubkey") || Map.fetch!(&1, "contract_pubkey")))
-      |> Enum.map(&Validate.id!/1)
-    else
-      _error_missing -> []
-    end
+    {:ok, hc_contracts_json} =
+      :aec_fork_block_settings.hc_seed_contracts(@iris_protocol, :aec_governance.get_network_id())
+
+    hc_contracts_json
+    |> Map.get("contracts", [])
```
jyeshe, line 150: During development json that defines the hc seed was also defined as a list. This is the reason for the existing code that handles both maps and keywords with:
```
hc_contracts_json
|> Enum.find_value(fn {item, list} -> if item == "contracts", do: list end)
```
In other words it covers what `|> Map.get("contracts", [])` does and handles keywords as well.

lib/ae_mdw/db/origin.ex
```diff
@@ -143,21 +143,12 @@ defmodule AeMdw.Db.Origin do
...
   defp hc_contracts do
-    with {:ok, {hardfork_key, _hf_val}} <-
-           :aeu_env.find_config(["chain", "hard_forks"], [
-             :user_config,
-             :schema_default,
-             {:value, {@iris_protocol, 0}}
-           ]),
-         {protocol, ""} <- Integer.parse(hardfork_key),
-         {:ok, hc_contracts_json} <-
-           :aec_fork_block_settings.hc_seed_contracts(protocol, :aec_governance.get_network_id()) do
```
jyeshe, line 154: During localnet tests there was a mapping from "5" to a custom protocol id like:
```
    "hard_forks": {
      "5": 0
    }
```
If the Node team won't use the mapping anymore it makes sense to pass 5 directly instead of looking to which protocol the hard fork is associated.

jyeshe (commented):
Could you provide a description of the issue (including input and the log)?

The processing of HC seed contracts (`hc_contracts_json`) here does not change the format except that it's more strict, handling less variations. This way it wouldn't fix the format but instead would not accept lists that were used during development.

lib/ae_mdw/db/origin.ex
sborrazas: It still uses that same format, but didn't return a tuple , it instead returned a map:
```elixir
iris_protocol = "5"
:aeu_env.find_config(["chain", "hard_forks"], [
  :user_config,
  :schema_default,
  {:value, {iris_protocol, 0}}
])
{:ok, %{"5" => 0}}
```
which ended up matching with `_error_missing` and not returning the contracts

lib/ae_mdw/db/origin.ex
sborrazas: `hc_seed_contracts(protocol, network_id)` is always going to return the output of JSON decoding: https://github.com/aeternity/aeternity/blob/v6.12.0/apps/aecore/src/aec_fork_block_settings.erl#L358 which is never going to return keyword lists, since JSON maps are returned as erlang maps

sborrazas (commented):
This error came up because there were missing contracts on the HC test server when calling `Origin.tx_index(...)`

lib/ae_mdw/db/origin.ex
jyeshe: `Enum.find_value` still handles the decoding of any JSON which might be a map or a list and does not depend on the implementation.

lib/ae_mdw/db/origin.ex
jyeshe: The only remaining question would be if we want to remove the mapping and use always the iris protocol (no more hardkfork mapping for dev).

jyeshe (changes requested):
The processing of HC seed contracts is currently more robust (handling both lists and maps). 
In order to make it clear the purpose of the change the name of the PR should indicate that we are fixing the hardfork mapping or actually getting hc contracts directly for a protocol. Not for a hardfork config anymore, the processing of HC seed contracts would be unchanged.

PR #1697, opened 2024-03-11
fix: allow same creation block to be used on by-hash aex9 balances

sborrazas:


sborrazas:
@thepiwo issue was aex9 contract balances where you can specify a block hash to get the balances of that specific block. Endpoint worked fine, but there was a validation where you were only able to specify a block created AFTER the contract was created, but couldn't use the SAME block. This change allows you to get the balances from the block in which it was first created.

E.g. the contract balances for the block in which it was created returns 404
https://testnet.aeternity.io/mdw/aex9/balances/hash/mh_2gTPvRatcckf7vr4DYAKW8az2eqKiDoiojdY11HEUqYcYzs8V7/ct_2ZjEWgr4BaLrqEHwRvgWwp4E834xnsknJ1HxSnCkdsHH2Mm5No
But returns 200 for the block right after that one:
https://testnet.aeternity.io/mdw/aex9/balances/hash/mh_cVa23fThaJctihX1NCqcuoBSdMcFjEaU7YvDVy9FQAfpycifo/ct_2ZjEWgr4BaLrqEHwRvgWwp4E834xnsknJ1HxSnCkdsHH2Mm5No

This is going to be improved anyway (pagination added) on a V3 endpoint

PR #1701, opened 2024-03-15
feat: allow getting block-specific AEx9 balances

sborrazas:
This requirements is needed by @kenodressel so that it can use v3 instead of the old v1 endpoint


thepiwo (commented):
should we also add this to the readme?

sborrazas:
@thepiwo good call, added to README as well

thepiwo (approved):

PR #1708, opened 2024-03-21
feat: add raw data pointers support for ceres

vatanasov:
This PR adds support for the new raw data pointers format, added in the Ceres protocol


sborrazas (approved):
LGTM

PR #1710, opened 2024-03-22
feat: resolve aens name to contract address when calling contract

yaboiishere:
Depends on #1708 since it's branched out from it

### DIFF
--- a/lib/ae_mdw/contract.ex
+++ b/lib/ae_mdw/contract.ex
@@ -57,6 +57,8 @@ defmodule AeMdw.Contract do
   @type fun_arg_res_or_error :: fun_arg_res() | {:error, any()}
   @type local_idx :: non_neg_integer()
   @type code() :: binary()
+  @type name_pubkey :: pubkey()
+  @type contract_pubkey :: pubkey()
   @opaque aecontract() :: tuple()
   @typep pubkey :: DBN.pubkey()
   @typep tx :: Node.tx()
@@ -346,6 +348,7 @@ defmodule AeMdw.Contract do
     Enum.group_by(get_events(micro_block), fn {_event_name, %{tx_hash: tx_hash}} -> tx_hash end)
   end
 
+  @spec maybe_resolve_contract_pk(name_pubkey() | contract_pubkey()) :: contract_pubkey()
   def maybe_resolve_contract_pk(contract_or_name_pk) do
     contract_or_name_pk
     |> case do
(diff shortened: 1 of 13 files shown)


lib/ae_mdw/contract.ex
```diff
@@ -346,6 +348,26 @@ defmodule AeMdw.Contract do
...
+    contract_or_name_pk
+    |> case do
+      <<217, 52, 92, _rest::binary>> = name_pk ->
+        "contract_pubkey"
+        |> :aec_chain.resolve_namehash(name_pk)
+        |> case do
+          {:ok, {:id, :contract, pubkey}} ->
+            pubkey
+
+          {:error, _} ->
```
sborrazas, line 362: Not sure why credo didn't catch this but unused variables should contain a reference to what it actually is (`_reason` in this case). But maybe it would be even better to include the reason as part of the error being raised

lib/ae_mdw/contract.ex
```diff
@@ -346,6 +348,26 @@ defmodule AeMdw.Contract do
...
 
+  @spec maybe_resolve_contract_pk(name_pubkey() | contract_pubkey()) :: contract_pubkey()
+  def maybe_resolve_contract_pk(contract_or_name_pk) do
+    contract_or_name_pk
+    |> case do
+      <<217, 52, 92, _rest::binary>> = name_pk ->
+        "contract_pubkey"
+        |> :aec_chain.resolve_namehash(name_pk)
+        |> case do
+          {:ok, {:id, :contract, pubkey}} ->
```
sborrazas, line 359: Can we avoid exposing the aeser_id structure and instead use the getters to retrieve the value? Something like
```elixir
{:ok, contract_id} -> :aeser_id.specialize(contract_id, :contract)
```

lib/ae_mdw/db/name.ex
```diff
@@ -55,12 +55,21 @@ defmodule AeMdw.Db.Name do
...
+  @spec ptr_resolve(state(), Blocks.block_index(), binary(), binary()) ::
           {:ok, binary()} | {:error, :name_revoked}
-  def ptr_resolve(state, block_index, name_hash) do
-    with {:ok, account_id} <-
-           :aens.resolve_hash("account_pubkey", name_hash, ns_tree!(state, block_index)) do
-      {:ok, Validate.id(account_id)}
+  def ptr_resolve(state, block_index, name_hash, pointer_key) do
+    with {:ok, id} <-
+           :aens.resolve_hash(pointer_key, name_hash, ns_tree!(state, block_index)) do
+      Validate.id(id)
```
sborrazas, line 63: `Validate.id/1` might return `{:error, Error.Id}`, which isn't what ptr_resolve returns here. Since we know it's a valid ID we don't need to validate, and we can use `:aeser_id.specialize` instead, right?

sborrazas (approved):

Issue #1635, opened 2023-11-27
Always getting some metrics error


dincho:
I'm always getting some metrics related error when running the mdw docker container:

```
2023-11-27 13:35:09.592 [error] Error when calling MFA defined by measurement: AeMdw.Apm.CustomMetrics :dispatch_status []
Class=:exit
Reason={:noproc, {GenServer, :call, [AeMdw.Sync.MemStoreCreator, :create, 5000]}}
Stacktrace=[
  {GenServer, :call, 3, [file: 'lib/gen_server.ex', line: 1019]},
  {AeMdw.Db.State, :create_mem_state, 0,
   [file: 'lib/ae_mdw/db/state.ex', line: 122]},
  {AeMdw.Db.State, :mem_state, 0, [file: 'lib/ae_mdw/db/state.ex', line: 115]},
  {AeMdw.Apm.CustomMetrics, :dispatch_status, 0,
   [file: 'lib/ae_mdw/apm/custom_metrics.ex', line: 9]},
  {:telemetry_poller, :make_measurement, 1,
   [
     file: '/home/aeternity/node/ae_mdw/deps/telemetry_poller/src/telemetry_poller.erl',
     line: 322
   ]},
  {:telemetry_poller, :"-make_measurements_and_filter_misbehaving/1-lc$^0/1-0-",
   1,
   [
     file: '/home/aeternity/node/ae_mdw/deps/telemetry_poller/src/telemetry_poller.erl',
     line: 318
   ]},
  {:telemetry_poller, :handle_info, 2,
   [
     file: '/home/aeternity/node/ae_mdw/deps/telemetry_poller/src/telemetry_poller.erl',
     line: 340
   ]},
  {:gen_server, :try_dispatch, 4, [file: 'gen_server.erl', line: 689]}
]
```

Metrics should be disabled by default, or if this is not exactly error, it's log level should be decreased

PR #1716, opened 2024-03-27
fix: telemetry error when application starts

vatanasov:
refs: #1635.

`AeMdw.APM.Telemetry` was issuing calls to `AeMdw.Sync.MemStoreCreator`, that is started at a later point (by `AeMdw.Sync.Supervisor`), so I switched the starting order, to prevent the error.


sborrazas (approved):
Nice find!

PR #1719, opened 2024-04-01
fix: tests using nonexisting column

vatanasov:


sborrazas (commented):
Can we rename this commits/PR to be `test: ...`, we use `fix` for bug fixes that users can benefit from

vatanasov:
> Can we rename this commits/PR to be `test: ...`, we use `fix` for bug fixes that users can benefit from

Sure, I will just create a new branch/PR, so the branch will also have the appropriate name.

vatanasov:
Closed in favor of https://github.com/aeternity/ae_mdw/pull/1720

Issue #1018, opened 2022-11-15
Replace `expand=true` 


janmichek:
We are using `expand=true`
On 3 places

1. https://github.com/aeternity/explorer-frontend/blob/7caf97828de582567aaff6a4000254ff4d2b0623/src/stores/names.js#L51
to display fees
https://master-explorer-frontend.prd.aepps.com/names?type=active

2. https://github.com/aeternity/explorer-frontend/blob/7caf97828de582567aaff6a4000254ff4d2b0623/src/stores/names.js#L65
to display fees
https://master-explorer-frontend.prd.aepps.com/names?type=expired

3. https://github.com/aeternity/explorer-frontend/blob/7caf97828de582567aaff6a4000254ff4d2b0623/src/stores/names.js#L70
to display fees
https://master-explorer-frontend.prd.aepps.com/
(section Names Recently activated)

thepiwo:
Can you elaborate to replace with what?

janmichek:
I don't have a strong preference.
The info I got, is that this will be deprecated and I should write down places in Explorer where we are using it.

janmichek:
To get those fees values from a response even without `expand=true` ?

thepiwo:
@sborrazas @jyeshe is there significant overhead in adding those in terms of calculation effort?

marc0olo:
is it only about the "buy price" then, right?

sborrazas:
The `expand=true` is being used on these three places:
- On the names nested list of claims/updates/transfers and on the revoke txi. When true, it turns the txis into a transaction object. Having these nested resources on separate endpoints (except on the revoke which is just one) we can always include the tx info without the need for `expand=true`.
- Same thing as before, but with nested oracle list of extends. It turns txis into txs.
- Same as before but with nested claims inside auctions. It turns txis into txs.

For the fees and last buyer on inactive names we can also always include the last claim, instead of the entire list of claims.

In addition, we could simply assume that `expand` is always `true` and make the necessary changes. I don't think we need to wait to get rid of unlimited nested lists to make this change.

thepiwo:
but its a breaking change to change default behaviour, so why is it a problem to currently use `expand=true`?

sborrazas:
@thepiwo yes, it would indeed be a breaking change. The advice on not using `expand=true` is because on the next versions most of the data obtained there will be removed

sborrazas:
`expand` param is no longer needed, it always acts as if it was `true`

PR #1725, opened 2024-04-05
fix: update names and oracles to v3

yaboiishere:
This addresses #1018

### DIFF
--- a/docs/swagger_v3/names.spec.yaml
+++ b/docs/swagger_v3/names.spec.yaml
@@ -1,4 +1,30 @@
 schemas:
+  NameTx:
+  properties:
+    block_hash: string
+    block_height: integer
+    hash: string
+    micro_index: integer
+    micro_time: integer
+    signatures:
+      type: array
+      items:
+        type: string
+      tx:
+        type: object
+        properties:
+          fee: integer
+          nonce: integer
+          oracle_id: string
+          oracle_ttl:
+            type: object
+            properties:
+              type: string
+              value: integer
+          ttl: integer
+          type: string
+          version: integer
+          tx_hash: string
   Name:
     description: Name
     type: object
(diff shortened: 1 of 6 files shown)


docs/swagger_v3/names.spec.yaml
```diff
@@ -1,4 +1,30 @@
 schemas:
+  NameTx:
+  properties:
```
sborrazas, line 3: This doesn't appear to be nested inside NameTx. Also, what is a NameTx? is it claim, update, revoke?

docs/swagger_v3/oracles.spec.yaml
```diff
@@ -1,17 +1,73 @@
 schemas:
+  OracleTx:
```
sborrazas, line 2: Same here, what's an OracleTx? If it's an extends tx, shouldn't it be OracleExtendsTx ?

docs/swagger_v3/oracles.spec.yaml
```diff
@@ -1,17 +1,73 @@
...
+      block_hash: string
+      block_height: integer
+      hash: string
+      micro_index: integer
+      micro_time: integer
+      signatures:
+        type: array
+        items:
+          type: string
+        tx:
```
sborrazas, line 13: `tx` here should not be part of signatures

lib/ae_mdw_web/controllers/oracle_controller.ex
```diff
@@ -15,6 +15,15 @@ defmodule AeMdwWeb.OracleController do
 
   @spec oracle(Conn.t(), map()) :: Conn.t()
   def oracle(%Conn{assigns: %{state: state, opts: opts}} = conn, %{"id" => id}) do
+    with opts <- [{:v3, true} | opts],
```
sborrazas, line 18: Why use `opts` as part of the `with` statement instead of a regular variable binding?

lib/ae_mdw_web/router.ex
```diff
@@ -57,6 +57,8 @@ defmodule AeMdwWeb.Router do
       get "/micro-blocks/:hash/transactions", TxController, :micro_block_txs
 
       get "/transactions", TxController, :txs
+      get "/transactions/count", TxController, :count
+      get "/transactions/count/:id", TxController, :count_id
```
sborrazas, line 61: This didn't change, we should leave it unchanged

sborrazas (changes requested):

docs/swagger_v3/names.spec.yaml
yaboiishere: This is so i can resue the structure inside of the latter schemas

lib/ae_mdw_web/controllers/oracle_controller.ex
yaboiishere: A habbit, on an older project we used to put the inside the with and if it's going to always match we used `=` instead of  `<-`. I don't have a strong preference i can change it back

lib/ae_mdw_web/router.ex
yaboiishere: I moved them because `/transactions/:hash` was above them and when i tried calling this it was trying to find a transaction with hash `count`

sborrazas: This is already fixed here https://github.com/aeternity/ae_mdw/pull/1712, not yet merged, though

docs/swagger_v3/oracles.spec.yaml
yaboiishere: I'm using it only for the shape of the object, so it's not written everytime it's being used

sborrazas (approved):

PR #1727, opened 2024-04-11
fix: use tx hash instead of index in v3 api version

yaboiishere:
ref: #764


lib/ae_mdw/activities.ex
```diff
@@ -664,7 +664,11 @@ defmodule AeMdw.Activities do
   @spec render_payload(state(), Db.pubkey(), height(), txi(), activity_value()) ::
           {activity_type(), map()}
   defp render_payload(state, _account_pk, _height, txi, {:field, tx_type, _tx_pos}) do
-    tx = state |> Txs.fetch!(txi) |> Map.delete("tx_index")
+    tx =
+      state
+      |> Txs.fetch!(txi)
+      |> Map.put("tx_hash", Enc.encode(:tx_hash, Txs.txi_to_hash(state, txi)))
+      |> Map.delete("tx_index")
```
sborrazas, line 671: Here we're fetching the transaction twice (on fetch! and on `txi_to_hash`). Plus, I think hash is included already on the txs render function

sborrazas (changes requested):

lib/ae_mdw/activities.ex
yaboiishere: Yeah, you're correct, I've removed this. Also had to remove it from the tests since the function it was comming from was mocked to always return %{},  so i decided to return the test as it was

sborrazas (approved):

PR #1732, opened 2024-04-18
chore: update node to version 6.13

vatanasov:
The node now requires OTP 24, so the versions of elixir and erlang were also bumped (elixir: 1.13 -> 1.16, erlang: 23 -> 26).

Most of the changes in this PR are related to the fact, that tests are now sometimes executing in different order or different speed (probably because of optimizations) and also changes in how maps are ordered (which is now not-deterministic).


lib/ae_mdw/aex9.ex
```diff
@@ -40,13 +39,14 @@ defmodule AeMdw.Aex9 do
 
   @type amounts :: map()
 
-  @spec fetch_balances(State.t(), pubkey(), boolean()) :: {:ok, amounts()} | {:error, Error.t()}
-  def fetch_balances(state, contract_pk, top?) do
+  @spec fetch_balances(State.t(), State.t(), pubkey(), boolean()) ::
+          {:ok, amounts()} | {:error, Error.t()}
+  def fetch_balances(state, async_state, contract_pk, top?) do
```
vatanasov, line 44: This change, and all that follows, where `async_state` is explicitly passed (and stored, in the case of `AsyncStoreServer`) are made so that the state can be reliably mocked and not relied on the global one.

mix.exs
```diff
@@ -187,10 +187,7 @@ defmodule AeMdw.MixProject do
       plt_add_apps: [:mix],
       plt_core_path: "priv/plts",
       plt_file: {:no_warn, "priv/plts/dialyzer.plt"},
-      flags: [
-        :unmatched_returns,
-        :race_conditions
-      ]
+      flags: [:unmatched_returns]
```
vatanasov, line 190: The removed flag was removed from the library itself: https://github.com/jeremyjh/dialyxir/pull/471

test/support/blockchain_sim.ex
```diff
@@ -152,7 +152,7 @@ defmodule AeMdwWeb.BlockchainSim do
 
     {_prev_hash, mock_blocks, mock_txs, _max_height} =
       blocks
-      |> Enum.reduce({@genesis_prev_hash, %{}, %{}, @initial_height}, fn
+      |> Enum.reduce({@genesis_prev_hash, %{}, [], @initial_height}, fn
```
vatanasov, line 155: Switching them to `Keyword` instead of `Map` makes the order deterministic (some tests rely on that)

lib/ae_mdw/db/async_store.ex
```diff
@@ -19,17 +19,17 @@ defmodule AeMdw.Db.AsyncStore do
   @typep key() :: Database.key()
   @typep record() :: Database.record()
   @typep table() :: Database.table()
-  @opaque t() :: %__MODULE__{tid: :single_async_store}
+  @type t() :: %__MODULE__{tid: :single_async_store}
```
sborrazas, line 22: Why do we need to expose the internal structure? I'd rather not

lib/ae_mdw/db/int_transfer.ex
```diff
@@ -40,7 +40,12 @@ defmodule AeMdw.Db.IntTransfer do
   def block_rewards_mutations(key_block) do
     height = :aec_blocks.height(key_block)
     delay = :aec_governance.beneficiary_reward_delay()
-    dev_benefs = Enum.map(:aec_dev_reward.beneficiaries(), &elem(&1, 0))
+
+    dev_benefs =
+      for {protocol, _height} <- :aec_hard_forks.protocols(),
+          {pk, _} <- :aec_dev_reward.beneficiaries(protocol) do
```
sborrazas, line 46: Not sure why credo didn't catch it, but can we name unused variables anyway?

sborrazas (changes requested):

lib/ae_mdw/db/async_store.ex
vatanasov: Otherwise dialyzer didn't allow using pattern match on the struct itself (not only on the properties, probably because the struct is in a property) like in here: https://github.com/aeternity/ae_mdw/pull/1732/files#diff-50003680f6579cccc558e6fd7799dad10972701a70a1a307a18f977e94634e07R32 and several other places. I prefer having explicit matches, because this will make it fail sooner and more obvious, if something else gets passed, but I can change it.

sborrazas: I'd rather keep structures encapsulated in their own modules than exposing them for runtime checks. We try to avoid runtime errors and prioritize encapsulation and dialyzer (static) checks. If anyone tries sending anything other than an AsyncStore is going to get caught even before executing the app

sborrazas (approved):

PR #1735, opened 2024-04-22
test: fix broadcast randomly failing test

vatanasov:


sborrazas:
Can there be an explanation of why the tests fail using the registered global name? I would try to avoid making the source code more complex simply for tests to pass

yaboiishere:
I believe the problem was that the mocking library was having problems unloading the mocks on shared processes and if two tests mock their functionality, sometimes the wrong mocks are used and the tests fail

sborrazas:
Tests should run synchronously for correct mocking and unmocking of the modules, if this is not the case we should figure out a different way to test this module/server (which is working correctly) without having to change it. If it's not possible let's get rid of these tests

vatanasov:
There were different problems that occurred. The problem with the mocking was, that for some runs, where were no `mock_blocks[:mb]`, so `find_tx_location` was not working correctly (I have no idea why there were runs where there were - the tests themselves does not have that key. My suspicion is that maybe something running in another process is sometimes running another version of the mocked function, that has an `:mb` key).

As for the registered global name - the problem is, that the Broadcaster using the registered global name is internally using `State.mem_state`, which is global and is changed in some of the other tests. The point to have it using a name is so that a broadcaster has his own internal state instead of always using the global one (and in prod - using the global one). I think that making the name and the state that the genservers are using configurable will make them much more easy to test (and more flexible to modify, in case we need more than one of them), without interfering with other tests, and only using tools already in the language. 

I get that changing such an integral part of the code, such as the Broadcaster is worrying, so if you are wary, we could just leave the test commented for now (but I would still suggest that we try to make future genservers more flexible with regards to global states).

venimus:
fixed in 1.107.0

Issue #1629, opened 2023-11-22
Add counter for account (active) names


janmichek:
![image](https://github.com/aeternity/ae_mdw/assets/15363559/3b2619a9-bf27-4d18-b487-d8c64e375bea)

PR #1737, opened 2024-04-23
feat: add endpoint to get names count

yaboiishere:
ref: #1629


lib/ae_mdw/names.ex
```diff
@@ -70,6 +70,16 @@ defmodule AeMdw.Names do
   @min_bin Util.min_bin()
   @max_bin Util.max_256bit_bin()
 
+  @spec count_names(state(), query()) :: non_neg_integer()
+  def count_names(state, query) do
+    with {:ok, filters} <- Util.convert_params(query, &convert_param/1) do
+      filters
+      |> build_height_streamer(state, nil, nil, nil)
+      |> then(fn streamer -> streamer.(:backward) end)
```
sborrazas, line 78: Is this going to go through all of the names? There's no limit to how many names there could be, this could take a lot of time

sborrazas (changes requested):
There's no need to add a new endpoint, this is already exposed via totalstats endpoint here https://testnet.aeternity.io/mdw/v3/totalstats

lib/ae_mdw/names.ex
yaboiishere: It is a slow query yeah, but the idea is so that this can be used to get all active names owned by an account. The problem with counting the results from the regular endpoint is that it's paginated and can't be counted

yaboiishere:
> There's no need to add a new endpoint, this is already exposed via totalstats endpoint here https://testnet.aeternity.io/mdw/v3/totalstats

Yeah, but it can't be filtered on based on the owner

sborrazas:
@yaboiishere we cannot add endpoints that expose features that can be easily exploited. We also shouldn't be adding any pagination counters, that's database dependent and our current database doesn't allow it.

yaboiishere:
So i should reimplement this in some other way. Maybe it can something like the account balances where a Genserver keeps track of those and it gets updated on every relevant mutation?

sborrazas:
> So i should reimplement this in some other way. Maybe it can something like the account balances where a Genserver keeps track of those and it gets updated on every relevant mutation?

The proper way of implementing this would be to add a new index that keeps track of all of the counters for all accounts, which will increase the size of the database. And if we do that for every paginated endpoint we will end up with a database double or triple the size, so we should avoid implementing these features

lib/ae_mdw/db/util.ex
```diff
@@ -358,4 +358,39 @@ defmodule AeMdw.Db.Util do
...
+
+  @spec decrement_names_count(state(), Db.pubkey()) :: State.t()
+  def decrement_names_count(state, owner_pk) do
+    State.update(
+      state,
+      Model.AccountNamesCount,
+      owner_pk,
+      fn
+        nil ->
+          Model.account_names_count(index: owner_pk, count: 0)
```
sborrazas, line 386: We can use this value as the default and remove this clause

lib/ae_mdw/names.ex
```diff
@@ -72,6 +72,22 @@ defmodule AeMdw.Names do
...
+    query
+    |> Util.convert_params(&convert_param/1)
+    |> case do
+      {:ok, %{owned_by: owned_by}} ->
+        {:ok, count_active_names(state, owned_by)}
+
+      {:ok, %{}} ->
+        {:ok, count_active_names(state, nil)}
+
+      {:error, _} ->
```
sborrazas, line 86: `{:error, _reason}`. Also, can't we just return this same error and use a with expresion?

lib/ae_mdw/names.ex
```diff
@@ -72,6 +72,22 @@ defmodule AeMdw.Names do
...
+  @spec count_names(state(), query()) :: {:ok, non_neg_integer()} | {:error, reason()}
+  def count_names(state, query) do
+    query
+    |> Util.convert_params(&convert_param/1)
+    |> case do
+      {:ok, %{owned_by: owned_by}} ->
+        {:ok, count_active_names(state, owned_by)}
+
+      {:ok, %{}} ->
+        {:ok, count_active_names(state, nil)}
```
sborrazas, line 84: We can remove these two clauses by doing `{:ok, filters} -> count_active_names(state, Map.get(filters, :owned_by))`

lib/ae_mdw/names.ex
```diff
@@ -1018,4 +1034,21 @@ defmodule AeMdw.Names do
       tx: :aens_update_tx.for_client(name_update_tx)
     }
   end
+
+  defp count_active_names(state, nil) do
+    state
+    |> Collection.stream(Model.AccountNamesCount, :forward, nil, nil)
+    |> Enum.reduce(0, fn owner, acc ->
+      count_active_names(state, owner) + acc
+    end)
```
sborrazas, line 1043: I thought we weren't going to traverse through an unlimited list of results

lib/ae_mdw/sync/server.ex
```diff
@@ -311,21 +313,61 @@ defmodule AeMdw.Sync.Server do
...
+            |> get_balances_from_node(accounts_set)
+            |> Enum.reduce([], fn {account_pk, balance}, acc ->
+              [UpdateBalanceAccountMutation.new(account_pk, balance) | acc]
+            end)
+
+          update_gen_mutations(gen_mutations, block_index, account_balances_mutations)
+
+        {:error, :no_mblocks} ->
+          gen_mutations
+      end
```
sborrazas, line 331: Why did this change?

lib/ae_mdw/sync/server.ex
```diff
@@ -345,14 +387,20 @@ defmodule AeMdw.Sync.Server do
...
+          list |> Enum.max_by(fn {block_index, _block, _mutations} -> block_index end)
 
-    {block_index, mb_hash, mblock}
+        {:ok, mb_hash} = :aec_headers.hash_header(:aec_blocks.to_micro_header(mblock))
+
+        {:ok, {block_index, mb_hash, mblock}}
+    end
   end
 
   defp exec_mem_mutations(empty_state, gens_mutations, from_height) do
```
sborrazas, line 406: I would keep this file unchanged

lib/ae_mdw_web/controllers/name_controller.ex
```diff
@@ -171,6 +171,16 @@ defmodule AeMdwWeb.NameController do
     end
   end
 
+  @spec names_count(Conn.t(), map()) :: Conn.t()
+  def names_count(%Conn{assigns: %{state: state, query: query}} = conn, _params) do
+    state
+    |> Names.count_names(query)
+    |> case do
+      {:ok, count} -> format_json(conn, count)
+      {:error, _} -> {:error, ErrInput.NotFound.exception(value: "names")}
```
sborrazas, line 180: Why can't we use this same error and use a `with` expression?
```elixir
with {:ok, count} <- Names.count_names(state, query) do
  format_json(conn, count)
end
```

lib/ae_mdw/db/util.ex
```diff
@@ -358,4 +358,39 @@ defmodule AeMdw.Db.Util do
...
+
+        Model.account_names_count(index: owner_pk, count: 0) ->
+          Model.account_names_count(index: owner_pk, count: 0)
+
+        Model.account_names_count(index: owner_pk, count: count) ->
+          Model.account_names_count(index: owner_pk, count: count - 1)
+      end
+    )
+  end
```
sborrazas, line 395: Can we move these functions to the `AeMdw.Names` module? These functions are not going to be used anywhere else to be a util kind of thing

sborrazas (changes requested):

lib/ae_mdw/names.ex
yaboiishere: The list is very short, so i thought it would be ok. There's only one record per account

lib/ae_mdw/sync/server.ex
yaboiishere: The last_microblock function was trying to get the max element of an empty list and was crashing the sync process

lib/ae_mdw/sync/server.ex
yaboiishere: I can revert it to check if the issue with last_microblock persists after removing the new mutation

lib/ae_mdw/sync/server.ex
sborrazas: Let's move these changes to a separate PR and mark it as "fix: ..."

lib/ae_mdw/names.ex
yaboiishere: No, because it raises a warning that a `with` with one clause and an else, should be a `case`

lib/ae_mdw/names.ex
```diff
@@ -72,6 +72,17 @@ defmodule AeMdw.Names do
...
+  @spec count_names(state(), query()) :: {:ok, non_neg_integer()} | {:error, reason()}
+  def count_names(state, query) do
+    case Util.convert_params(query, &convert_param/1) do
+      {:ok, query_params} ->
+        {:ok, count_active_names(state, Map.get(query_params, :owned_by))}
+
+      {:error, error} ->
+        {:error, ErrInput.Query.exception(value: error)}
+    end
+  end
```
sborrazas, line 84: `else` clause not needed
```elixir
with {:ok, filters} <- Util.convert_params(query, &convert_param/1) do
  {:ok, count_active_names(state, Map.get(filters, :owned_by))
end
```

lib/ae_mdw/names.ex
sborrazas: We never go through unlimited lists, this wouldn't be a proper implementation. For total active names we can simply take the last stat from the database:
```elixir
defp count_active_names(state, nil) do
  case State.prev(state, Model.TotalStat, nil) do
    {:ok, Model.total_state(active_names: active_names) -> active_names
    :none -> 0
  end
end
```

lib/ae_mdw/sync/server.ex
```diff
@@ -294,7 +297,6 @@ defmodule AeMdw.Sync.Server do
 
   defp exec_db_height(state, blocks_mutations, clear_mem?) do
     blocks_mutations
-    |> maybe_add_accounts_balance_mutations()
```
sborrazas, line 297: These changes are not part of the feature

sborrazas (changes requested):

sborrazas (approved):

PR #1743, opened 2024-04-24
fix: extend auction properly

yaboiishere:


lib/ae_mdw/db/mutations/name_claim_mutation.ex
```diff
@@ -82,14 +82,21 @@ defmodule AeMdw.Db.NameClaimMutation do
...
+          protocol_version: protocol_version
         },
         state
       ) do
     m_plain_name = Model.plain_name(index: name_hash, value: plain_name)
 
     state2 = State.put(state, Model.PlainName, m_plain_name)
 
+    timeout =
+      if State.get(state2, Model.AuctionBid, plain_name) == :not_found do
```
sborrazas, line 94: We can use `State.exists?/3` here instead

lib/ae_mdw/db/mutations/name_claim_mutation.ex
```diff
@@ -48,7 +48,7 @@ defmodule AeMdw.Db.NameClaimMutation do
           boolean(),
           Txs.txi_idx(),
           Blocks.block_index(),
-          Names.auction_timeout()
+          non_neg_integer()
```
sborrazas, line 51: Can we turn into a type from the `Node.Db` module and use that globally?

sborrazas (changes requested):

sborrazas (approved):

Issue #1611, opened 2023-10-17
Dex trades general listing


janmichek:
Please expose the endpoint listing all dex trades

jyeshe:
> Please expose the endpoint listing all dex trades

Could you provide the list of contracts (addresses) for `ae_uat` and `ae_mainnet`? If not possible, which other criteria should we consider to classify an AEX9 transfer as a DEX trade?

janmichek:
@michele-franchi ?

michele-franchi:
@jyeshe @janmichek these are the DEX contracts:

https://github.com/aeternity/dex-contracts-v2#aeternity-deployed-contracts

vatanasov:
As per discussion with the aescan team, a route getting the swaps by contract id will be implemented.

janmichek:
How long does it usually take from a swap on dex to be projected into the response? @vatanasov

vatanasov:
It should be the time it takes for the mdw to sync with the node.

janmichek:
Which should usually be...?

vatanasov:
It should usually be almost instant.

PR #1762, opened 2024-05-09
feat: add route for dex swaps by contract_id

vatanasov:
refs: https://github.com/aeternity/ae_mdw/issues/1611

### DIFF
--- a/lib/ae_mdw/sync/dex_cache.ex
+++ b/lib/ae_mdw/sync/dex_cache.ex
@@ -112,4 +112,12 @@ defmodule AeMdw.Sync.DexCache do
       %{pair: contract_pk, token1: token1, token2: token2}
     end)
   end
+
+  @spec get_contract_pk_txi(pubkey(), State.t()) :: pos_integer() | nil
+  def get_contract_pk_txi(contract_pk, state) do
+    case Origin.tx_index(state, {:contract, contract_pk}) do
+      {:ok, create_txi} -> create_txi
+      :not_found -> nil
+    end
+  end
 end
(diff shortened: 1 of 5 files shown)


lib/ae_mdw/sync/dex_cache.ex
```diff
@@ -112,4 +112,12 @@ defmodule AeMdw.Sync.DexCache do
       %{pair: contract_pk, token1: token1, token2: token2}
     end)
   end
+
+  @spec get_contract_pk_txi(pubkey(), State.t()) :: pos_integer() | nil
+  def get_contract_pk_txi(contract_pk, state) do
```
sborrazas, line 117: Perhaps we can move this out of the cache module, since it doesn't use it

sborrazas (changes requested):
Can we move everything into the Dex module like we do with the rest of the resources (validate_contract_id, render_swap, fetch_contract_swaps, etc) ?

lib/ae_mdw_web/controllers/dex_controller.ex
```diff
@@ -48,16 +46,20 @@ defmodule AeMdwWeb.DexController do
   def swaps(%Conn{assigns: assigns} = conn, %{"from_symbol" => token_symbol}) do
     %{state: state, pagination: pagination, cursor: cursor} = assigns
 
-    with {:ok, create_txi} <- validate_token(token_symbol),
+    with {:ok, create_txi} <- Dex.validate_token(token_symbol),
```
sborrazas, line 49: I meant removing validations from the controller as well. I wouldn't want the controller to know about the `create_txi` or any other dex internal concept. Controllers should all look like this
```elixir
def route(conn, params) do
  with {:ok, something} <- Context.fetch_something(state, ...) do
    Util.paginate(conn, something, ...)
  end  
end
```

sborrazas (approved):

yaboiishere (approved):

Issue #1425, opened 2023-06-27
Add further fields to include approximate time


sborrazas:
The fields and endpoints are the following:

- `approximate_last_updated_time` in `channels/{stateChannelId}`
- `approximate_last_updated_time` in `channels` endpoint
- `approximate_activated_time` in `names` (with expand=true) endpoint
- `approximate_auction_end_time` in `names/auctions` endpoint
- `approximate_expiration_time` and `approximate_auction_end_time` in `names/{name}` endpoint
- `approximate_created_time` in `accounts/{nameHash}/activities` (for cases if `micro_time` is not available) endpoint
- `approximate_created_time`  in  `contracts/logs?contract_id={contractId}` endpoint
- `approximate_activated_time` in `oracles` endpoint
- `approximate_registered_time` in `oracles/{oracleId}` endpoint

refs #831

janmichek:
In the state channel responses, I see the attribute name `lastUpdatedTime` so this is the exact time, not an approximate?

sborrazas:
> In the state channel responses, I see the attribute name `lastUpdatedTime` so this is the exact time, not an approximate?

@janmichek that is correct

janmichek:
Question for the oracles responses:

![image](https://github.com/aeternity/ae_mdw/assets/15363559/a16fcd1f-94db-4db9-a001-c44b792372ed)

So the `registerTime` is the time value of `activeFrom`?
What is the value of `register` for? Is it a keyblock value? How it can be registered in future?

sborrazas:
@janmichek the value for `register` is the transaction index of the register, same with `extends`. We had to keep txis for backwards compatibility, but please use `tx_hash=true` for these endpoints, which will be the behavior of v3. e.g. https://mainnet.aeternity.io/mdw/v2/oracles/ok_2NRBaMsgSDjZRFw4dU82KCqLa5W7aQdbJAzaFprTpjEGLAzroV?tx_hash=true

sborrazas:
@janmichek that's true, `approximate_auction_end` should be present there to indicate when the auction will terminate, as oppose to expire_time which indicates when the name will expire. I'll add it to names

janmichek:
> approximate_activated_time in names (with expand=true) endpoint

It seems to be missing (or some other similarly named value) in response

https://mainnet.aeternity.io/mdw/v2/names?state=active&by=activation&direction=backward&limit=4&expand=true&by=activation

janmichek:
ping

[...]

PR #1763, opened 2024-05-09
Merge name/pointers endpoint into the name endpoint

yaboiishere:
ref: #1425


lib/ae_mdw_web/router.ex
```diff
@@ -88,8 +88,6 @@ defmodule AeMdwWeb.Router do
       get "/names/auctions/:id", NameController, :auction
       get "/names/auctions/:id/claims", NameController, :auction_claims
       get "/names/:id", NameController, :name
-      get "/names/:id/pointers", NameController, :pointers
-      get "/names/:id/pointees", NameController, :pointees
```
sborrazas, line 92: Why is pointees route being removed?

yaboiishere: I've removed it by mistake

Issue #1750, opened 2024-04-29
Docker container crash with the defaults


dincho:
```
docker run -it aeternity/ae_mdw:1.74.4
```

crashes with:

```
11:31:28.655 [error] gen_server <0.1339.0> terminated with reason: bad argument in call to ets:next(async_tasks_pending, {store_acc_balance,[]}) in 'Elixir.AeMdw.Sync.AsyncTasks.Store':next_unprocessed/1 line 60
11:31:28.656 [error] gen_server <0.1344.0> terminated with reason: bad argument in call to ets:next(async_tasks_pending, {store_acc_balance,[]}) in 'Elixir.AeMdw.Sync.AsyncTasks.Store':next_unprocessed/1 line 60
11:31:28.658 [error] CRASH REPORT Process <0.1344.0> with 0 neighbours crashed with reason: bad argument in call to ets:next(async_tasks_pending, {store_acc_balance,[]}) in 'Elixir.AeMdw.Sync.AsyncTasks.Store':next_unprocessed/1 line 60
11:31:28.658 [error] gen_server <0.1340.0> terminated with reason: bad argument in call to ets:next(async_tasks_pending, {store_acc_balance,[]}) in 'Elixir.AeMdw.Sync.AsyncTasks.Store':next_unprocessed/1 line 60
11:31:28.658 [error] gen_server <0.1346.0> terminated with reason: bad argument in call to ets:next(async_tasks_pending, {store_acc_balance,[]}) in 'Elixir.AeMdw.Sync.AsyncTasks.Store':next_unprocessed/1 line 60
11:31:28.658 [error] gen_server <0.1341.0> terminated with reason: bad argument in call to ets:next(async_tasks_pending, {store_acc_balance,[]}) in 'Elixir.AeMdw.Sync.AsyncTasks.Store':next_unprocessed/1 line 60
11:31:28.658 [error] gen_server <0.1348.0> terminated with reason: bad argument in call to ets:next(async_tasks_pending, {store_acc_balance,[]}) in 'Elixir.AeMdw.Sync.AsyncTasks.Store':next_unprocessed/1 line 60
11:31:28.658 [error] gen_server <0.1342.0> terminated with reason: bad argument in call to ets:next(async_tasks_pending, {store_acc_balance,[]}) in 'Elixir.AeMdw.Sync.AsyncTasks.Store':next_unprocessed/1 line 60
[...]

PR #1773, opened 2024-05-16
fix: arm docker incompatibility

vatanasov:
refs https://github.com/aeternity/ae_mdw/issues/1750

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@
 #
 ARG ELIXIR_VERSION=1.16.2
 ARG OTP_VERSION=26.2.4
-ARG DEBIAN_VERSION=bullseye-20240423-slim
+ARG DEBIAN_VERSION=bullseye-20240408-slim
 
 ARG BUILDER_IMAGE="hexpm/elixir:${ELIXIR_VERSION}-erlang-${OTP_VERSION}-debian-${DEBIAN_VERSION}"
 ARG RUNNER_IMAGE="debian:${DEBIAN_VERSION}"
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -43,14 +43,14 @@ ARG DEV_MODE="false"
 ENV DEV_MODE=${DEV_MODE}
 ENV NODEROOT=/home/aeternity/node/local
 ARG NODE_VERSION=7.0.0
-ARG NODE_URL=https://github.com/aeternity/aeternity/releases/download/v${NODE_VERSION}/aeternity-v${NODE_VERSION}-ubuntu-x86_64.tar.gz
 ENV NODEDIR=/home/aeternity/node/local/rel/aeternity
 RUN mkdir -p ./local/rel/aeternity/data/mnesia
-RUN curl -L --output aeternity.tar.gz ${NODE_URL} && tar -C ./local/rel/aeternity -xf aeternity.tar.gz
 
+COPY --from=aeternity/aeternity:v7.0.0-bundle /home/aeternity/node ./local/rel/aeternity
```
dincho, line 49: - you shouldn't need the `-bundle` flavour
- the version argument should be used to allow rebuilds with different version without changing the dockerfile and to not repeat the version below

```suggestion
COPY --from=aeternity/aeternity:v${NODE_VERSION} /home/aeternity/node ./local/rel/aeternity
```

Dockerfile
```diff
@@ -135,6 +135,10 @@ COPY --from=builder /home/aeternity/node/local ./local
 COPY ./docker/aeternity.yaml /home/aeternity/aeternity.yaml
 COPY ./docker/aeternity-dev.yaml /home/aeternity/aeternity-dev.yaml
 COPY ./docker/healthcheck.sh /home/aeternity/healthcheck.sh
+
+COPY --from=aeternity/aeternity:v7.0.0-bundle /usr/local/lib/librocksdb.so.7.10.2 /usr/local/lib/
```
dincho, line 139: same as above

```suggestion
COPY --from=aeternity/aeternity:v${NODE_VERSION} /usr/local/lib/librocksdb.so.7.10.2 /usr/local/lib/
```

Dockerfile
```diff
@@ -135,6 +135,10 @@ COPY --from=builder /home/aeternity/node/local ./local
...
+
+RUN ln -fs librocksdb.so.7.10.2 /usr/local/lib/librocksdb.so.7.10     && ln -fs librocksdb.so.7.10.2 /usr/local/lib/librocksdb.so.7     && ln -fs librocksdb.so.7.10.2 /usr/local/lib/librocksdb.so     && ldconfig
```
dincho, line 141: please reformat to multiline and remove the extra spaces i.e. https://github.com/aeternity/aeternity/blob/master/Dockerfile#L24-L28

dincho (commented):
Perhaps `aeternity/aeternity` image should be an argument as well for extra flex

dincho (approved):

PR #1794, opened 2024-05-31
fix: handle transfers txi_idx ref tuple

sborrazas:


sborrazas:
> I think the type in the model should also be changed?

The Model type is already the correct one: https://github.com/aeternity/ae_mdw/blob/master/lib/ae_mdw/db/model.ex#L1032

sborrazas:
> > > I think the type in the model should also be changed?
> > 
> > 
> > The Model type is already the correct one: https://github.com/aeternity/ae_mdw/blob/master/lib/ae_mdw/db/model.ex#L1032
> 
> Right, but isn't the deserialization of the cursor also affecting the `TargetKindIntTransferTx` and `KindIntTransferTx` tables? They have `txi` as the fourth element in their index tuple.

You're right, those weren't adjusted. Done now. Thanks!

vatanasov (approved):

yaboiishere (approved):

PR #1799, opened 2024-06-05
feat: mark aex9 as invalid

vatanasov:
Refs: https://github.com/aeternity/ae_mdw/issues/1257 . Marking invalid Aex9 contracts as such, also unifies the logic for calculating the holders.


lib/ae_mdw/aex9.ex
```diff
@@ -39,6 +39,16 @@ defmodule AeMdw.Aex9 do
...
 
+  @spec invalid_number_of_holders() :: String.t()
+  def invalid_number_of_holders() do
+    "Invalid number of holders"
+  end
+
+  @spec invalid_holder_balance() :: String.t()
+  def invalid_holder_balance() do
+    "Invalid holder balance"
+  end
```
sborrazas, line 50: What's the purpose of these functions? Can we use module a module attribute or directly include it in the code instead?

lib/ae_mdw/aexn_tokens.ex
```diff
@@ -76,10 +80,13 @@ defmodule AeMdw.AexnTokens do
       state
       |> Collection.stream(@aexn_creation_table, direction, key_boundary, cursor)
       |> Stream.map(fn {aexn_type, txi_idx} ->
-        Model.aexn_contract_creation(contract_pk: pubkey) =
-          State.fetch!(state, @aexn_creation_table, {aexn_type, txi_idx})
-
-        State.fetch!(state, @aexn_table, {aexn_type, pubkey})
+        State.fetch!(state, @aexn_creation_table, {aexn_type, txi_idx})
+      end)
+      |> Stream.reject(fn Model.aexn_contract_creation(contract_pk: pubkey) ->
```
sborrazas, line 85: If we ignore the contracts from the listings we're going to get a lot of complaints, what I would do instead is display `invalid: true` when rendering them

lib/ae_mdw/error.ex
```diff
@@ -25,6 +25,7 @@ defmodule AeMdw.Error do
   def to_string(Input.Expired, x), do: concat("expired", x)
   def to_string(Input.NotAex9, x), do: concat("not AEX9 contract", x)
   def to_string(Input.NotAex141, x), do: concat("not AEX141 contract", x)
+  def to_string(Input.AexnContractInvalid, x), do: concat("invalid AEXN contract", x)
```
sborrazas, line 28: 👍

lib/ae_mdw_web/controllers/aexn_token_controller.ex
```diff
@@ -90,21 +90,22 @@ defmodule AeMdwWeb.AexnTokenController do
       query: query
     } = assigns
 
-    with {:ok, paginated_balances} <-
-           Aex9.fetch_event_balances(state, contract_id, pagination, cursor, order_by, query) do
+    with {:ok, contract_pk} <- validate_aex9(contract_id, state),
+         {:ok, paginated_balances} <-
+           Aex9.fetch_event_balances(state, contract_pk, pagination, cursor, order_by, query) do
```
sborrazas, line 95: Why are we moving validations back to the controller? Let's leave them on the context modules, validations, rendering and paginating.

lib/mix/tasks/gen.migration.ex
sborrazas, line 1: Let's get rid of this file

sborrazas (changes requested):

lib/ae_mdw/aex9.ex
vatanasov: They are used in different modules, so a module attribute won't do. As for directly in the code - there could be a hard to notice difference, and a change in the future might be incomplete. Having them in a single place has a better guarantee for consistency. 

The other way we could do it is with macros, that just returns the strings - that way we could also use them in pattern matches if we want, the downside is that we have to `require` the module that contains them (I personally don't mind this, rather I prefer it).

sborrazas: Got it, it's being reused in several places, so perhaps we can do something like `def invalid_reason(:holder_balance)` and have dialyzer check if the input is either `:holder_balance | :other_reason`

vatanasov: Do we gain something doing it that way though? Because now instead of a compile-time error, we potentially have a static analysis one?

sborrazas: Functions should describe what these values represent, not what these values are. Which is why we write
```elixir
def maximum_transactions, do: 77

if length(transactions) >= maximum_transactions() do ...
```
and not
```elixir
def seventy_seven, do: 77

if length(transactions) >= seventy_seven() do ...
```

These are reasons for a contract being invalid, not an invalid holder's balance or an invalid number of holders

vatanasov: Going with that, the function can be named `invalid_holder_balance_message`, that way it would describe what the value represents. Even if that was not possible though, I think sacrificing the intended use of a language features to gain better guarantees (compile-time checks are always better than dialyzer ones) is worth it.

In any case, I see you feel strongly about it and I don't think in this particular case it would provide much additional value, I will switch it as suggested.

sborrazas: I agree compile-time errors are way superior to external static checks like dialyzer, so I wouldn't mind either way `invalid_holder_balance_reason` is fine to me

PR #1800, opened 2024-06-05
chore: task for generating migrations

vatanasov:


yaboiishere (approved):

lib/mix/tasks/gen.migration.ex
```diff
@@ -0,0 +1,61 @@
...
+
+        if Path.wildcard(fuzzy_path) != [] do
+          Mix.raise(
+            "migration can't be created, there is already a migration file with name #{name}."
+          )
+        end
+
+        # The :change option may be used by other tasks but not the CLI
+        assigns = [
+          # credo:disable-for-next-line
```
sborrazas, line 29: Let's avoid skipping credo errors. We're using them for a reason

lib/mix/tasks/gen.migration.ex
```diff
@@ -0,0 +1,61 @@
...
+  defp timestamp do
+    {{y, m, d}, {hh, mm, ss}} = :calendar.universal_time()
+    "#{y}#{pad(m)}#{pad(d)}#{pad(hh)}#{pad(mm)}#{pad(ss)}"
+  end
+
+  defp pad(i) when i < 10, do: <<?0, ?0 + i>>
+  defp pad(i), do: to_string(i)
+
+  embed_template(:migration, """
+  defmodule <%= inspect @mod %> do
```
sborrazas, line 54: We can do `AeMdw.Migration.<%= camelized_name %>` here to avoid using macro functions

lib/mix/tasks/gen.migration.ex
```diff
@@ -0,0 +1,61 @@
...
+    @spec run(State.t(), boolean()) :: {:ok, non_neg_integer()}
```
sborrazas, line 55: Can we add the State alias `alias AeMdw.Db.State?

sborrazas (changes requested):

lib/mix/tasks/gen.migration.ex
vatanasov: If we use `safe_concat` here (as it suggests) it fails, because the atom is not yet created. The warning is specifically against dynamically creating atoms - but this is exactly what is wanted here.

sborrazas (approved):

PR #1824, opened 2024-06-25
feat: add oracle extends endpoint on v3

sborrazas:


yaboiishere (changes requested):
This needs to be added to the swagger

yaboiishere (approved):

PR #1830, opened 2024-07-03
fix: use same format for pointers as the node

sborrazas:


lib/ae_mdw/db/name.ex
```diff
@@ -29,6 +29,7 @@ defmodule AeMdw.Db.Name do
   @typep height :: Blocks.height()
   @typep txi_idx :: AeMdw.Txs.txi_idx()
   @typep state :: State.t()
+  @typep pointer() :: %{key: binary(), id: binary()}
```
vatanasov, line 32: This type seems to differ from the openapi specs (the `encoded_key` is missing)

sborrazas: Done, thanks

vatanasov (approved):

Issue #1549, opened 2023-08-23
Make token search case insensitive


janmichek:
The search results from endpoint `/v2/aex9?prefix=${query}` are sensitive for search query. 
It's not very user friendly and cannot be resolved on frontend

Extra points if you could add search by `symbol`. It's only possible to search by `name`

janmichek:
The same insensitiveness would be nice to do for a search by NFT name/symbol

yaboiishere:
Should this be the case for the aex141 endpoint?

yaboiishere:
It's implemented for aex9 and aex141. Also it's already possible to search by the symbol, just isn't documented. I've updated the swagger spec. The way it's implemented is that the prefix or exact parameter works for symbols or names depending on the value of the 'by' parameter. By default it's name, so it works for the names if 'by' isn't passed

janmichek:
Good job! When do you plan to deploy this?

yaboiishere:
I hope I'll be done with it soon, Valentin found a pretty big issue with the way I was doing this and I've had to create a new table, which introduced some new problems with the request.

janmichek:
It seems to work now. Do you plan to work on it?

yaboiishere:
It's done, I've forgotten to give an update here

PR #1836, opened 2024-07-09
feat: make aexn prefix search case insensitive

yaboiishere:
ref: #1549

### DIFF
--- a/lib/ae_mdw/aexn_tokens.ex
+++ b/lib/ae_mdw/aexn_tokens.ex
@@ -127,7 +127,28 @@ defmodule AeMdw.AexnTokens do
       {type, prefix <> Util.max_name_bin(), <<>>}
     }
 
-    do_build_tokens_streamer(state, table, cursor, scope)
+    results = do_build_tokens_streamer(state, table, cursor, scope)
+
+    downcase_scope = {
+      {type, String.upcase(prefix), <<>>},
+      {type, String.upcase(prefix) <> Util.max_name_bin(), <<>>}
+    }
+
+    downcase_results = do_build_tokens_streamer(state, table, cursor, downcase_scope)
+
+    upcase_scope = {
+      {type, String.downcase(prefix), <<>>},
+      {type, String.downcase(prefix) <> Util.max_name_bin(), <<>>}
+    }
+
+    upcase_results = do_build_tokens_streamer(state, table, cursor, upcase_scope)
+
+    fn direction ->
+      Collection.merge(
+        [results.(direction), downcase_results.(direction), upcase_results.(direction)],
+        direction
+      )
+    end
   end
 
   defp build_tokens_streamer(_filters, state, type, table, cursor),
--- a/lib/ae_mdw/aexn_tokens.ex
+++ b/lib/ae_mdw/aexn_tokens.ex
@@ -46,8 +46,7 @@ defmodule AeMdw.AexnTokens do
   }
   @downcased_sorting_table %{
     name: @aexn_downcased_name_table,
-    symbol: @aexn_downcased_symbol_table,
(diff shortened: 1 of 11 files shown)


lib/ae_mdw/aexn_tokens.ex
```diff
@@ -42,6 +44,11 @@ defmodule AeMdw.AexnTokens do
     symbol: @aexn_symbol_table,
     creation: @aexn_creation_table
   }
+  @downcased_sorting_table %{
+    name: @aexn_downcased_name_table,
+    symbol: @aexn_downcased_symbol_table,
+    creation: @aexn_creation_table
```
sborrazas, line 50: This key is never going to be accessed

lib/ae_mdw/aexn_tokens.ex
```diff
@@ -110,24 +115,33 @@ defmodule AeMdw.AexnTokens do
...
     }
 
-    do_build_tokens_streamer(state, table, cursor, scope)
+    do_build_tokens_streamer(state, sorting_table, cursor, scope)
   end
 
-  defp build_tokens_streamer(%{prefix: prefix}, state, type, table, cursor) do
+  defp build_tokens_streamer(%{prefix: prefix}, state, type, order_by, cursor) do
+    sorting_table = Map.fetch!(@downcased_sorting_table, order_by)
     prefix = String.slice(prefix, 0, @max_sort_field_length)
```
sborrazas, line 131: Can we do the downcasing here instead?

lib/ae_mdw/aexn_tokens.ex
```diff
@@ -110,24 +115,33 @@ defmodule AeMdw.AexnTokens do
...
 
+    cursor =
+      if not is_nil(cursor) do
+        {type, symbol_name, pubkey} = cursor
+        {type, String.downcase(symbol_name), pubkey}
+      end
```
sborrazas, line 137: This block isn't needed since these values are the values obtained from the downcased table keys

priv/migrations/20240712092935_add_normalized_aexn_tables.ex
```diff
@@ -0,0 +1,42 @@
...
+      |> Enum.reduce(new_state, fn {type, symbol, pubkey}, acc_state ->
+        State.put(
+          acc_state,
+          Model.AexnContractDowncasedSymbol,
+          Model.aexn_contract_downcased_symbol(
+            index: {type, String.downcase(symbol), pubkey},
+            original_symbol: symbol
+          )
+        )
+      end)
```
sborrazas, line 38: Can we group these inside a transaction or several transactions using this writemutation to make it faster?

sborrazas (changes requested):

lib/ae_mdw/aexn_tokens.ex
sborrazas: The cursor is formed by using the keys from the table which are always already downcased

sborrazas: Doesn't the cursor come from the serialization of `do_build_tokens_streamer(state, table, cursor, scope)` where the `table` is one of these downcased tables?

yaboiishere: At the end of that function the results are from the AexnContract table, so after it when the stream is being paginated it returns the cursor based on that table

sborrazas: I see that now! You're right. We should change it, though, since we don't need to `fetch!` records to form the pagination next/prev urls, that's not necessary. I'll update it later. Thanks for clarification 🙏

sborrazas (approved):

Issue #1759, opened 2024-05-08
Accept `/mdw` path in the container ENV variable


davidyuk:
Currently, at https://mainnet.aeternity.io/mdw/v2/api we have
```json
"servers": [{
  "url": "/mdw/v2",
  "description": "V2 version path"
}],
```
which is correct, but when I'm running mdw image locally at http://localhost:4000/v2/api there are still
```json
"servers": [{
  "url": "/mdw/v2",
  "description": "V2 version path"
}],
```
It is incorrect, it should be just `/v2`. I see several ways to solve it.
- try the relative url, like `./`, not sure if is it supported by clients
- try to guess the right server url by a request url, like if openapi is requested at http://example.com/path/to-mdw/v2/api then `servers[0].url` should be `/path/to-mdw/v2`
- accept mdw path in ENV variable, so it should be set correctly according to the environment

The last way assumes openapi spec should be served by a handler that would substitute the path to mdw in runtime. This seems to be the most easy way.

vatanasov:
Solved in #1837, there is a `PATH_PREFIX` ENV variable now.

PR #1837, opened 2024-07-09
fix: correct url in openapi specs

vatanasov:
refs: https://github.com/aeternity/ae_mdw/issues/1759
Decided to go with the second proposed solution, it would seem to work for the different cases and it would not require additional configurations.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -98,6 +98,7 @@ COPY config/runtime.exs config/
 
 # Generate swagger V2 file
 RUN mix run --no-start -e 'IO.puts(Mix.Project.config[:version])' >AEMDW_VERSION
+ARG PATH_PREFIX
 RUN scripts/swagger-docs.py >priv/static/swagger/swagger_v3.yaml
 
 # Install devmode
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -8,6 +8,7 @@ services:
         RUNNER_IMAGE: "hexpm/elixir:1.16.2-erlang-26.2.4-debian-bullseye-20240408-slim"
         USER: root
         MIX_ENV: dev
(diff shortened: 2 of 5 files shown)


sborrazas:
Have you tried this with the staging/prod setup? I don't think this works as intended, we are running the middleware app under the `/mdw/` namespace via an nginx proxy, which removes the `/mdw/` from the path to call the middleware app, which is how the middleware responds

vatanasov:
> Have you tried this with the staging/prod setup? I don't think this works as intended, we are running the middleware app under the `/mdw/` namespace via an nginx proxy, which removes the `/mdw/` from the path to call the middleware app, which is how the middleware responds

Ah, it seems you are correct, I will try something else then, thank you.

Dockerfile
```diff
@@ -98,6 +98,7 @@ COPY config/runtime.exs config/
 
 # Generate swagger V2 file
 RUN mix run --no-start -e 'IO.puts(Mix.Project.config[:version])' >AEMDW_VERSION
+ARG PATH_PREFIX
```
sborrazas, line 101: I don't think ARGs in dockerfiles are env variables. This wouldn't be accessed on the swagger script

sborrazas (changes requested):

Dockerfile
vatanasov: They are, for `RUN` commands: https://docs.docker.com/reference/dockerfile/#using-arg-variables . If you wanna test it, you can rebuild the docker container using the `docker-compose-dev.yml` included in the PR

sborrazas: You're right, it works the same! sorry, merging

sborrazas (approved):

Issue #1850, opened 2024-07-22
Missed definitions of transaction properties provided by node in v3


davidyuk:
https://github.com/aeternity/ae_mdw/blob/6f6f7477a0f66dea65c8610708015df4243294f5/docs/swagger_v3/transactions.spec.yaml#L29-L32

In js sdk I'm using autorest which converts snake_case properties to camelCase and this stops working if properties are not defined. So, instead
```js
          tx: {
            accountId: 'ak_2Fh6StA76AKdy8qsGdkEfkQyVmAYc2XE1irWRnDgXKhmBLKoXg',
            nameFee: 500000000000000,
            nameId: 'nm_27MpcZebrNCLB8AWaJBjp6TrzdjwMWKDJQDBnTWb82E2sut7P7',
            nameSalt: 0,
            ...
          },
```
I'm getting
```js
          tx: {
            account_id: 'ak_2Fh6StA76AKdy8qsGdkEfkQyVmAYc2XE1irWRnDgXKhmBLKoXg',
            name_fee: 500000000000000,
            name_id: 'nm_27MpcZebrNCLB8AWaJBjp6TrzdjwMWKDJQDBnTWb82E2sut7P7',
            name_salt: 0,
            ...
          },
```
There should be a way to reference `tx` definition from aenode's api.

PR #1862, opened 2024-07-23
feat: add node schemas to swagger definitions for mdw

yaboiishere:
ref: #1850

### DIFF
--- a/docs/swagger_v3/transactions.spec.yaml
+++ b/docs/swagger_v3/transactions.spec.yaml
@@ -38,9 +38,6 @@ schemas:
       tx:
         description: The transaction
         $ref: '#/components/schemas/Tx' 
-      ttl:
-        description: The time to live
-        type: integer
     example:
       block_hash: mh_ufiYLdN8am8fBxMnb6xq2K4MQKo4eFSCF5bgixq4EzKMtDUXP
       block_height: 1
@@ -59,7 +56,6 @@ schemas:
         sender_id: ak_26dopN3U2zgfJG4Ao4J4ZvLTf5mqr7WAgLAq6WxjxuSapZhQg5
         type: SpendTx
         version: 1
-        ttl: 10000
 paths:
   /transactions:
     get:
--- a/docs/swagger_v3/transactions.spec.yaml
+++ b/docs/swagger_v3/transactions.spec.yaml
@@ -56,6 +56,7 @@ schemas:
         sender_id: ak_26dopN3U2zgfJG4Ao4J4ZvLTf5mqr7WAgLAq6WxjxuSapZhQg5
         type: SpendTx
         version: 1
+        ttl: 10000
 paths:
   /transactions:
     get:
--- a/docs/swagger_v3/transactions.spec.yaml
+++ b/docs/swagger_v3/transactions.spec.yaml
@@ -37,8 +37,10 @@ schemas:
           type: string
       tx:
         description: The transaction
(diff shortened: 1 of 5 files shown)


davidyuk (approved):
It looks good, the only concern is that there would be some node schemas not used in middleware

yaboiishere:
Yeah, I started by trying to get only specific ones, but I think It's worth it to be a little less specific, so that if we need a definition it's there. If it's a problem I can get only the specific schemas and make it easy to add more

scripts/swagger-docs.py
```diff
@@ -23,14 +24,18 @@
...
+
 for filepath in Path(SWAGGER_DOCS_DIR).glob("*.spec.yaml"):
     with open(filepath, 'r') as stream:
         data_loaded = yaml.safe_load(stream)
         schemas = {**schemas, **data_loaded['schemas']}
         paths = {**paths, **data_loaded['paths']}
 
 swagger['paths'] = paths
-swagger['components']['schemas'] = {**swagger['components']['schemas'], **schemas}
+swagger['components']['schemas'] = {**swagger['components']['schemas'], **schemas, **node_schemas}
```
davidyuk, line 38: `node_schemas` have own schemas for `Channel`, `KeyBlock`, `OracleQuery`, `Status`. Looks like it overrides mdw's schemas.

davidyuk: I propose to add a code removing schemas not referenced in other places, like this
```python
schemas = swagger['components']['schemas']
schemas_count = 0
while schemas_count != len(schemas):
    schemas_count = len(schemas)
    swagger_string = str(swagger)
    for name in list(schemas):
        if '#/components/schemas/' + name not in swagger_string:
            del schemas[name]
```
It needs to be run multiple times to remove a chain of schemas referencing each other, but not referenced in other places. It reduces the schemas count from 142 to 97. This way you don't need to import specific schemas, but there won't be extra schemas in the final document.

davidyuk (commented):
In general, I don't have problems with extra schemas because autorest removes them (at least I think so). But won't they be rendered in Swagger UI confusing developers?

docs/swagger_v3/transactions.spec.yaml
```diff
@@ -59,7 +56,6 @@ schemas:
         sender_id: ak_26dopN3U2zgfJG4Ao4J4ZvLTf5mqr7WAgLAq6WxjxuSapZhQg5
         type: SpendTx
         version: 1
-        ttl: 10000
```
davidyuk, line 60: This `ttl` is fine because it is in `tx` field.

davidyuk (approved):

vatanasov:
Maybe look into adding a command to `scripts/do.sh`, that would copy the `node_oas3.yaml` file from the docker container to the swagger directory and then run the `swagger-docs.py`. Otherwise while developing you need to get the node file from somewhere (which is kinda confusing) and than it will probably be left stale, unless you remember to update it manually the next time you are working on the openapi schemas. It's not that big of a problem, but having it automated would be less error-prone and provide better development experience.

davidyuk:
I found out that we can pass an external url to `$ref` 🙃 
https://swagger.io/docs/specification/using-ref/

yaboiishere:
That was my first aproach, but then we would have to figure out what the version of the node that's running with the mdw, so that we reference the correct swagger.

sborrazas (approved):

Issue #1858, opened 2024-07-23
Add `/v3/stats` to Open Api


davidyuk:
Middleware has an endpoint [`/v3/stats`](https://mainnet.aeternity.io/mdw/v3/stats), but I can't find it definition in [`/v3/api`](https://mainnet.aeternity.io/mdw/v3/api)

vatanasov:
Solved in https://github.com/aeternity/ae_mdw/pull/1863 (docs: https://mainnet.aeternity.io/mdw/swagger/#/default/GetStats)

PR #1863, opened 2024-07-23
feat: add openapi missing stats route

vatanasov:
refs: https://github.com/aeternity/ae_mdw/issues/1858

### DIFF
--- a/docs/swagger_v3/stats.spec.yaml
+++ b/docs/swagger_v3/stats.spec.yaml
@@ -203,6 +203,39 @@ schemas:
     example:
       miner: ak_2whjDhTTmbN13vU7gAUsRbosBhmycho4h8LqHVqKwyGofDetQ9
       total_reward: 945000000000000000000
+  Stats:
+    type: object
+    description: Stats
+    properties:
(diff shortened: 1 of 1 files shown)


yaboiishere (approved):

docs/swagger_v3/stats.spec.yaml
```diff
@@ -426,3 +459,21 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorResponse'
+  /stats:
+    get:
+      deprecated: true
+      description: Deprecated. Use /totalstats or /deltastats instead.
```
sborrazas, line 465: This is not deprecated, the /stats endpoint provides data that /totalstats or /deltatstats doesn't provide

sborrazas (changes requested):

docs/swagger_v3/stats.spec.yaml
vatanasov: My bad, mistakenly read the comment above the v1 version.

sborrazas (approved):

PR #1866, opened 2024-07-24
fix: calculate auction expiration correctly when extension present

sborrazas:
Requires a full re-sync for prior changes to work

### DIFF
--- a/lib/ae_mdw/db/mutations/name_claim_mutation.ex
+++ b/lib/ae_mdw/db/mutations/name_claim_mutation.ex
@@ -90,113 +90,107 @@ defmodule AeMdw.Db.NameClaimMutation do
 
     state2 = State.put(state, Model.PlainName, m_plain_name)
 
-    timeout =
-      if State.exists?(state2, Model.AuctionBid, plain_name) do
-        :aec_governance.name_claim_bid_extension(plain_name, protocol_version)
-      else
-        :aec_governance.name_claim_bid_timeout(plain_name, protocol_version)
-      end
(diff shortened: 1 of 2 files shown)


lib/ae_mdw/db/mutations/name_claim_mutation.ex
```diff
@@ -90,113 +90,107 @@ defmodule AeMdw.Db.NameClaimMutation do
...
+          {:ok,
+           Model.auction_bid(
+             start_height: start_height,
+             block_index_txi_idx: {_bi, prev_txi_idx},
+             expire_height: prev_auction_end,
+             owner: prev_owner
+           ) = auction_bid} ->
+            auction_end =
+              prev_auction_end +
+                :aec_governance.name_claim_bid_extension(plain_name, protocol_version)
```
yaboiishere, line 156: The auction should only be extended if the bid is near the end (https://forum.aeternity.com/t/the-proposed-changes-in-ceres-protocol/12056#aens-adjusted-auction-parameters-7)

sborrazas: It's extended when auction already exists, which is this case, am I missing anything? https://github.com/aeternity/aeternity/blob/master/apps/aeprimop/src/aeprimop.erl#L936

vatanasov (approved):

lib/ae_mdw/db/mutations/name_claim_mutation.ex
yaboiishere: It's ok in that case

yaboiishere (approved):

PR #1867, opened 2024-07-24
fix: filter aex9 balances by amount correctly

vatanasov:
refs: https://github.com/aeternity/ae_mdw/issues/1257

### DIFF
--- a/lib/ae_mdw/aex9.ex
+++ b/lib/ae_mdw/aex9.ex
@@ -123,7 +123,11 @@ defmodule AeMdw.Aex9 do
       contract_id = encode_contract(contract_pk)
 
       paginated_balances =
-        (&Collection.stream(state, Model.Aex9BalanceAccount, &1, key_boundary, cursor_key))
+        fn direction ->
+          state
+          |> Collection.stream(Model.Aex9BalanceAccount, direction, key_boundary, cursor_key)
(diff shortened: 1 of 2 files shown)


lib/ae_mdw/aex9.ex
```diff
@@ -123,7 +123,11 @@ defmodule AeMdw.Aex9 do
       contract_id = encode_contract(contract_pk)
 
       paginated_balances =
-        (&Collection.stream(state, Model.Aex9BalanceAccount, &1, key_boundary, cursor_key))
+        fn direction ->
+          state
+          |> Collection.stream(Model.Aex9BalanceAccount, direction, key_boundary, cursor_key)
+          |> Stream.reject(fn {_contract_pk, amount, _account_pk} -> amount == 0 end)
```
sborrazas, line 129: This is going to cause to potentially have to iterate over an unlimited list of records. To avoid this you can include the minimum of 0 in the key_boundary:
```elixir
      key_boundary = {{contract_pk, 0, <<>>}, {contract_pk, Util.max_int(), <<>>}}
```

vatanasov: The amount doesn't go to less than 0, and we have to filter out the ones that equal 0, so with this suggestion, the stream is going the contain the same entries.

sborrazas: You're right, amount is an integer, though, right? We can start at 1

Issue #1872, opened 2024-07-26
"register" type different in `/oracles` and `/oracles/{id}`


davidyuk:
In `/oracles` it is a transaction object as per specification.

```json
// 20240726165838
// https://mainnet.aeternity.io/mdw/v3/oracles
{
  "data": [
    {
      "active": true,
      "register": {
        "block_hash": "mh_2f1gyBmtMMb8Sd3kbSu95cADRMwsYE8171KXP4W8wa2osRp4tZ",
        "block_height": 289005,
        "encoded_tx": "tx_+JELAfhCuEBX6CvfGClpszz2kg4qjUL8BtvNE0JpCtuwBHbZwGfTd6O6cQ9fr723QhuML50EIzZlFafB6UUi1y5S6RpZH2wJuEn4RxYBoQFtrWD1RJ5rxuHksnEXADAPHPT2jRT6uqVIr8X45BNYqgGFcXVlcnmIcmVzcG9uc2UAAIOYloCHBowmEhDKAIMEauEAmOjXSw==",
        "hash": "th_K5aPLdEN4H6QduiFtqdkv61gUCvaQpDjX3z9pHKNopD8F65LJ",
        "micro_index": 20,
        "micro_time": 1595571086808,
        "signatures": [
          "sg_CW3T2W6Ryi2kcDcSTeuwvL8xGhKYUDnGHygBCPLrF2aqfWA1RiybKqRRafrctK4c9vvL9DS9kCYzWkWSmD8mN9g6yhQPG"
        ],
        "tx": {
          "abi_version": 0,
          "account_id": "ak_qJZPXvWPC7G9kFVEqNjj9NAmwMsQcpRu6E3SSCvCQuwfqpMtN",
          "fee": 1842945000000000,
          "nonce": 1,
          "oracle_id": "ok_qJZPXvWPC7G9kFVEqNjj9NAmwMsQcpRu6E3SSCvCQuwfqpMtN",
          "oracle_ttl": {
            "type": "delta",
            "value": 10000000
          },
          "query_fee": 0,
          "query_format": "query",
          "response_format": "response",
          "ttl": 289505,
          "type": "OracleRegisterTx",
          "version": 1
        },
        "tx_hash": "th_K5aPLdEN4H6QduiFtqdkv61gUCvaQpDjX3z9pHKNopD8F65LJ"
      },
      "format": {
        "response": "response",
        "query": "query"
      },
[...]

PR #1874, opened 2024-07-26
fix: oracles v3 endpoint was returning v2 data

yaboiishere:
ref: #1872

### DIFF
--- a/lib/ae_mdw/oracles.ex
+++ b/lib/ae_mdw/oracles.ex
@@ -385,31 +385,24 @@ defmodule AeMdw.Oracles do
     query_format = :aeo_oracles.query_format(oracle_rec)
     response_format = :aeo_oracles.response_format(oracle_rec)
     query_fee = :aeo_oracles.query_fee(oracle_rec)
-    v3? = Keyword.get(opts, :v3?, false)
 
-    oracle =
-      %{
-        oracle: Enc.encode(:oracle_pubkey, pk),
-        active: is_active?,
-        active_from: register_height,
-        register_time: DbUtil.block_index_to_time(state, register_bi),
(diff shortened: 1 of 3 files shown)


test/ae_mdw_web/controllers/oracle_controller_test.exs
```diff
@@ -156,6 +157,77 @@ defmodule AeMdwWeb.OracleControllerTest do
...
+        assert oracle =
+                 conn
+                 |> with_store(store)
+                 |> get("/v3/oracles/#{encoded_oracle_id}")
+                 |> json_response(200)
+
+        exp1_time = @node_times[@exp1]
+        reg_time1 = @node_times[901]
+
+        assert %{
```
vatanasov, line 223: Maybe add an assertion that checks the existance of `tx_hash`, to prove it returns `v3` data?

lib/ae_mdw/oracles.ex
```diff
@@ -385,31 +385,24 @@ defmodule AeMdw.Oracles do
...
+      approximate_expire_time:
+        DbUtil.height_to_time(state, expire_height, last_gen, last_micro_time),
+      register: expand_bi_txi_idx(state, register_bi_txi_idx, opts),
+      register_tx_hash: Enc.encode(:tx_hash, Txs.txi_to_hash(state, register_txi)),
+      query_fee: query_fee,
+      format: %{
+        query: query_format,
+        response: response_format
+      },
+      extends: Enum.map(extends, &expand_bi_txi_idx(state, &1, opts))
```
sborrazas, line 404: We do not want to include extends (an unlimited list of items) on v3

sborrazas (approved):

PR #1876, opened 2024-07-29
chore: redefine node module functions directly and avoid SmartGlobal

sborrazas:


lib/ae_mdw/node.ex
```diff
@@ -69,188 +79,195 @@ defmodule AeMdw.Node do
...
-    MapSet.new()
+  defmemo id_fields() do
+    {_tx_field_types, _tx_fields, tx_ids} = types_fields_ids()
+
+    tx_ids
+    |> Enum.reduce(%{}, fn {type, ids_map}, acc ->
+      for {field, pos} <- ids_map,
+          reduce: acc,
+          do: (acc -> Map.put(acc, [field], Map.put(Map.get(acc, [field], %{}), type, pos)))
+    end)
```
vatanasov, line 155: I think this can be simplified as:
```suggestion
    for {type, ids_map} <- tx_ids,
        {field, pos} <- ids_map,
        reduce: %{},
        do: (acc -> Map.put(acc, [field], Map.put(Map.get(acc, [field], %{}), type, pos)))
```

I also think that using a normal `do` block would make it easier to read

vatanasov (approved):

Issue #1855, opened 2024-07-23
Choose between "stats" and "statistics" in v3 api


davidyuk:
"stats" use cases:
https://github.com/aeternity/ae_mdw/blob/f7cc3fad76a4bbe1ec85766bf3e5a4fdb43065b4/docs/swagger_v3/stats.spec.yaml#L167-L171
https://github.com/aeternity/ae_mdw/blob/f7cc3fad76a4bbe1ec85766bf3e5a4fdb43065b4/docs/swagger_v3/stats.spec.yaml#L196-L200
https://github.com/aeternity/ae_mdw/blob/f7cc3fad76a4bbe1ec85766bf3e5a4fdb43065b4/docs/swagger_v3/stats.spec.yaml#L225-L229

"statistics" use cases:
https://github.com/aeternity/ae_mdw/blob/f7cc3fad76a4bbe1ec85766bf3e5a4fdb43065b4/docs/swagger_v3/stats.spec.yaml#L136-L148
https://github.com/aeternity/ae_mdw/blob/f7cc3fad76a4bbe1ec85766bf3e5a4fdb43065b4/docs/swagger_v3/stats.spec.yaml#L253-L260
https://github.com/aeternity/ae_mdw/blob/f7cc3fad76a4bbe1ec85766bf3e5a4fdb43065b4/docs/swagger_v3/stats.spec.yaml#L316-L323
https://github.com/aeternity/ae_mdw/blob/f7cc3fad76a4bbe1ec85766bf3e5a4fdb43065b4/docs/swagger_v3/stats.spec.yaml#L376-L383

PR #1880, opened 2024-07-31
chore: rename statistics to stats to remain consistent

sborrazas:
refs #1855


docs/swagger_v3/stats.spec.yaml
```diff
@@ -364,11 +364,11 @@ paths:
   /statistics/transactions:
```
yaboiishere, line 364: It should be stats here

sborrazas: Done

yaboiishere (approved):

vatanasov (approved):

PR #1889, opened 2024-08-12
chore: structure oracle response as base64

sborrazas:


yaboiishere (approved):

vatanasov:
Maybe a test should be added for this? Also why was it previously decoded and now it is encoded?

sborrazas:
> Maybe a test should be added for this? Also why was it previously decoded and now it is encoded?

@vatanasov done

vatanasov (approved):

janmichek:
Is this a thing that should unite the encoding? Or do you plan to follow up with another issue?
This PR itself didn't resolve the issue,

Issue #1883, opened 2024-08-06
Next path returns the same activities, stuck on "fee_lock_name"


davidyuk:
There is a case when `next`, `prev` paths lead to the same page, making it impossible to navigate to subsequent pages.

### Steps to reproduce

docker compose
```yaml
services:
  middleware:
    image: aeternity/ae_mdw:1.82.2
    ports: [4000:4000, 4001:4001, 3013:3013, 3014:3014, 3313:3313]
    volumes:
      - ./middleware.yaml:/home/aeternity/aeternity.yaml
```
middleware.yaml
```yaml
http:
  endpoints:
    dry-run: true

chain:
  persist: false
  hard_forks:
    "1": 0
    "6": 1
  genesis_accounts:
    ak_21A27UVVt3hDkBE5J7rhhqnH5YNb4Y1dqo4PnSybrH85pnWo7E: 10000000000000000000000

fork_management:
  network_id: ae_local

mining:
  beneficiary: ak_21A27UVVt3hDkBE5J7rhhqnH5YNb4Y1dqo4PnSybrH85pnWo7E
  autostart: true
  expected_mine_rate: 1000
  micro_block_cycle: 300
  strictly_follow_top: true
  cuckoo:
    miner:
      executable: mean15-generic
      extra_args: ""
      edge_bits: 15
```
package.json
```json
{
  "dependencies": {
    "@aeternity/aepp-sdk": "github:aeternity/aepp-sdk-js#3fc8b0a60db8a666b1bf0e338d2ad94297f58e6f"
  }
}
```
init.mjs
```js
import {
  Node, AeSdkMethods, MemoryAccount, CompilerHttp, Name
} from '@aeternity/aepp-sdk';

const aeSdk = new AeSdkMethods({
  onAccount: new MemoryAccount('sk_2CuofqWZHrABCrM7GY95YSQn8PyFvKQadnvFnpwhjUnDCFAWmf'),
  onNode: new Node('http://localhost:3013'),
  onCompiler: new CompilerHttp('http://localhost:3080'),
  _expectedMineRate: 0,
  _microBlockCycle: 0,
});

const presetAccount1 = new MemoryAccount('sk_2bmJRanV8TmJzts8SYvBhR2kAd5pceVLczT5Sr8phybZYk4DRD');
const presetAccount1Address = presetAccount1.address;

async function initData() {
  console.log('waiting for height 30 to have stable block time');
  await aeSdk.awaitHeight(30);
[...]

PR #1900, opened 2024-08-16
fix: activities pagination

vatanasov:
refs: https://github.com/aeternity/ae_mdw/issues/1883

### DIFF
--- a/lib/ae_mdw/activities.ex
+++ b/lib/ae_mdw/activities.ex
@@ -216,14 +216,9 @@ defmodule AeMdw.Activities do
               {{^gen_cursor, _txi, _local_idx} = index, _data}
               when direction == :forward and index < {gen_cursor, txi_cursor, local_idx_cursor}
               when direction == :backward and index > {gen_cursor, txi_cursor, local_idx_cursor} ->
-                IO.inspect({index, {gen_cursor, txi_cursor, local_idx_cursor}})
                 true
 
-              {{^gen_cursor, _txi, _local_idx} = index, _data} ->
-                IO.inspect({index, {gen_cursor, txi_cursor, local_idx_cursor}},
-                  label: "not dropped"
-                )
-
+              {_index, _data} ->
                 false
             end)
           else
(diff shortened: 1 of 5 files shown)


lib/ae_mdw/activities.ex
```diff
@@ -527,11 +522,11 @@ defmodule AeMdw.Activities do
           nil ->
             nil
 
-          {txi_cursor, local_idx_cursor} when direction == :forward ->
-            {account_pk, pos, txi_cursor, local_idx_cursor}
+          {txi_cursor, _local_idx_cursor} when direction == :forward ->
+            {account_pk, pos, txi_cursor, @min_int}
```
sborrazas, line 526: Why is this local cursor ignored? What happens when there's more than 10 int calls in the same txi? Wouldn't the pagination loop infinitely here?

vatanasov: It shouldn't. This `local_idx` works when there are only int calls with the same `txi`, because it corresponds to the supporting idx assigned at https://github.com/aeternity/ae_mdw/blob/master/lib/ae_mdw/activities.ex#L684. But when there are other types of activities with the same txi, they are now basically different variables. For example, if we have a `NameClaimEvent` and two internal calls (with their internal `local_idx` being 0 and 1), at https://github.com/aeternity/ae_mdw/blob/master/lib/ae_mdw/activities.ex#L684 they will get assigned 0 for the name claim, 1 for the first internal call (even though his local idx is 0) and 2 for the second internal call (even though his local idx is 1). That means, that if we paginate by 1, the first page will be fine, but on the second page (where the cursor is at `local_idx` 1), it will skip over the name claim (at https://github.com/aeternity/ae_mdw/blob/master/lib/ae_mdw/activities.ex#L212) and on the int call stream it will skip over the first int call.

Basically we have a global `local_txi`, that is assigned to all the type of activities, that is being compared to the local `local_txi` for the int calls, and in case there is more than one type of activity for a given txi, this leads to skipping on pagination, because they don't match.

I am not sure if the provided explanation is clear enough, feel free to say so and I will try to provide a more descriptive one.

sborrazas: I understand how activities work, I've implemented it and the concept of using it to list activities of the same transaction. My question was the following: if we are using local_idx in the cursor to skip the first N results of things we've already seen, why are we ignoring it?

sborrazas: Nvm, I was wrong, we're using another index for txi activities. Thank you for explaining

sborrazas (approved):

Issue #1897, opened 2024-08-15
endpoint for Difficulty chart


janmichek:

yaboiishere:
What should the endpoint return? Some aggregate like the average difficulty for a day/week/month like the other stats endpoints or should it return all blocks with a dificulty attached to them?

janmichek:
Hello. yea the data should be aggregated for the chart. Thanks

yaboiishere:
The endpoint is `/v3/stats/blocks_difficulty` and it returns the average of the difficulty in for each day/week/month

janmichek:
Thanks for the preparation. I can only check it after it is deployed, right?

janmichek:
Are you sure with the url? Usually the delimiter is '-'

janmichek:
Ok good, I will chek it in a few weeks then.

janmichek:
Works nice, thank you!

PR #1911, opened 2024-08-26
feat: Add block dificulty stat

yaboiishere:
ref: #1897


docs/swagger_v3/stats.spec.yaml
```diff
@@ -394,6 +394,61 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorResponse'
+  /stats/blocks_difficulty:
```
sborrazas, line 397: Can we name this `/stats/difficulty`? There's no other difficulty related concept other than blocks

lib/ae_mdw/db/mutations/key_block_statistics_mutation.ex
```diff
@@ -0,0 +1,49 @@
...
+    State.update(
+      state,
+      Model.Statistic,
+      index,
+      fn
+        Model.statistic(count: nil) = statistic ->
+          Model.statistic(statistic, count: new_difficulty)
+
+        Model.statistic(count: old_difficulty) = statistic ->
+          Model.statistic(statistic, count: round((old_difficulty + new_difficulty) / 2))
```
sborrazas, line 45: This is not the average

lib/ae_mdw/db/mutations/key_block_statistics_mutation.ex
```diff
@@ -0,0 +1,49 @@
...
+
+  @spec new(Blocks.time(), integer()) :: t()
+  def new(time, difficulty), do: %__MODULE__{time: time, difficulty: difficulty}
+
+  @spec execute(t(), State.t()) :: State.t()
+  def execute(%__MODULE__{time: time, difficulty: difficulty}, state) do
+    time
+    |> Stats.time_intervals()
+    |> Enum.reduce(state, fn {interval_by, interval_start}, state ->
+      index = {:blocks_difficulty, interval_by, interval_start}
```
sborrazas, line 29: `{:difficulty, interval_by, interval_start}` here as well

sborrazas (changes requested):

lib/ae_mdw/db/mutations/key_block_statistics_mutation.ex
yaboiishere: Yeah, the problem was that we can't write floats to the statistic table, I tried changing the type to accept pos_integers and floats, but it didn't work (It was still giving a serialization error)

sborrazas: We already store the amount of blocks, we can simply accumulate the sum of difficulties and when we render then we divide them by the number of blocks

yaboiishere: That would be good, since it would also remove the separate mutation. Where are they stored?

sborrazas: https://github.com/aeternity/ae_mdw/blob/master/lib/ae_mdw/db/sync/stats.ex#L101

lib/ae_mdw/stats.ex
```diff
@@ -238,6 +238,25 @@ defmodule AeMdw.Stats do
...
+    with {:ok, filters} <- Util.convert_params(query, &convert_blocks_param/1),
+         {:ok, {prev, counts, next}} <-
+           fetch_statistics(state, pagination, filters, range, cursor, {:blocks, :key}),
+         {:ok, {^prev, difficulties, ^next}} <-
+           fetch_statistics(state, pagination, filters, range, cursor, :difficulty) do
+      [difficulties, counts]
+      |> Enum.zip()
+      |> Enum.map(fn {%{count: difficulty, start_date: start_date, end_date: end_date},
+                      %{count: count, start_date: start_date, end_date: end_date}} ->
+        average = if count == 0, do: 0, else: Float.round(difficulty / count, 2)
```
sborrazas, line 253: Difficulty is a pretty high number, decimals don't really matter. I would round it to the nearest value and return always an integer

sborrazas (approved):

PR #1915, opened 2024-08-27
chore: add generation of key boundaries

vatanasov:
This change is going to simplify and unify generation of key boundaries for streaming of collections (and prevent small mistakes like using the same value for beginning and end of range while typing out key boundaries).

### DIFF
--- a/lib/ae_mdw/collection.ex
+++ b/lib/ae_mdw/collection.ex
@@ -5,6 +5,7 @@ defmodule AeMdw.Collection do
 
   alias AeMdw.Database
   alias AeMdw.Db.State
+  alias AeMdw.Util
 
   @typep table() :: Database.table()
   @typep key() :: Database.key()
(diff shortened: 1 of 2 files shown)


lib/ae_mdw/collection.ex
```diff
@@ -118,6 +119,38 @@ defmodule AeMdw.Collection do
...
 
+  @spec generate_key_boundary(tuple()) :: key_boundary()
+  def generate_key_boundary(key) do
+    size = tuple_size(key)
+
+    (size - 1)..0
+    |> Enum.reduce({[], []}, fn i, {first, last} ->
+      {[get_min_key(elem(key, i)) | first], [get_max_key(elem(key, i)) | last]}
+    end)
+    |> then(fn {first, last} -> {List.to_tuple(first), List.to_tuple(last)} end)
```
sborrazas, line 130: maybe we can try something like this to avoid indexes
```elixir
key
|> Tuple.to_list()
|> Enum.map(&{get_min_key(&1), get_max_key(&1)})
|> Enum.unzip()
|> then(fn {first, last} -> {List.to_tuple(first), List.to_tuple(last)} end)
```

sborrazas (commented):
What is this change going to be used for? Can we have more context in the description?

Also, this is not a feature, features provide value to users, not to maintainers. We need to rename this to `chore`

sborrazas (approved):

Issue #1896, opened 2024-08-15
endpoint for AEX-9 token transfers


janmichek:

vatanasov:
This is resolved by https://github.com/aeternity/ae_mdw/pull/1921. Url: https://mainnet.aeternity.io/mdw/v3/stats/aex9-transfers

PR #1921, opened 2024-08-29
feat: aex9 transfer statistics

vatanasov:
refs: https://github.com/aeternity/ae_mdw/issues/1896

### DIFF
--- a/lib/ae_mdw/db/mutations/block_statistics_mutation.ex
+++ b/lib/ae_mdw/db/mutations/block_statistics_mutation.ex
@@ -27,26 +27,9 @@ defmodule AeMdw.Db.BlockStatisticsMutation do
         state
 
       count ->
-        time
-        |> Stats.time_intervals()
-        |> Enum.reduce(state, fn {interval_by, interval_start}, state ->
-          index = {:names_activated, interval_by, interval_start}
-
-          increment_statistic(state, index, count)
-        end)
+        state
+        |> Stats.increment_statistics(:names_activated, time, count)
         |> State.clear_stat(:micro_block_names_activated)
     end
-  end
-
-  defp increment_statistic(state, index, increment) do
-    State.update(
-      state,
-      Model.Statistic,
-      index,
-      fn Model.statistic(count: count) = statistic ->
-        Model.statistic(statistic, count: count + increment)
-      end,
-      Model.statistic(index: index, count: 0)
-    )
   end
 end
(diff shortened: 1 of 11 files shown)


yaboiishere (approved):

lib/ae_mdw/db/mutations/block_statistics_mutation.ex
```diff
@@ -27,26 +27,9 @@ defmodule AeMdw.Db.BlockStatisticsMutation do
...
       count ->
-        time
-        |> Stats.time_intervals()
-        |> Enum.reduce(state, fn {interval_by, interval_start}, state ->
-          index = {:names_activated, interval_by, interval_start}
-
-          increment_statistic(state, index, count)
-        end)
+        state
+        |> Stats.increase_statistics(:names_activated, time, count)
```
sborrazas, line 31: Let's keep using the increment term instead of increase

lib/ae_mdw/collection.ex
```diff
@@ -136,17 +136,21 @@ defmodule AeMdw.Collection do
...
   defp get_min_key(:integer_256bit), do: Util.min_256bit_int()
   defp get_min_key(:binary), do: Util.min_bin()
+  defp get_min_key(:interval_by), do: :day
   defp get_min_key(x), do: x
 
   defp get_max_key(:pos_integer), do: Util.max_int()
   defp get_max_key(:integer), do: Util.max_int()
   defp get_max_key(:integer_256bit), do: Util.max_int()
   defp get_max_key(:binary), do: Util.max_256bit_bin()
+  defp get_max_key(:interval_by), do: :week
```
sborrazas, line 153: This is never used and doesn't really make sense. We never want to get a range of values that includes week and day. `:day < :week < :month` is not true either

sborrazas (changes requested):

lib/ae_mdw/collection.ex
vatanasov: It is used in the tests here: https://github.com/aeternity/ae_mdw/pull/1921/files#diff-a571664afdbaba1cd53426420278d7439b14675881d08b82f57454b8a2e070fcR800 basically for getting all the related statistics. But yes, it probably won't be used outside of the tests, so I could write them out there without helpers (but as a side not, I made the last value `:week`, so it works, e.g. `:day < :month < :week`).

sborrazas (approved):

PR #1936, opened 2024-09-04
feat: add contracts count stats endpoint

yaboiishere:


lib/ae_mdw/db/mutations/stats_mutation.ex
```diff
@@ -58,7 +59,9 @@ defmodule AeMdw.Db.StatsMutation do
         },
         state
       ) do
-    m_delta_stat = make_delta_stat(state, height, from_txi, next_txi, all_cached?)
+    Model.delta_stat(contracts_created: contracts_created) =
+      m_delta_stat = make_delta_stat(state, height, from_txi, next_txi, all_cached?)
```
sborrazas, line 63: Instead of re-fetching something that has been just created, can we do `get(state, :contracts_created, 0),` ?

lib/ae_mdw/db/mutations/stats_mutation.ex
```diff
@@ -76,6 +79,14 @@ defmodule AeMdw.Db.StatsMutation do
       nil ->
         Model.stat(index: Stats.max_tps_key(), payload: {tps, key_hash})
     end)
+    |> then(fn state ->
+      time =
+        key_hash
+        |> :aec_db.get_block()
+        |> :aec_blocks.time_in_msecs()
+
+      SyncStats.increase_statistics(state, :contracts, time, contracts_created)
```
sborrazas, line 88: This looks like it belongs on the `BlockStatisticsMutation` instead of here, can we move it there?

sborrazas (changes requested):

lib/ae_mdw/db/mutations/stats_mutation.ex
yaboiishere: I'm unsure what you mean, It's only being pattern matched on the response of make_delta_stat, which is called either way for other purposes

sborrazas: you're right, we're not fetching it, we're just pattern-matching on a value that's being built

lib/ae_mdw/db/mutations/stats_mutation.ex
sborrazas: If we move it to the BlockStatisticsMutation we can re-use `increment_statistic/3`, we also wouldn't need to re-fetch the block too get the time

yaboiishere: Yes, but then we would need to calculato how many contracts have been created, since we wouldn't know if the stats in the state have been updated yet

sborrazas: That'd be better, passing the time

lib/ae_mdw_web/router.ex
```diff
@@ -94,6 +94,7 @@ defmodule AeMdwWeb.Router do
       get "/stats/total", StatsController, :total_stats
       get "/stats/delta", StatsController, :delta_stats
       get "/stats/miners", StatsController, :miners_stats
+      get "/stats/contracts-count", StatsController, :contracts_count_stats
```
sborrazas, line 97: Can we rename this route to just `/status/contracts`, none of the other routes include the word count.

sborrazas (changes requested):

sborrazas (approved):

Issue #1931, opened 2024-09-04
Query response decode bug


janmichek:
I decode Query and Response with the same algorithm. Query is obviously decoded, but Response is obviously wrong. Is there difference between encoding Query and Response?

![image](https://github.com/user-attachments/assets/d7eb439a-058e-4107-ad74-cb988e871d49)

PR #1937, opened 2024-09-09
chore: add default padding to base64 oracles response/queries

sborrazas:
refs #1931

### DIFF
--- a/lib/ae_mdw/oracles.ex
+++ b/lib/ae_mdw/oracles.ex
@@ -187,7 +187,7 @@ defmodule AeMdw.Oracles do
         query_id: Enc.encode(:oracle_query_id, query_id)
       }
       |> Map.merge(:aeo_query_tx.for_client(query_tx))
-      |> update_in(["query"], &Base.encode64(&1, padding: false))
+      |> update_in(["query"], &Base.encode64/1)
 
     if include_response? do
       Map.put(
@@ -223,7 +223,7 @@ defmodule AeMdw.Oracles do
         query_id: Enc.encode(:oracle_query_id, query_id)
       }
       |> Map.merge(:aeo_response_tx.for_client(response_tx))
-      |> update_in(["response"], &Base.encode64(&1, padding: false))
+      |> update_in(["response"], &Base.encode64/1)
 
     if include_query? do
       Map.put(response, :query, render_query(state, {oracle_pk, query_id}, false))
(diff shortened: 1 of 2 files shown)


vatanasov:
I don't think this will solve the issue. I think the problem is, some of the responses are in Chinese/Japanese (e.g. in here: https://mainnet.aeternity.io/mdw/v3/oracles/ok_qJZPXvWPC7G9kFVEqNjj9NAmwMsQcpRu6E3SSCvCQuwfqpMtN/queries ) and if my findings are correct, aescan is using `atob` which doesn't decode directly encoded Chinese/Japanese correctly. It seems that a working solution would be to first use `URI.encode` and then `Base.encode64` and in turn aescan could use `decodeURIComponent(window.atob(<response>))` which seems to return the correct response. We could also do that for the query, so that everything is uniform.

sborrazas:
@vatanasov I'm pretty sure you don't need any of that, we don't even need the base64 padding. I simply added the padding because it works with all base64 libraries, but it works with atob regardless:
Screen Shot 2024-09-11 at 8 08 06 AM

@janmichek can you describe what you are using to decode base64 strings?

sborrazas:
> image This is the frontend function depends if its server (nodejs) or browser rendered, that's what is the condition for

@janmichek can we switch to using `decodeURIComponent(escape(atob(STRING)))` in the browser and `Buffer.from(STRING, "base64").toString("utf8")` in the server? Works both with and without padding

```javascript
console.log(Buffer.from("5oiQ5Yqf5LqGIQ==", "base64").toString('utf8')); // 成功了!
console.log(Buffer.from("5oiQ5Yqf5LqGIQ", "base64").toString('utf8')); // 成功了!
```

janmichek:
> `decodeURIComponent`

Yea I guess that should work in browser

vatanasov (approved):

janmichek:
@sborrazas 
There seems to be some issue with it, can you check it please?


https://aescan.io/oracles/ok_2fVpBb8j58mNQcVjY7bwUVpBWZv6AySGDjqepNofU8XGGmEN53

image

The decoding you suggested says its a malformed string, or am I doing something wrong?

sborrazas:
@janmichek The response/query values for oracles are just binaries, meaning not all binaries have valid utf8 or ascii representations. Do you know what the utf8 representation for that example you gave should be?

sborrazas:
That's right, I would display something for when the binary is not a UTF-8/ASCII representation

janmichek:
It will be fixed in here
https://github.com/aeternity/aescan/pull/948

You can close this PR

PR #1941, opened 2024-09-11
feat: add endpoint for account claims

yaboiishere:


lib/ae_mdw/names.ex
```diff
@@ -338,6 +338,109 @@ defmodule AeMdw.Names do
     )
   end
 
+  @spec fetch_account_claims(state(), Db.pubkey(), pagination(), range(), cursor()) ::
+          {:ok, {page_cursor(), [claim()], page_cursor()}} | {:error, reason()}
+  def fetch_account_claims(state, owner_pk, pagination, scope, cursor) do
```
sborrazas, line 343: Having a hard time following this, what is this function meant to return?

yaboiishere: It's a list of all NameClaim/AuctionBidClaims an account has made

sborrazas: This does not consider accounts that made contract calls to AENS.claim

yaboiishere: I bealieve it does since some of the name_claim_tx transactions in the field table point to a contract_call transaction. An example of this would be: 

```Elixir
State.mem_state()
|> State.fetch!(
Model.Field,
{:name_claim_tx, nil,
 <<234, 64, 165, 81, 237, 92, 71, 112, 16, 48, 47, 60, 129, 49, 186, 132, 181, 221, 216,
   33, 51, 203, 170, 48, 227, 60, 235, 59, 95, 88, 168, 180>>, 60_678_614}
)
|> then(fn Model.field(index: {_type, _pos_id, _account_id, txi}) ->
    Model.tx(id: hash) = State.fetch!(State.mem_state(), Model.Tx, txi)
    AeMdw.Node.Db.get_tx(hash)
end)
```

sborrazas: On that specific case `<<234, 64, 165, 81, 237, 92, 71, 112, 16, 48, 47, 60, 129, 49, 186, 132, 181, 221, 216,
   33, 51, 203, 170, 48, 227, 60, 235, 59, 95, 88, 168, 180>>` is the name hash, not the pubkey of the account that made the claim

sborrazas (changes requested):
This can be simplified quite a lot:
1. We don't need all positions of a transaction, we want only the accounts that made the claim, not the name hash too.
2. We just need to merge these two streams:

```elixir
def fetch_account_claims(...) do
  [
    contract_call_streams(state, account_id, direction, scope, cursor),
    name_claim_tx_streams(state, account_id, direction, scope, cursor)
  ]
  |> Collection.merge(direction)
  |> Collection.paginate(pagination, &render_nested_resource(state, &1), ...)
  |> then(&{:ok, &1})
end

defp contract_call_streams(state, account_pk, ...) do
  key_boundary = {
    {account_pk, "AENS.claim", 1, Util.min_int(), nil},
    {account_pk, "AENS.claim", 1, Util.max_int(), nil}
  }
  cursor = if cursor ...
  state
  |> Collection.stream(Model.IdFnameIntContractCall, direction, key_boundary, cursor)
  |> Stream.map(fn {^account_pk, "AENS.claim", 1, txi, idx} -> {txi, idx} end)
end

defp name_claim_tx_streams(state, account_id, direction, scope, cursor) do
  tx_field_pos = 1
  key_boundary =
    case scope do
      nil ->
        Collection.generate_key_boundary(
          {:name_claim_tx, tx_field_pos, account_id, Collection.integer()}
        )

      {:gen, first_gen..last_gen} ->
        Collection.generate_key_boundary(
          {:name_claim_tx, tx_field_pos, account_id,
           Collection.gen_range(
             DbUtil.first_gen_to_txi(state, first_gen),
             DbUtil.last_gen_to_txi(state, last_gen)
           )}
        )
    end

  cursor =
    case cursor do
      nil ->
        nil

      {{txi, _idx}, tx_field_pos} ->
        {:name_claim_tx, tx_field_pos, account_id, txi}
    end

  state
  |> Collection.stream(Model.Field, direction, key_boundary, cursor)
  |> Stream.map(fn {:name_claim_tx, ^tx_field_pos, ^account_pk, txi} ->
    {txi, -1}
  end)
end
```

this should be it, no need to go through the NameClaim or AuctionClaim table

yaboiishere:
I was able to skip fetching the name and bid claims, but i still needed to get the active_from field

sborrazas (approved):

PR #1952, opened 2024-09-18
fix: contemplate case sensitivity when looking for names

vatanasov:
Currently, accessing a name in a case any other than lowercase doesn't work, e.g.:  https://mainnet.aeternity.io/mdw/v3/names/test.chain works, but https://mainnet.aeternity.io/mdw/v3/names/tEst.chain doesn't. This PR fixes that.

### DIFF
--- a/lib/ae_mdw/names.ex
+++ b/lib/ae_mdw/names.ex
@@ -959,7 +959,7 @@ defmodule AeMdw.Names do
 
   defp locate_name_or_auction_source(state, plain_name_or_hash) do
     plain_name =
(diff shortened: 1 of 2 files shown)


yaboiishere (approved):

sborrazas:
@vatanasov can there be two names with the same lowercase representation? if so, what would happen in that case?

vatanasov:
There should not be, names are normalized according to IDNA2008 - https://docs.aeternity.com/protocol/AENS/#name (which technically is not lowercasing for every language). This PR bases the search on the hash generated for the queried name by the node itself, so there should not be overlaps.

sborrazas (approved):

PR #1956, opened 2024-09-23
fix: add empty chain key-blocks validation

sborrazas:


lib/ae_mdw/db/util.ex
```diff
@@ -158,23 +166,21 @@ defmodule AeMdw.Db.Util do
...
-        if kbi <= last_gen do
-          {:ok, kbi}
-        else
-          {:error, ErrInput.NotFound.exception(value: hash_or_kbi)}
-        end
-
-      {:ok, _kbi} ->
+    with {:ok, kbi} when kbi >= 0 <- Util.parse_int(hash_or_kbi),
+         {:ok, last_gen} <- last_gen(state) do
+      if kbi <= last_gen do
```
vatanasov, line 171: I think it will be better if this is moved out in the `with` part, either as-is (the false will match with the `_invalid_kbi` case, or with wrapping it in a tuple with some descriptive atom.

sborrazas: Good call, done

yaboiishere (approved):

vatanasov (approved):

Issue #1943, opened 2024-09-13
Add counting of transactions for specific account with scoping


vatanasov:
Aescan requires information about count of transactions for specific account with scoping by height. [/transactions/count](https://mainnet.aeternity.io/mdw/swagger/#/default/GetTransactionsCount) can't take two filters and [/accounts/{id}/transactions/count](https://mainnet.aeternity.io/mdw/swagger/#/default/GetAccountTransactionsCount) doesn't support filtering by height. Probably making the second one work would be more appropriate.

janmichek:
For the feature I build for aescan, I would ideally filter by time. I would use 24hours window, actually.

yaboiishere:
Would something similar to the way stats endpoint be ok for this?

janmichek:
> Would something similar to the way stats endpoint be ok for this?

Yeah, is the 24h window also possible like this per account? That would be ideal

yaboiishere:
I think so, but the database might become way bigger. I'll try to implement it and see how it performs

yaboiishere:
@janmichek ping

janmichek:
Hey, just got back from vacation. Will think about it and decide, thanks for the suggestion.

janmichek:
This is what I want to achieve. How and which miners were active lately 

https://www.aeknow.org/miner



<img width="505" alt="image" src="https://github.com/user-attachments/assets/d3c3d742-8eb4-4e11-825b-acf8bf19b1b9">

yaboiishere:
Hmm, this sounds like we would need to modify the `stats/miners` endpoint to function like the other ones where you would be able to filter by time intervals. How do you plan to use the transactions stats endpoint to get all the miners?

janmichek:
Yeah, I should state whats my intention behind that. Sorry, at least we have it now .]
The solution would not be perfect ad would require some data transformation on frontend.
As much as you are able to pre process data for me for this chart above, the happier I will be .]]] 
I guess your last suggestion ( modify the stats/miners) seems good.

yaboiishere:
In that case I think it would be best to leave this feature as it is and create a new issue about modifying the stats/miners or creating a new endpoint to provide the data as you need it

janmichek:
you can close this one

PR #1958, opened 2024-09-27
feat: account transactions count

yaboiishere:


yaboiishere:
Closing this since it will make the database heavier and it's not needed anymore #1943

PR #1961, opened 2024-10-03
chore: add contract call index to origin

sborrazas:
This is going to allow us later on to identify creation txis by {txi, idx} tuple. Just txi is ambiguous.

### DIFF
--- a/lib/ae_mdw/db/model.ex
+++ b/lib/ae_mdw/db/model.ex
@@ -38,7 +38,6 @@ defmodule AeMdw.Db.Model do
   @typep txi() :: Txs.txi()
   @typep txi_idx() :: Txs.txi_idx()
   @typep log_idx() :: Contracts.log_idx()
-  @typep tx_hash() :: Txs.tx_hash()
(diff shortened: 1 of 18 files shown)


lib/ae_mdw/db/model.ex
```diff
@@ -180,22 +179,20 @@ defmodule AeMdw.Db.Model do
   @type id_count() :: record(:id_count, index: id_count_index(), count: non_neg_integer())
 
   # object origin :
-  #     index = {tx_type, pubkey, tx_index}, tx_id = tx_hash
-  @origin_defaults [index: {nil, nil, nil}, tx_id: nil]
```
yaboiishere, line 184: Wouldn't it be a better to leave the tx_hash field in the record? Unless the idea is for it to be only in the RevOrigin table to not duplicate the data

sborrazas: It alread holds {txi, idx} tuple, and you can compute the tx_hash using the txi, no need to store redundant data, specially since hashes take a lot more space

vatanasov (approved):

PR #1962, opened 2024-10-07
chore: remove further unused v1 routes

sborrazas:


test/ae_mdw_web/controllers/name_controller_test.exs
yaboiishere, line 1: Here you're removing some fo the v3 tests too

yaboiishere (changes requested):

test/ae_mdw_web/controllers/name_controller_test.exs
sborrazas: Hadn't realized I had removed v3 endpoint tests too, thanks 👍

yaboiishere (approved):

Issue #1942, opened 2024-09-12
Include minutes per block statistics


janmichek:
please include value

blocks per minute to v3/stats
It should be around 3 minutes per block

The formula is fairly simple. This is how I count it on FE.

<img width="776" alt="image" src="https://github.com/user-attachments/assets/60c0b336-eaf9-41bb-9b89-9759213e946a">

 I believe you are already using this approximate value in your computations, so it should be easy to expose, right?

yaboiishere:
Just to make sure firstKeyblockTime is the block at height 1, right?

janmichek:
Not super sure with it. Keyblock 0 does not have time set.

yaboiishere:
Right now I'm using the block at height 1, is something like this ok? 
```
{
    "miners_count": 211,
    "fees_trend": 0,
    "max_transactions_per_second": 52.36,
    "max_transactions_per_second_block_hash": "kh_2ek9oteSQ6KTULak4RHiqSTLebuQjMNNEujmLRZsgirMginJgW",
    "last_24hs_transactions": 0,
    "transactions_trend": 0,
    "last_24hs_average_transaction_fees": 0,
    "minutes_per_block": "0h 2m 59s 656ms"
}
```

janmichek:
I will think about the output format and let you know next week

janmichek:
> Right now I'm using the block at height 1, is something like this ok?
> 
> ```
> {
>     "miners_count": 211,
>     "fees_trend": 0,
>     "max_transactions_per_second": 52.36,
>     "max_transactions_per_second_block_hash": "kh_2ek9oteSQ6KTULak4RHiqSTLebuQjMNNEujmLRZsgirMginJgW",
>     "last_24hs_transactions": 0,
>     "transactions_trend": 0,
>     "last_24hs_average_transaction_fees": 0,
>     "minutes_per_block": "0h 2m 59s 656ms"
> }
> ```

Please format the response in milliseconds.


I am not sure about it, why the block 0 does not have any time entry?

yaboiishere:
Forgot to give an update, I'm not sure why it's 0, but it's that way on the node aswell, so it should be ok to use height 1. The feature should be in the next release

yaboiishere:
This has been deployed -> https://mainnet.aeternity.io/mdw/v3/stats

janmichek:
Ok, cool! I would just expect the value  to be named accordingly - milliseconds, or mircro time to be consistent with mdw naming?

PR #1967, opened 2024-10-11
chore: rename minutes_per_block ot milliseconds_per_block

yaboiishere:
ref: #1942


sborrazas (commented):
Can we do the full `milliseconds_per_block` instead?

sborrazas (approved):

PR #1970, opened 2024-10-14
chore: add error description to invalid aexn contracts

sborrazas:


yaboiishere (approved):

priv/migrations/20241011120000_restructure_aexn_invalid_contracts.ex
```diff
@@ -0,0 +1,40 @@
...
+    |> Collection.stream(Model.AexnInvalidContract, nil)
+    |> Stream.map(&State.fetch!(state, Model.AexnInvalidContract, &1))
+    |> Stream.filter(&match?(aexn_invalid_contract(), &1))
+    |> Stream.map(fn aexn_invalid_contract(
+                       index: index,
+                       reason: reason
+                     ) ->
+      WriteMutation.new(
+        Model.AexnInvalidContract,
+        Model.aexn_invalid_contract(index: index, reason: reason, description: "unknown")
```
vatanasov, line 28: We have one more place where we mark a contract as invalid ( https://github.com/aeternity/ae_mdw/blob/master/lib/ae_mdw/db/sync/stats.ex#L56 ), which the current PR leaves with default description (i.e. `nil`). We should probably have all the contracts without specific description as `unknown` (add description to the other place where contracts are being marked as invalid) or nil (remove the migration).

vatanasov (approved):

Issue #1968, opened 2024-10-11
Contract transfers returning Internal Server Error


yaboiishere:
Aex141 contract transfers are returning internal server errors. 

https://mainnet.aeternity.io/mdw/v3/aex141/ct_2t7TnocFw7oCYSS7g2yGutZMpGEJta6dq2DTX38SmuqmwtN6Ch/transfers

janmichek:
Looks fixed now, thanks. You can close it

PR #1971, opened 2024-10-14
fix: make transfers endpoint work with new table fields order

yaboiishere:
ref: #1968

### DIFF
--- a/lib/ae_mdw/db/contract.ex
+++ b/lib/ae_mdw/db/contract.ex
@@ -844,9 +844,9 @@ defmodule AeMdw.Db.Contract do
     create_txi = Origin.tx_index!(state, {:contract, contract_pk})
 
     m_ct_from =
-      Model.aexn_contract_from_transfer(index: {create_txi, from_pk, txi, i, to_pk, value})
+      Model.aexn_contract_from_transfer(index: {create_txi, from_pk, txi, to_pk, i, value})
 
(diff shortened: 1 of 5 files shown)


lib/ae_mdw/db/contract.ex
```diff
@@ -844,9 +844,9 @@ defmodule AeMdw.Db.Contract do
     create_txi = Origin.tx_index!(state, {:contract, contract_pk})
 
     m_ct_from =
-      Model.aexn_contract_from_transfer(index: {create_txi, from_pk, txi, i, to_pk, value})
+      Model.aexn_contract_from_transfer(index: {create_txi, from_pk, txi, to_pk, i, value})
 
-    m_ct_to = Model.aexn_contract_to_transfer(index: {create_txi, to_pk, txi, i, from_pk, value})
+    m_ct_to = Model.aexn_contract_to_transfer(index: {create_txi, to_pk, txi, from_pk, i, value})
```
sborrazas, line 849: We want to sort transfers by {txi, idx}, not by {txi, pubkey} where pubkey is not really relevant for sorting. The only thing we need here is a migration to reindex the AexnContractToTransfer and FromTransfer, which wasn't reindexed here https://github.com/aeternity/ae_mdw/commit/b32a95ca36f385e5277f04ef5edfe83e1c434d22#diff-845330bd4377f6efe9113d9a750cacdc9c00c7706c4134b59d6d12fe92e6a9c6

sborrazas (approved):

Issue #1950, opened 2024-09-18
Expose pending transactions / mempool


davidyuk:
We need it to implement features in aescan
- https://github.com/aeternity/aescan/issues/725
- https://github.com/aeternity/aescan/issues/860

Mempool can be quite big, so probably it needs to be paginated.

I would ask to implement it on the node side since this data is already tracked by the node but seems they are not willing to do so because they don't have any pagination approach.

PR #1983, opened 2024-10-21
Add mempool endpoint

yaboiishere:
ref: #1950


lib/ae_mdw_web/controllers/tx_controller.ex
```diff
@@ -175,6 +177,24 @@ defmodule AeMdwWeb.TxController do
...
 
+  @spec pending_txs(Conn.t(), map()) :: Conn.t()
+  def pending_txs(%Conn{assigns: assigns} = conn, _params) do
+    %{state: _state, pagination: pagination, cursor: cursor, scope: scope} =
+      assigns
+
+    node_state = State.new(NodeStore.new())
+
+    with {:ok, paginated_txs} <-
+           Txs.fetch_pending_txs(node_state, pagination, scope, cursor, render_v3?: true) do
```
vatanasov, line 188: As this only has a `v3` version, the opts is probably redundant.

lib/ae_mdw_web/controllers/tx_controller.ex
```diff
@@ -175,6 +177,24 @@ defmodule AeMdwWeb.TxController do
...
+      WebUtil.render(conn, paginated_txs)
+    end
+  end
+
+  @spec pending_txs_count(Conn.t(), map()) :: Conn.t()
+  def pending_txs_count(%Conn{} = conn, _params) do
+    format_json(conn, :aec_tx_pool.size())
```
vatanasov, line 195: We should probably put this function in the `Node` module, so that it is more easily switchable in the future.

lib/ae_mdw/db/model.ex
```diff
@@ -1248,7 +1265,8 @@ defmodule AeMdw.Db.Model do
       name_tables(),
       oracle_tables(),
       stat_tables(),
-      tasks_tables()
+      tasks_tables(),
+      node_tables()
```
sborrazas, line 1269: This is going to create a new rocksdb table that's not going to be used

lib/ae_mdw/db/node_store.ex
```diff
@@ -0,0 +1,85 @@
...
+  @typep record() :: tuple()
+  @typep table() :: table()
+  @opaque t() :: %__MODULE__{}
+
+  @spec new() :: t()
+  def new, do: %__MODULE__{}
+
+  @spec put(t(), table(), record()) :: t()
+  def put(_store, _table, _record) do
+    raise "Not implemented"
```
sborrazas, line 23: How are values getting written into ETS?

lib/ae_mdw/txs.ex
```diff
@@ -209,6 +211,22 @@ defmodule AeMdw.Txs do
     tx_hash
   end
 
+  @spec fetch_pending_txs(state(), pagination(), range(), cursor()) ::
+          {:ok, {page_cursor(), [tx()], page_cursor()}} | {:error, Error.t()}
```
sborrazas, line 215: Looks like there's no possible `{:error, Error.t()}` flow

lib/ae_mdw/db/node_store.ex
yaboiishere: The records are controlled by the node, the mdw just reads the table. I don't think we need to modify data there

lib/ae_mdw/txs.ex
yaboiishere: You're correct, there was in an earlier implementation that could. Removing it

lib/ae_mdw/db/node_store.ex
sborrazas: But values are being read from a table called Model.Mempool in ETS. Is that what the node is using?

yaboiishere: That's just a handle for the table, the actual table being read from is :mempool. The read functions in the store resolve Model.Mempool to :mempool

sborrazas: I see now, Model.Mempool is being converted to :mempool using Model.record/1. Thanks

sborrazas (approved):

PR #2020, opened 2024-11-18
feat: always validate cursor deserialization

sborrazas:


lib/ae_mdw/miners.ex
```diff
@@ -17,13 +18,15 @@ defmodule AeMdw.Miners do
...
-  @spec fetch_miners(state(), pagination(), cursor()) :: {cursor(), [miner()], cursor()}
+  @spec fetch_miners(state(), pagination(), cursor()) ::
+          {:ok, {cursor(), [miner()], cursor()}} | {:error, Error.t()}
   def fetch_miners(state, pagination, cursor) do
-    cursor = deserialize_cursor(cursor)
-
-    state
-    |> build_streamer(cursor)
-    |> Collection.paginate(pagination, &fetch_miner!(state, &1), &serialize_cursor/1)
+    with {:ok, cursor} <- deserialize_cursor(cursor) do
```
yaboiishere, line 24: The error clause of `deserialize_cursor` here returns nil instead of an error clause which gives an Internal Server Error

yaboiishere (changes requested):

lib/ae_mdw/miners.ex
yaboiishere: And in this case it might be a good idea to validate that the cursor is an account_pk

sborrazas: This is a feature PR which is the preamble of a bug fix. I wanted to do these two separately:
validating cursors and returning errors if it fails (feature)
fixing miners cursors when cursor is invalid (fix)
I did it wrong anyway, because this PR already fixes the bug, it just fails when the cursor is tampered with. I'll update the issue to turn it into a fix. Thank you

Issue #2019, opened 2024-11-18
Microblock gas used


janmichek:
https://mainnet.aeternity.io/mdw/v3/micro-blocks/mh_eNt38hAoDrnz4umgB3HjAMK3ocJbUPJmJcQjQVSShPEyJSKGi/

add gas used value to response, please

PR #2023, opened 2024-11-21
feat: add gas to microblocks

yaboiishere:
resolves: #2019


vatanasov (approved):

lib/ae_mdw/blocks.ex
```diff
@@ -259,10 +259,13 @@ defmodule AeMdw.Blocks do
 
     header = :aec_db.get_header(mb_hash)
 
+    gas = mb_hash |> :aec_db.get_block() |> :aec_blocks.gas()
```
sborrazas, line 262: Can we instead fetch the block and then get the header from it? That way we avoid fetching it twice
```elixir
block = :aec_db.get_block(mb_hash)
header = :aec_blocks.to_header(block)
gas = :aec_blocks.gas(block)
```

sborrazas (changes requested):

sborrazas (approved):

PR #2024, opened 2024-11-26
feat: add debug endpoint temporarily to list swaps by txi_idx

sborrazas:
Temporarily adding a debug endpoint to allow us to test. Will move it to `/dex/:contract_id/swaps` once we feel confident about it.

### DIFF
--- a/lib/ae_mdw/dex.ex
+++ b/lib/ae_mdw/dex.ex
@@ -12,6 +12,7 @@ defmodule AeMdw.Dex do
   alias AeMdw.Db.Util, as: DbUtil
   alias AeMdw.Error
   alias AeMdw.Error.Input, as: ErrInput
+  alias AeMdw.Node.Db
   alias AeMdw.Util
   alias AeMdw.Validate
   alias AeMdw.Txs
@@ -20,6 +21,7 @@ defmodule AeMdw.Dex do
 
   @account_swaps_table Model.DexAccountSwapTokens
   @contract_swaps_table Model.DexContractSwapTokens
+  @contract_debug_swaps_table Model.DexContractTokenSwap
   @swaps_table Model.DexSwapTokens
 
   @ae_token_contract_pks Application.compile_env(:ae_mdw, :ae_token)
@@ -41,6 +43,7 @@ defmodule AeMdw.Dex do
   @typep page_cursor :: Collection.pagination_cursor()
   @typep query() :: map()
   @typep range :: {:gen, Range.t()} | nil
+  @typep pubkey() :: Db.pubkey()
 
   @spec fetch_account_swaps(State.t(), binary(), pagination(), range(), cursor(), query()) ::
           {:ok, paginated_swaps()} | {:error, Error.t()}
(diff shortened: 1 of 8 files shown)


lib/ae_mdw/dex.ex
```diff
@@ -82,6 +101,28 @@ defmodule AeMdw.Dex do
     end
   end
 
+  @spec get_pair_pk(State.t(), Txs.txi(), Txs.txi(), Contracts.log_idx()) :: pubkey()
```
vatanasov, line 104: Judging from the other similar function (`get_create_txi`) there should be a case, where the `ext_contract` is `nil`, that is not handled here. Also could we combine those two functions into one (or at least extract the specific internals into separate function), because it seems that changes into one of them should be reflected in the other one?

lib/ae_mdw/dex.ex
```diff
@@ -41,6 +43,7 @@ defmodule AeMdw.Dex do
   @typep page_cursor :: Collection.pagination_cursor()
   @typep query() :: map()
   @typep range :: {:gen, Range.t()} | nil
+  @typep pubkey() :: Db.pubkey()
```
vatanasov, line 46: Do we gain something from having aliases internal to the modules, instead of using the type directly? For me it is a bit confusing, as I need to check the definition at the top of the module, instead of being certain we are using a global definition and not one specific to this module.

lib/ae_mdw/dex.ex
```diff
@@ -211,6 +252,23 @@ defmodule AeMdw.Dex do
     end
   end
 
+  defp build_debug_contract_swaps_streamer(state, create_txi_idx, cursor) do
+    fn direction ->
+      cursor =
+        case cursor do
```
vatanasov, line 258: Isn't this redundant? Unless I am mistaken, the `create_txi_idx` should also be the same, so basically the cursor always returns itself.

lib/ae_mdw/dex.ex
```diff
@@ -239,12 +297,37 @@ defmodule AeMdw.Dex do
...
 
+  defp deserialize_debug_contract_swaps_cursor(nil), do: {:ok, nil}
+
+  defp deserialize_debug_contract_swaps_cursor(cursor_bin) do
+    with {:ok, cursor_bin} <- Base.hex_decode32(cursor_bin, padding: false),
+         {{create_txi, create_idx}, txi, log_idx}
+         when is_integer(create_txi) and is_integer(create_idx) and is_integer(txi) and
+                is_integer(log_idx) <-
+           :erlang.binary_to_term(cursor_bin) do
+      {:ok, {{create_txi, create_idx}, txi, log_idx}}
```
vatanasov, line 308: Maybe using a variable to signify there are no rotation would make it easier to read, as in:
```suggestion
         {{create_txi, create_idx}, txi, log_idx} = cursor
         when is_integer(create_txi) and is_integer(create_idx) and is_integer(txi) and
                is_integer(log_idx) <-
           :erlang.binary_to_term(cursor_bin) do
      {:ok, cursor}
```

vatanasov:
We should also probably add some integration tests for the current endpoint before switching, so that we are able to notice regressions.

lib/ae_mdw/dex.ex
sborrazas: In this case it's doesn't change much, but it does help when the type signature is too long. This way it won't break it into multiple lines

lib/ae_mdw/dex.ex
sborrazas: That is true, but if we want the cursor to ONLY contain the `{txi, idx}` part then we would need to change `serialize_cursor` as well, so that it doesn't serialize what comes directly from the DB

lib/ae_mdw/dex.ex
sborrazas: Done

lib/ae_mdw/dex.ex
sborrazas: The test case where `ext_contract` is `nil` should never happen because all swaps are going to be done via the proxy contract. Added the clause anyway for tests to pass. At some point I'd like to get rid of `get_create_txi` so that that is called separately (via `Origin.tx_index(pair_pk)`)

lib/ae_mdw/dex.ex
vatanasov: I meant the `case` construct seems redundant, because it returns what it matches.

vatanasov (approved):

PR #2035, opened 2024-12-05
fix: hyperchain sync

yaboiishere:
The hyperchains implementation in the node reverses the way key and microblocks are ordered. This is only done for hyperchains.

The current way is that there is a keyblock followed by all of it's microblocks until a new keyblock is produced, where after it are it's microblocks. The breaking change here is that in hc the microblocks come first followed by the keyblock they belong to. This resulted in the mdw creating sync mutations for key blocks at height N with the microblocks inside them for the next keyblock N+1.


sborrazas (commented):
Minor comments, rest is looking good to me!

sborrazas (approved):

PR #2037, opened 2024-12-06
fix: adjust off by one stats

vatanasov:


lib/ae_mdw/stats.ex
```diff
@@ -620,7 +620,7 @@ defmodule AeMdw.Stats do
          {:ok, last_tx_index} <- State.prev(state, Model.Tx, nil),
          time_48hs_ago <- time_24hs_ago - @seconds_per_day * 1_000,
          {:ok, {_time, tx_index_48hs_ago}} <- State.next(state, Model.Time, {time_48hs_ago, -1}),
-         txs_count_24hs when txs_count_24hs > 0 <- last_tx_index - tx_index_24hs_ago,
+         txs_count_24hs when txs_count_24hs > 0 <- last_tx_index - tx_index_24hs_ago + 1,
```
sborrazas, line 623: Why plus one? If the tx_index 24 ago and the last_tx_index is the same it means there were no transactions

vatanasov: The first clause of the `with` statement gets the first tx_index that has been created since the last 24h. In the case that there have been no transaction in the 24h, then it will not pattern match and it will end in the `else` clause. 

But if there were 2 transactions, e.g. with indexes `1` and `2`, the first transaction for the last 24h will be with index `1`, the last will be with index `2`, and we will get `txs_count_24hs` equal to 1. Basically off by one bug, as we technically don't count the first transaction.

sborrazas: Fair enough

sborrazas (approved):

Issue #2054, opened 2024-12-27
Enable transactions count multiple filters 


janmichek:
https://mainnet.aeternity.io/mdw/v3/transactions/count?scope=time:1734600720-1734687120&type=spend

error	"invalid query: can't query by multiple filters and/or invalid filters"

But that's exactly what I want .] please enable filtering by multiple filters / combined parameters

yaboiishere:
I've looked into this and sadly it won't be as easy as removing the restriction. This is due to a limitation in the current implementation, where in order to have the API respond within reasonable time we count differently depending on the filters. If it's a filter by time/gen we can derive the count, but it it's by type/type_group we have to check in a table that keeps track of the count of all transaction types

janmichek:
Aha, thanks for input.
This is the use case I am trying to achieve 
https://pr-993-aescan-testnet.stg.service.aepps.com/transactions try to filter both by date and type
Maybe it will spark some idea for you

janmichek:
I have checked that. Its not going to work as the data are grouped by interval

janmichek:
That should probably work good enough, thanks. Waiting for the deployment then

PR #2074, opened 2025-01-23
feat: add endpoint to get the total count of transactions for a given period

yaboiishere:
resolves: #2054


lib/ae_mdw/stats.ex
```diff
@@ -228,6 +228,25 @@ defmodule AeMdw.Stats do
...
+  def fetch_transactions_total_stats(state, query, range) do
+    with {:ok, filters} <- Util.convert_params(query, &convert_transactions_param/1) do
+      tx_tag = Map.get(filters, :tx_type, :all)
+
+      streamer =
+        build_statistics_streamer(state, {:transactions, tx_tag}, filters, range, nil, false)
+
+      :backward
+      |> streamer.()
+      |> Enum.reduce(0, fn index, acc ->
```
sborrazas, line 242: Like we've been over before. We should never go through an unlimited list of things, this sooner or later is going to impact performance and is going to cause server downtime. We need to index everything that we query for in the database

docs/swagger_v3/stats.spec.yaml
```diff
@@ -622,6 +622,47 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorResponse'
+  /stats/transactions/total:
```
sborrazas, line 625: I would use the existing route we have for counting transactions `/transactions/count` with a `scope`  param where you scope by generation or time and the existing `tx_type` filter

sborrazas (changes requested):

lib/ae_mdw/stats.ex
yaboiishere: I don't think it's a problem here since we have ~3k records and we can also limit the request to be able to scope up to some time

docs/swagger_v3/stats.spec.yaml
yaboiishere: That route doesn't allow using scope and tx_type at the same time

lib/ae_mdw/stats.ex
sborrazas: This would be the first request that depends on the amount of records stored in the database in order to work properly. I would rather avoid going down that rabbit hole, results should be always immediate, regardless of the amount of records

docs/swagger_v3/stats.spec.yaml
sborrazas: Maybe we can allow that since it would be the same as what this new route does already

sborrazas (approved):

PR #2078, opened 2025-01-30
chore: update dialyzer to catch further errors

sborrazas:


lib/ae_mdw/sync/server.ex
```diff
@@ -496,26 +496,27 @@ defmodule AeMdw.Sync.Server do
...
-
-      mutations_map =
-        mutations
-        |> List.flatten()
-        |> Enum.reject(&is_nil/1)
-        |> Enum.frequencies_by(fn
-          %mod{} -> mod
-          nil -> nil
-        end)
+    _lookup_return =
```
yaboiishere, line 499: What is the warning that requires this?

yaboiishere (approved):

PR #2084, opened 2025-02-03
fix: replace hc check to use the node functions

yaboiishere:
This PR changes the way the mdw checks for a connection to the parent chain in a way that delegates the error handling to the node. The biggest difference is that now the node process responsible for the connection controlls when the code errors out. Now if the parent is reachable it checks periodically to see if the start_height is reached. Otherwise if it is unreachable the node gives a warning and checks again after a short interval.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -106,7 +106,8 @@ defmodule AeMdw.MixProject do
           :sext,
           :rocksdb,
           :telemetry,
-          :aeapi
+          :aeapi,
+          :aec_parent_connector
         ]
       ],
       dialyzer: dialyzer(),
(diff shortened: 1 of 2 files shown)


dincho:
Would that still crash the mdw if the parent  ode is temporarily unavailable?

yaboiishere:
> Would that still crash the mdw if the parent  ode is temporarily unavailable?

It will retry until it's available

mix.exs
```diff
@@ -106,7 +106,8 @@ defmodule AeMdw.MixProject do
           :sext,
           :rocksdb,
           :telemetry,
-          :aeapi
+          :aeapi,
+          :aec_parent_connector
```
sborrazas, line 110: Can we keep this sorted alphabetically?

sborrazas (approved):

PR #2085, opened 2025-02-04
fix: move tx fees to txs table to fix stats endpoint

yaboiishere:
This speeds the the function that gets the average tx fees in the stats endpoint. Before it timedout the endpoint


sborrazas (commented):
We are writing the fee value in a migration, but we're not setting it when it syncs. Or am I missing anything?

sborrazas (approved):

PR #2086, opened 2025-02-04
feat!: make docker paths the same as in the node images

yaboiishere:
This PR aims to unify the config and data paths between the mdw and node images so that the mdw image is a drop in replacement for the node. 

~~This relies on a change in this PR https://github.com/aeternity/ae_plugin/pull/11 and the mdw in this branch is using a fork that has implemented it. That PR would need to be merged so that this one can be changed back to the correct repo~~


sborrazas (approved):

davidyuk:
This breaking change wasn't mentioned in the release notes :/

PR #2087, opened 2025-02-04
OpenApi fixes

davidyuk:
This PR is supported by the Æternity Foundation

Check carefully. I'm not sure about all of the proposals.


docs/swagger_v3/oracles.spec.yaml
```diff
@@ -39,55 +35,29 @@ schemas:
...
-          fee:
-            type: integer
-          nonce:
-            type: integer
-          oracle_id:
-            $ref: '#/components/schemas/OracleAddress'
-          oracle_ttl:
-            description: Expiration of the oracle
-            type: object
+        allOf:
```
yaboiishere, line 38: Im not sure, but you might be able to create Oracles by a ContractCallTx and a ContractCreateTx as well which would change the tx type to them

docs/swagger_v3/names.spec.yaml
```diff
@@ -27,28 +27,18 @@ schemas:
...
-            type: integer
+        allOf:
+          - $ref: '#/components/schemas/NameRevokeTx'
+          - type: object
+            required:
+              - type
+              - version
+            properties:
+              type:
+                # TODO: define as enum
```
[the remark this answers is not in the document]
davidyuk: yep, but I can fill issues if you prefer

docs/swagger_v3/oracles.spec.yaml
davidyuk: I got it, so I can cover it with tests on sdk side and implement types accordingly. The only problem is that `oneOf` is not implemented in autorest.

davidyuk:
Can you cherry-pick good-looking commits to the master branch? I don't think I can finish this week.

docs/swagger_v3/names.spec.yaml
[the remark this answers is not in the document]
yaboiishere: I think issues would be better

docs/swagger_v3/oracles.spec.yaml
yaboiishere: Wouldn't anyOf work here?

yaboiishere:
> Can you cherry-pick good-looking commits to the master branch? I don't think I can finish this week.

We can't write directly to master. The way would be to make a separate PR with the changes you want now and then merge that

yaboiishere (approved):

Issue #2075, opened 2025-01-24
Introduce a new `scope` type for filtering results by transaction hash


sborrazas:
We should be able to scope activities (and possibly other listing) via a specific transaction. e.g. `?scope=transaction: th_djtiJjeN16qNvfi2nsGY7P8DxhTrVm8nwgvZ5RyLHabhqVXyv`.

Later on we can add this sort of scoping to other endpoints.

PR #2096, opened 2025-02-12
feat: introduce now transaction hash range scoping

sborrazas:
refs #2075

### DIFF
--- a/lib/ae_mdw/txs.ex
+++ b/lib/ae_mdw/txs.ex
@@ -557,23 +557,8 @@ defmodule AeMdw.Txs do
 
   @spec fetch(State.t(), txi() | tx_hash(), opts()) :: {:ok, tx()} | {:error, Error.t()}
   def fetch(state, tx_hash, opts) when is_binary(tx_hash) do
-    encoded_tx_hash = :aeser_api_encoder.encode(:tx_hash, tx_hash)
-
-    with mb_hash when is_binary(mb_hash) <- :aec_db.find_tx_location(tx_hash),
-         {:ok, mb_height} <- Db.find_block_height(mb_hash) do
-      state
-      |> Blocks.fetch_txis_from_gen(mb_height)
-      |> Stream.map(&State.fetch!(state, @table, &1))
-      |> Enum.find_value(
-        {:error, ErrInput.NotFound.exception(value: encoded_tx_hash)},
-        fn
-          Model.tx(id: ^tx_hash) = tx -> {:ok, render(state, tx, opts)}
-          _tx -> nil
-        end
-      )
-    else
-      _no_block_or_header ->
-        {:error, ErrInput.NotFound.exception(value: encoded_tx_hash)}
+    with {:ok, tx} <- tx_hash_to_tx(state, tx_hash) do
+      {:ok, render(state, tx, opts)}
     end
   end
 
@@ -584,8 +569,8 @@ defmodule AeMdw.Txs do
     end
   end
 
-  @spec tx_hash_to_txi(State.t(), binary()) :: {:ok, txi()} | {:error, Error.t()}
-  def tx_hash_to_txi(state, tx_hash) do
+  @spec tx_hash_to_tx(State.t(), binary()) :: {:ok, Model.tx()} | {:error, Error.t()}
+  def tx_hash_to_tx(state, tx_hash) do
     encoded_tx_hash = :aeser_api_encoder.encode(:tx_hash, tx_hash)
 
     with mb_hash when is_binary(mb_hash) <- :aec_db.find_tx_location(tx_hash),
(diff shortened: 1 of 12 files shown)


lib/ae_mdw_web/plugs/paginated_plug.ex
```diff
@@ -57,27 +63,32 @@ defmodule AeMdwWeb.Plugs.PaginatedPlug do
...
-    params
-    |> Map.delete("range_or_dir")
-    |> Map.put("range", range)
-    |> extract_direction_and_scope(txi_scope?, state)
+  defp extract_direction_and_scope(
+         %{"scope" => "transaction:" <> transaction_scope} = params,
+         _txi_scope,
+         state
+       ) do
+    with tx_hashes when length(tx_hashes) > 0 and length(tx_hashes) <= 2 <-
```
jyeshe, line 81: an alternative suggest would be to pattern match with `[_hash1, _hash2] = tx_hashes`, similar, just another flavour

lib/ae_mdw_web/plugs/paginated_plug.ex
```diff
@@ -148,40 +137,87 @@ defmodule AeMdwWeb.Plugs.PaginatedPlug do
...
-  defp extract_range(range), do: {:error, "invalid range: #{range}"}
+  defp generate_range(_state, scope_type, first, last, _params) when first < last do
+    {:ok, :forward, {scope_type, first..last}}
+  end
+
+  defp generate_range(_state, scope_type, first, last, _params) when last < first do
+    {:ok, :backward, {scope_type, last..first}}
+  end
+
+  defp generate_range(_state, scope_type, first, last, params) do
```
jyeshe, line 198: since this is the match for `first = last`, maybe instead of calling `first` and `last` just call it smth like `gen`? 
It made me thinking about it after looking at the
```
      {:ok, :forward, {scope_type, last..first}}
```
while above we have
```
    {:ok, :forward, {scope_type, first..last}}
```

lib/ae_mdw/txs.ex
```diff
@@ -584,6 +584,28 @@ defmodule AeMdw.Txs do
...
+        fn
+          Model.tx(index: txi, id: ^tx_hash) -> {:ok, txi}
+          _tx -> nil
+        end
+      )
+    else
+      _no_block_or_header ->
+        {:error, ErrInput.NotFound.exception(value: encoded_tx_hash)}
+    end
+  end
```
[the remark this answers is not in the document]
sborrazas: Not sure what you mean here, you mean fetching the full tx with the rendering by calling `fetch/3`?

lib/ae_mdw_web/plugs/paginated_plug.ex
sborrazas: I can't pattern-match on this since it's valid for it to be a single element array too

jyeshe: I mean this format `[_hash1 | _hash2] = tx_hashes` for a more compact condition but it's fine the lengthy one (:

lib/ae_mdw/txs.ex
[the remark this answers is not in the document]
jyeshe: Looking at the difference to the `fetch/3`, it would be only on the line of the `{:ok, txt}` return. 
```
def fetch(state, tx_hash, opts) when is_binary(tx_hash) do
    encoded_tx_hash = :aeser_api_encoder.encode(:tx_hash, tx_hash)

    with mb_hash when is_binary(mb_hash) <- :aec_db.find_tx_location(tx_hash),
         {:ok, mb_height} <- Db.find_block_height(mb_hash) do
      state
      |> Blocks.fetch_txis_from_gen(mb_height)
      |> Stream.map(&State.fetch!(state, @table, &1))
      |> Enum.find_value(
        {:error, ErrInput.NotFound.exception(value: encoded_tx_hash)},
        fn
          Model.tx(id: ^tx_hash) = tx -> {:ok, render(state, tx, opts)}
          _tx -> nil
        end
      )
    else
      _no_block_or_header ->
        {:error, ErrInput.NotFound.exception(value: encoded_tx_hash)}
    end
  end
```
There would be multiple ways to refactor to reuse most of it in a way that only the rendering varies. Maybe a good reason to decouple the rendering from the fetching or would consider just rendering in another way as well.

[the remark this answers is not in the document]
jyeshe: Brilliant!

jyeshe (approved):

lib/ae_mdw_web/plugs/paginated_plug.ex
jyeshe: nvm, it would be lengthier checking the rest of the list

yaboiishere (approved):

PR #2105, opened 2025-02-22
docs: remove dependency between `OracleQuery` and `OracleResponse` in OA

### LINKED ISSUE #2100: Recursive dependency between `OracleQuery` and `OracleResponse`
In `oracles.spec.yaml` there are `OracleQuery` with a `response` pointing to `OracleResponse` and `OracleResponse` with a `query` field pointing to `OracleQuery`. The problem is that per definition they are recursively nested which is not true since `OracleQuery.reponse.query` is missed in the real response.

https://github.com/aeternity/ae_mdw/blob/4220311ef39373a420fea56a53939d613f7cb5e2/docs/swagger_v3/oracles.spec.yaml#L234-L235

https://github.com/aeternity/ae_mdw/blob/4220311ef39373a420fea56a53939d613f7cb5e2/docs/swagger_v3/oracles.spec.yaml#L334-L335

Ideally, when `OracleQuery` references to `OracleResponse` the `query` field should be omitted in `OracleResponse` and vice versa.

davidyuk:
This PR is supported by the Æternity Foundation

fixes #2100

### DIFF
--- a/docs/swagger_v3/oracles.spec.yaml
+++ b/docs/swagger_v3/oracles.spec.yaml
@@ -187,21 +187,6 @@ schemas:
       block_time: 1595571314482
       height: 289005
       query_id": "oq_su6wv4MZsnJfopjAhd1Sx7HuBBtqhdHhahkoTPWk7bg24UWNR"
-      response:
-        block_hash: "mh_GWc3YBsoeCtsKW2FScDPAMz4jmK47Zdba4pPstJFJzREpWr9o"
-        block_time: 1595577843373
-        height: 289044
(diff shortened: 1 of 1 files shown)


docs/swagger_v3/oracles.spec.yaml
```diff
@@ -534,7 +494,12 @@ paths:
                       data:
                         type: array
                         items:
-                          $ref: '#/components/schemas/OracleQuery'
+                          allOf:
+                            - $ref: '#/components/schemas/OracleQuery'
+                            - type: object
+                              properties:
```
davidyuk, line 500: `example:` omitted because it is actually the same as the example in `OracleResponse`

yaboiishere (approved):

davidyuk:
I don't have permission to merge this one 🙃

PR #2108, opened 2025-02-26
chore: update docker OS versions to fix docker multiarch pipeline build

sborrazas:


jyeshe (approved):

yaboiishere (approved):
Hmm this is better than the fix I found 😅 https://github.com/aeternity/ae_mdw/tree/revert-test

PR #2109, opened 2025-02-27
feat: add accounts count to delta and total stats

### LINKED ISSUE #2040: add total acccounts to statistics endpoint
I suggest serving it from here 
https://mainnet.aeternity.io/mdw/v3/stats/total

yaboiishere:
This PR includes a long-ish migration around ~12 hours locally 
resolves: #2040


edogggy (approved):

lib/ae_mdw/db/mutations/stats_mutation.ex
```diff
@@ -252,4 +262,35 @@ defmodule AeMdw.Db.StatsMutation do
       amount_acc + amount
     end)
   end
+
+  defp get_accounts(state, from_txi, next_txi) do
+    state
+    |> Collection.stream(Model.Tx, {from_txi, next_txi})
```
sborrazas, line 268: I don't think this is right. `Model.Tx` keys are just txis, how is this key going to compare with a plain number?

lib/ae_mdw/db/mutations/stats_mutation.ex
```diff
@@ -252,4 +262,35 @@ defmodule AeMdw.Db.StatsMutation do
...
+    |> Stream.take_while(&match?(txi when txi < next_txi, &1))
```
sborrazas, line 269: No need for matching here `Stream.take_while(&(&1 < next_txi))`

lib/ae_mdw/db/mutations/stats_mutation.ex
yaboiishere: Oh, it is supposed to be the other clause of stream that accepts a key_boundary fixing it

Issue #1963, opened 2024-10-09
Create top miners endpoint


janmichek:
- fixed list of:
  { hash: 'ak_wM8yFU8eSETXU7VSN48HMDmevGoCMiuveQZgkPuRn1nTiRqyv', name: 'WoolyPooly' },
  { hash: 'ak_dArxCkAsk1mZB1L9CX3cdz1GDN4hN84L3Q8dMLHN4v8cU85TF', name: '2miners' },
  { hash: 'ak_9MsbDuBTtKegKpj5uSxfPwmJ4YiN6bBdtXici682DgPk8ycpM', name: '2miners.solo' },

- fixed time window of 24h
- blocks mined count per entry
- reference https://www.aeknow.org/miner
- data will be used for table and chart in aescan as in reference

_Originally posted by @janmichek in https://github.com/aeternity/ae_mdw/issues/1943#issuecomment-2399159977_

janmichek:
@yaboiishere I am afraid the result is something else that I requested.


This seems like grouped by time (day), which is not the same as 24hour window. Right?

yaboiishere:
Won't thss work as well? Im cautious of implementing something like that for the actual last 24 hours since that would require going over all data points and summing it up. We would prefer not doing that since if there are too many records it will result into having slow endpoints that might even timeout (an example of this is the bug that appeared in the testnet for the stats endpoint, where because of that reason it timed out).

Also we do actually know how many blocks can happen in a day for the mainnet/testnet, but implementing this would be a problem for hyperchains since there you can configure way more blocks than that, which would make the endpoint slow/unresponsive

janmichek:
Yeah, so what can I do... I am developer, not a manager. I have been waiting for over 3 months for rejecting message. Nevermind I will code it on frontend

yaboiishere:
I've given it more thought over the weekend and I've realised that hyperchains don't have miners they have validators, so the problem I had with implementing this the way you want now should be possible without slowing down the hc. This endpoint would only be available to parent chains though (i.e. PoW ones)

janmichek:
Ok, good. So when could you deliver it?

janmichek:
@yaboiishere ?

yaboiishere:
Hey, sorry I've missed this. I'm currently working on another task, but in the meantime I've starting planning on how the change woul look like. If the change goes as planned and there are no major issues with the mdw on hyperchains I hope to have it developed in the next few days

PR #2111, opened 2025-03-06
feat: top miners for the last actual 24hs

yaboiishere:
resolves: #1963 

This goes over the records of the new table, but I don't think it will be a problem since we get ~400-500 key blocks per day. The only exception to this is hyperchains, for which the endpoint has been disabled, since they don't have miners.

### DIFF
--- a/lib/ae_mdw/stats.ex
+++ b/lib/ae_mdw/stats.ex
@@ -37,6 +37,7 @@ defmodule AeMdw.Stats do
   @typep pagination() :: Collection.direction_limit()
   @typep pagination_cursor() :: Collection.pagination_cursor()
   @typep reason() :: Error.t()
+  @type miner :: map()
 
   @type nft_stats :: %{nfts_amount: non_neg_integer(), nft_owners: non_neg_integer()}
   @typep template_id() :: AeMdw.Aex141.template_id()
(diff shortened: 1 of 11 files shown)


lib/ae_mdw_web/plugs/hyperchain_plug.ex
```diff
@@ -11,13 +11,25 @@ defmodule AeMdwWeb.Plugs.HyperchainPlug do
...
+  end
+
   def call(conn, _opts) do
     if Hyperchain.hyperchain?() do
       conn
     else
       conn
       |> Conn.put_status(:bad_request)
-      |> Controller.json(%{"error" => "Not on a hyperchain node"})
+      |> Controller.json(%{"error" => "This endpoint is only available on a hyperchain node"})
```
jyeshe, line 32: This is on the else of is `hyperchain?`

lib/ae_mdw/stats.ex
```diff
@@ -870,10 +871,30 @@ defmodule AeMdw.Stats do
...
+    kb = {time_24h_ago, now}
+
+    state
+    |> Collection.stream(Model.KeyBlockTime, :backward, kb, nil)
+    |> Enum.reduce(%{}, fn index, acc ->
+      Model.key_block_time(miner: miner) = State.fetch!(state, Model.KeyBlockTime, index)
+      Map.update(acc, miner, 1, &(&1 + 1))
+    end)
+    |> Enum.map(&render_miner/1)
+    |> then(&{:ok, &1})
```
jyeshe, line 887: Neat!

jyeshe (changes requested):
Hey @yaboiishere this is great! There is only an error message that would need to be reverted.

lib/ae_mdw_web/plugs/hyperchain_plug.ex
yaboiishere: Yeah this is for endpoints only available for hyperchains like the validators and epochs ones, so if you try to access them on a regular node and mdw it will give you this error instead of an empty list

lib/ae_mdw/stats.ex
```diff
@@ -870,10 +871,30 @@ defmodule AeMdw.Stats do
     end
   end
 
+  @spec fetch_top_miners_24hs(State.t()) :: {:ok, [miner()]} | {:error, Error.t()}
+  def fetch_top_miners_24hs(state) do
```
sborrazas, line 875: Why are we returning `{:ok, ..} | {:error, ...}` when it doesn't really return error? Let's leave it returning `[miner()]`

sborrazas (changes requested):

lib/ae_mdw/stats.ex
edogggy: ok yah bumping sborrazas comment Simplify and move on.

edogggy (changes requested):
i came here for nothing  > Why are we returning `{:ok, ..} | {:error, ...}` when it doesn't really return error? Let's leave it returning `[miner()]`   
had it covered

jyeshe (approved):

jyeshe (commented):
This condition could be improved.

sborrazas (approved):

janmichek:
and so how to access this? Does not seems to be included in this https://mainnet.aeternity.io/mdw/v3/stats

yaboiishere:
It was moved here, because of the higher frequency it updates at
https://mainnet.aeternity.io/mdw/v3/stats/miners/top-24h

The docs of the endpoint haven't updated themselves in the last deployment. Looking into it

janmichek:
Got it, thanks!

Issue #2115, opened 2025-03-10
Definition of `/transactions/{hash}` lacks certain properties depending on the transaction type


davidyuk:
It [returns](https://github.com/aeternity/ae_mdw/blob/0dd63dca35ea098e824bfea00c4f25699f4fb1af/docs/swagger_v3/transactions.spec.yaml#L315) a [`#/components/schemas/Transaction`](https://github.com/aeternity/ae_mdw/blob/0dd63dca35ea098e824bfea00c4f25699f4fb1af/docs/swagger_v3/transactions.spec.yaml#L2). The [`tx`](https://github.com/aeternity/ae_mdw/blob/0dd63dca35ea098e824bfea00c4f25699f4fb1af/docs/swagger_v3/transactions.spec.yaml#L35) property is `#/components/schemas/Tx` (inherited from the node). The issue is that the node's definition does not include additional fields introduced by middleware.

For example,
https://mainnet.aeternity.io/mdw/v3/transactions/th_2ZtAgxftmG8ts2VoxxPiCWr1q83gxHHmxE3kUsJQgnEgLwYxdV
In the case of a contract call, the `tx` property of the returned JSON contains fields that are not defined in OpenAPI: `aexn_type`, `args`, `compiler_version`, `gas_used`, `log`, `return_type`, `return_value`, `source_hash`.

PR #2119, opened 2025-03-13
docs: add tx types to transactions_spec.yaml

yaboiishere:
Added the missing fields in the swagger for transactions and added the ttl field to the transactions as in the docs

resolves: #2115

### DIFF
--- a/docs/swagger_v3/contracts.spec.yaml
+++ b/docs/swagger_v3/contracts.spec.yaml
@@ -1,4 +1,132 @@
 schemas:
+  MdwContractCallExtras:
+    type: object
+    required:
+      - aexn_type
+      - arguments
+      - function
+      - gas_used
+      - log
+      - result
+      - return
+      - return_type
+      - ttl
+    properties:
+      aexn_type:
+        type: string
+        example: aex9
+      arguments:
+        type: array
+        items:
+          type: object
+          properties:
+            type:
+              type: string
+            value:
+              type: string
+        example:
+          - type: int
+            value: 1
+          - type: string
+            value: "hello"
+      function:
+        type: string
+        example: "main"
+      gas_used:
+        type: integer
+        example: 1000
+      log:
+        type: array
+        items:
+          type: object
+          properties:
+            address:
+              type: string
+            data:
+              type: string
+            topics:
+              type: array
+              items:
+                type: string
+        example:
+          - address: "ct_2U1usf3A8ZNUcZLkZe5rEoBTxk7eJvk9fcbRDNqmRiwXCHAYN"
+            data: cb_+Jg==
+            topics:
+              - "topic1"
+              - "topic2"
+      result:
+        type: string
+        example: "ok"
+      return:
+        type: object
+        properties:
+          type:
+            type: string
+          value:
+            type: string
+        example:
+          type: int
+          value: 1
+      return_type:
+        type: string
+        example: "int"
+      ttl:
+        type: integer
+        example: 1000
+  MdwContractCreateExtras:
+    type: object
+    required:
+      - aexn_type
+      - args
+      - caller_id
+      - compiler_version
+      - contract_id
+      - gas_used
+      - return_type
+      - return_value
+      - source_hash
+      - ttl
+    properties:
+      aexn_type:
+        type: string
+        example: aex9
+      args:
+        type: array
+        items:
+          type: object
+          properties:
+            type:
+              type: string
+            value:
+              type: string
+        example:
+          - type: int
+            value: 1
+          - type: string
+            value: "hello"
+      caller_id:
+        $ref: '#/components/schemas/AccountAddress'
+      compiler_version:
+        type: string
+        example: "8.0.0"
+      contract_id:
+        $ref: '#/components/schemas/ContractAddress'
+      gas_used:
+        type: integer
+        example: 1000
+      return_type:
+        type: string
+        example: "int"
+      return_value:
+        type: string
+        example: "cb_Xfbg4g=="
+      source_hash:
+        type: string
+        example: "cb_Xfbg4gasdasdasdasdadasdasd=="
+      ttl:
+        type: integer
+        example: 1000
+
   ContractInternalTx:
     description: Contract internal transaction
     type: object
@@ -56,7 +184,7 @@ schemas:
           gas: 200000
           gas_price: 1000000000
           nonce: 66
-          owner_id: ak_7wqP18AHzyoqymwGaqQp8G2UpzBCggYiq7CZdJiB71VUsLpR4 
+          owner_id: ak_7wqP18AHzyoqymwGaqQp8G2UpzBCggYiq7CZdJiB71VUsLpR4
           ttl: 0
           vm_version: 5
   ContractCall:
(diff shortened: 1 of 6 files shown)


docs/swagger_v3/transactions.spec.yaml
```diff
@@ -1,4 +1,170 @@
...
+        channel_close_solo: '#/components/schemas/ChannelCloseSoloTx'
+        channel_create: '#/components/schemas/ChannelCreateTx'
+        channel_deposit: '#/components/schemas/ChannelDepositTx'
+        channel_force_progress: '#/components/schemas/ChannelForceProgressTx'
+        channel_settle: '#/components/schemas/ChannelSettleTx'
+        channel_slash: '#/components/schemas/ChannelSlashTx'
+        channel_snapshot_solo: '#/components/schemas/ChannelSnapshotSoloTx'
+        channel_withdraw: '#/components/schemas/ChannelWithdrawTx'
+        contract_call: '#/components/schemas/ContractCallTx'
+        contract_create: '#/components/schemas/ContractCreateTx'
```
davidyuk, line 150: in the example from the issue
https://mainnet.aeternity.io/mdw/v3/transactions/th_2ZtAgxftmG8ts2VoxxPiCWr1q83gxHHmxE3kUsJQgnEgLwYxdV
```
    "type": "ContractCreateTx",
```
so, as I understand, here should be
```suggestion
        ContractCreateTx: '#/components/schemas/ContractCreateTx'
```

docs/swagger_v3/transactions.spec.yaml
```diff
@@ -1,4 +1,170 @@
...
+          - $ref: '#/components/schemas/ChannelResponderInitiatorRound'
+        - title: ChannelSnapshotSoloTx
+          allOf:
+          - $ref: '#/components/schemas/ChannelSnapshotSoloTx'
+          - $ref: '#/components/schemas/ChannelResponderInitiatorRound'
+        - $ref: '#/components/schemas/ChannelWithdrawTx'
+        - title: ContractCallTx
+          allOf:
+          - $ref: '#/components/schemas/ContractCallTx'
+          - $ref: '#/components/schemas/MdwContractCallExtras'
```
davidyuk, line 53: `ttl` is already defined in `ContractCallTx`; the difference is that it is not required (it is omitted if 0). Does mdw change this behaviour to "always present"? I think it should be consistent with the node.

docs/swagger_v3/transactions.spec.yaml
```diff
@@ -1,4 +1,170 @@
...
+        - title: GAAttachTx
+          allOf:
+          - $ref: '#/components/schemas/GAAttachTx'
+          - type: object
+            required:
+              - args
+              - auth_fun_name
+              - contract_id
+              - gas_used
+              - return_type
```
davidyuk, line 67: the only `auth_fun_name` is not defined in MdwContractCreateExtras

docs/swagger_v3/transactions.spec.yaml
```diff
@@ -1,4 +1,170 @@
...
+                items:
+                  type: string
+              auth_fun_name:
+                type: string
+              contract_id:
+                $ref: '#/components/schemas/ContractAddress'
+              gas_used:
+                type: integer
+              return_type:
+                type: string
```
davidyuk, line 80:
```suggestion
                type: string
                enum:
                  - ok
```

docs/swagger_v3/transactions.spec.yaml
```diff
@@ -1,4 +1,170 @@
...
+          - $ref: '#/components/schemas/GAMetaTx'
+          - type: object
+            required:
+              - gas_used
+              - return_type
+            properties:
+              gas_used:
+                type: integer
+              return_type:
+                type: string
```
davidyuk, line 92:
```suggestion
                type: string
                enum:
                  - ok
```

docs/swagger_v3/transactions.spec.yaml
```diff
@@ -1,4 +1,170 @@
...
+        - title: GAMetaTx
+          allOf:
+          - $ref: '#/components/schemas/GAMetaTx'
+          - type: object
+            required:
+              - gas_used
+              - return_type
```
davidyuk, line 87: also a duplicate definition

docs/swagger_v3/transactions.spec.yaml
```diff
@@ -1,4 +1,170 @@
...
+        name_update: '#/components/schemas/NameUpdateTx'
+        oracle_extend: '#/components/schemas/OracleExtendTx'
+        oracle_query: '#/components/schemas/OracleQueryTx'
+        oracle_register: '#/components/schemas/OracleRegisterTx'
+        oracle_response: '#/components/schemas/OracleRespondTx'
+        paying_for: '#/components/schemas/PayingForTx'
+        spend: '#/components/schemas/SpendTx'
+    required:
+      - type
+    type: object
```
davidyuk, line 166:
```suggestion
```
`type` already marked as required at line 7 🤔

docs/swagger_v3/contracts.spec.yaml
```diff
@@ -1,4 +1,132 @@
...
+        type: object
+        properties:
+          type:
+            type: string
+          value:
+            type: string
+        example:
+          type: int
+          value: 1
+      return_type:
```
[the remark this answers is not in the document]
yaboiishere: No, the return_type is the return status of the contract call i.e. ok, revert...

docs/swagger_v3/transactions.spec.yaml
```diff
@@ -86,7 +252,7 @@ schemas:
           $ref: '#/components/schemas/Signature'
       tx:
         description: The transaction
-        $ref: '#/components/schemas/Tx'
+        $ref: '#/components/schemas/MdwTx'
```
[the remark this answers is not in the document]
yaboiishere: You're correct it doesn't actually render the same way as the passed transactions do

docs/swagger_v3/transactions.spec.yaml
yaboiishere: I've made it always show since it was in all the specs from the node, I'll revert the change if that's the case

docs/swagger_v3/transactions.spec.yaml
yaboiishere: Can you elaborate this, I don't think I understood you

davidyuk: I meant to reuse some of the contract definitions, but maybe it isn't worth it 🤷‍♀️

docs/swagger_v3/contracts.spec.yaml
```diff
@@ -1,4 +1,140 @@
...
+      gas_used:
+        type: integer
+        example: 1000
+      return_type:
+        type: string
+        example: ok
+        enum:
+          - ok
+          - error
+          - revert
```
davidyuk, line 26: contract create can't be reverted as I know

davidyuk (approved):

davidyuk (approved):

sborrazas (approved):

Issue #2097, opened 2025-02-15
Auction bid doesn't show up in names/claims and accounts/activities


davidyuk:
I've made a bid on `auction.chain` on testnet
https://testnet.aeternity.io/mdw/v3/transactions/th_2VDreYZP63NqabKRusLscCuwszRxT7owFsRQRzRWDmJ5dMVjdR
I expect this to appear in both
https://testnet.aeternity.io/mdw/v3/accounts/nm_2AxdovXuQtVEyT4kwVzcrEAb7eb48aqZyuitnHfKHwTYiBh5dG/activities?limit=100
https://testnet.aeternity.io/mdw/v3/names/auction.chain/claims
but I can't find it there.

I think it is related to the fact that there was a different auction before on this name because when I created a fresh auction
https://testnet.aeternity.io/mdw/v3/transactions/th_2jFn6MMhmzyfHqSigXsnzb42oJFtwnBj6Ej7Kcg1QBkDUN6AyM
it worked as expected
https://testnet.aeternity.io/mdw/v3/accounts/nm_2RXLng1ZEtadnBt7Lvqwfaikz3VzxdqdBFGZpiEevXuwWZj9AW/activities?limit=100
https://testnet.aeternity.io/mdw/v3/names/auctiob.chain/claims

PR #2120, opened 2025-03-17
fix: make the claims endpoint return all claims instead of just from last auction

yaboiishere:
resolves: #2097


lib/ae_mdw/names.ex
```diff
@@ -170,25 +170,52 @@ defmodule AeMdw.Names do
...
+
+      # paginate_nested_resource(
+      #   state,
+      #   nested_table,
+      #   plain_name,
+      #   height,
+      #   scope,
+      #   cursor,
+      #   pagination
+      # )
```
sborrazas, line 218: 👀

sborrazas (approved):

lib/ae_mdw/names.ex
yaboiishere: I need to make more coffee :D

